PR #14, opened 2021-01-19
Keras pruning notebook

natuan:
Add a notebook for Keras pruning using simple MNIST model/dataset. 

TODO: download model and recipe


bfineran (commented):
Only reviewing the notebook commit since it looks like the rest is rebased from https://github.com/neuralmagic/sparseml/pull/12. Notebook looks great overall, main comment being maybe to break up the core cell.

markurtz (approved):

bfineran (approved):

PR #38, opened 2021-01-28
Turn debug off for keras onnx export for notebooks and examples

natuan:


bfineran (changes requested):
What is the downside to keeping debug mode on? My concern with explicitly turning it off for examples is users who use this flow for export will have unintended side effects when using these models elsewhere

markurtz (changes requested):
We handle nearly all optimizations that we would want in the engine. We should not be overriding in any of our pipeline for now. Everything should work end to end with any of our products.

I'm even wary of putting the warning in the export when debug=True as that is the expected flow for our system now.

markurtz (approved):

bfineran (approved):

PR #65, opened 2021-02-05
add decorator for flaky tf sparsity tests

bfineran:
Some of our tensorflow_v1 sparsity mask tests use randomly initialized tensors for testing. This can cause errors on the unlikely (but not impossible) chance that the randomly initialized tensor has a 0 in it as well as other situations.  We've seen this happen less than 5 times over the last 6 months, but it's happened twice in the last week now, so am nipping this in the bud.  @kevinaer suggested looking into flaky tests when this first occured.

Adding the flaky decorator, has pytest rerun the flaky test once (can be set to other values) if it fails.  This should be greatly decrease the chance that one of these tests will fail due to a 0 popping up in the test tensor.

https://pypi.org/project/flaky/

even though the decorator is called through pytest, I've checked that flaky must be installed for the logic to properly function.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,6 +57,7 @@ _dev_deps = [
     "sphinx-markdown-tables>=0.0.15",
     "wheel>=0.36.2",
     "pytest>=6.0.0",
+    "flaky>=3.0.0",
     "sphinx-rtd-theme",
 ]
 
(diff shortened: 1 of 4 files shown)


mgoin (approved):

mgoin:
cool, didn't know about flaky

kevinaer (approved):
LGTM

markurtz (approved):

PR #66, opened 2021-02-05
enable modifier groups in SparseML recipes

bfineran:
This change will enable SparseML to parse multiple lists of modifiers (groups) from a recipe.  The resulting structure will look like this:
```
training_modifiers:
  - !EpochRangeModifier
    ...

  - !SetLearningRateModifier
    ...

pruning_modifiers:
  - !GMPruningModifier
      ...
  - !GMPruningModifier
        ...
  - !GMPruningModifier
        ...

quantization_modifiers:
  - !QuantizationModifier
      ...
  - !SetWeightDecayModifier
      ...
```

All modifiers will be parsed into a single `Modifier` list at run time.  Not adding any filtering in the lists to see if they are actually of modifiers, so the user will be able to see any errors if the recipe is incorrect.

### DIFF
--- a/src/sparseml/optim/modifier.py
+++ b/src/sparseml/optim/modifier.py
@@ -296,9 +296,22 @@ class BaseModifier(BaseObject):
             modifiers = container
         else:  # Dict
             modifiers = []
-            for key, modifier_list in container.items():
-                if "modifiers" in key and isinstance(modifier_list, List):
-                    modifiers.extend(modifier_list)
+            for name, item in container.items():
+                if "modifiers" in name and isinstance(item, List):
+                    modifiers.extend(item)
+                elif isinstance(item, BaseModifier):
+                    modifiers.append(item)
+                elif isinstance(item, List) and any(
+                    isinstance(element, BaseModifier) for element in item
+                ):
+                    modifier_type = type(
+                        [mod for mod in item if isinstance(mod, BaseModifier)][0]
+                    )
+                    raise ValueError(
+                        "Invalid modifier location. Grouped modifiers in recipes must "
+                        "be listed in lists with 'modifiers' in its name. A modifier of "
+                        f"type {modifier_type} was found in recipe list {name}"
+                    )
 
         return modifiers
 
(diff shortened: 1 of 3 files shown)


mgoin (approved):

natuan (changes requested):
Given this will get more interesting, should we have some unit tests for it?

src/sparseml/optim/modifier.py
```diff
@@ -294,8 +294,11 @@ def load_framework_list(yaml_str: str, framework: str):
             modifiers = [container]
         elif isinstance(container, List):
             modifiers = container
-        else:
-            modifiers = container["modifiers"]
+        else:  # Dict
+            modifiers = []
+            for key, modifier_list in container.items():
+                if "modifiers" in key and isinstance(modifier_list, List):
```
markurtz, line 300: if a user accidentally mispells a group, we'll have a silent ignore here. We should at least warn if a key doesn't have modifiers in it and won't be used, and potentially could throw an exception here to be explicit since we aren't supporting anything else currently

markurtz (changes requested):
I like the simplicity, need a bit more visibility into the parsing for the user and I agree with Tuan on getting some unit tests in

src/sparseml/optim/modifier.py
bfineran: agreed. this was definitely a concern the issue is that not every top level dict element will necessarily be a modifier list.  We could have for instance a list of parameter names stored as a variable to reference elsewhere.  If this wasn't the case I would not have required "modifier" to be in the name at all.

One alternative is we could search all lists for `BaseModifier` objects and flatten that way, letting group names just be for organization. Will message in Slack.

bfineran: Actually, will raise an error when a `Modifier` object is found outside of a named "modifier" list.  That way we could keep the YAML flexibility, but enforce some simple structure in how modifiers are defined.

bfineran:
@markurtz @natuan updated with unit tests and validation for modifier list names.

markurtz (approved):

natuan (approved):

PR #68, opened 2021-02-08
match types explicitly in torch qat quant observer wrappers

bfineran:
this fixes bugs in Yolo V3 QAT where quant/dequant observers were placed in incorrect parts within the model graph

### DIFF
--- a/src/sparseml/pytorch/optim/quantization/helpers.py
+++ b/src/sparseml/pytorch/optim/quantization/helpers.py
@@ -17,14 +17,17 @@ Helper functions for performing quantization aware training with PyTorch
 """
 
 from copy import deepcopy
+from inspect import getmembers, isclass
 
 import torch
 from torch.nn import BatchNorm2d, Conv2d, Module, ReLU
 
 
 try:
+    import torch.nn.intrinsic as nni
     from torch import quantization as torch_quantization
 except Exception:
+    nni = None
     torch_quantization = None
 
 from sparseml.pytorch.nn import ReLU as ReLU_nm
@@ -37,15 +40,27 @@ __all__ = [
 ]
 
 
+_QUANTIZABLE_MODULE_TYPES = (
+    {
+        module
+        for name, module in getmembers(nni.modules.fused, isclass)
+        if "Conv" in name or "Linear" in name
+    }  # i.e. Conv2d, ConvBnReLU2d, LinearReLU, etc
+    if nni
+    else None
+)
+
+
 def add_quant_dequant(module):
     """
     Wraps all Conv and Linear submodule with a qconfig with a QuantWrapper
     :param module: the module to modify
     """
-    module_type = str(type(module)).split(".")[-1].lower()
-    is_quantizable_module = "conv" in module_type or "linear" in module_type
-
-    if is_quantizable_module and hasattr(module, "qconfig") and module.qconfig:
+    if (
+        type(module) in _QUANTIZABLE_MODULE_TYPES
+        and hasattr(module, "qconfig")
+        and module.qconfig
+    ):
         return torch_quantization.QuantWrapper(module)
 
     for name, child in module.named_children():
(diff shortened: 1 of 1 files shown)


src/sparseml/pytorch/optim/quantization/helpers.py
```diff
@@ -37,15 +40,27 @@
 ]
 
 
+_QUANTIZABLE_MODULE_TYPES = (
+    {
+        module
+        for name, module in getmembers(nni.modules.fused, isclass)
+        if "Conv" in name or "Lienar" in name
```
markurtz, line 47: typo: Lienar

markurtz, line 47: Also, this seems somewhat fragile. Is there any reason we can list out the module types or the base types?

markurtz (changes requested):

src/sparseml/pytorch/optim/quantization/helpers.py
bfineran: no, we can list them, only reason I didn't was because there are 12, will update

src/sparseml/pytorch/optim/quantization/helpers.py
```diff
@@ -37,15 +39,40 @@
...
     """
     Wraps all Conv and Linear submodule with a qconfig with a QuantWrapper
     :param module: the module to modify
     """
-    module_type = str(type(module)).split(".")[-1].lower()
-    is_quantizable_module = "conv" in module_type or "linear" in module_type
-
-    if is_quantizable_module and hasattr(module, "qconfig") and module.qconfig:
+    if (
+        type(module) in _QUANTIZABLE_MODULE_TYPES
```
natuan, line 72: When does `nni` fail to import? And in that case `_QUANTIZABLE_MODULE_TYPES` would be `None`, and this check `type(module) in None` will fail to compile.

src/sparseml/pytorch/optim/quantization/helpers.py
```diff
@@ -37,15 +39,40 @@
...
+        nni.ConvReLU3d,
+        nni.ConvBnReLU1d,
+        nni.ConvBnReLU2d,
+        nni.ConvBnReLU3d,
+        # Linear Layers
+        torch.nn.Linear,
+        nni.LinearReLU,
+    }
+    if nni
+    else None
```
natuan, line 62: Better to be [] instead of None? (see below)

natuan (changes requested):

src/sparseml/pytorch/optim/quantization/helpers.py
bfineran: the `nni` submodule isn't available on lower versions of PyTorch. our `optim.quantization` submodule isn't meant to be used with these lower versions, anyone using this function externally would already be using a compatible torch version, and internally, there are already version checks before this function is called.

src/sparseml/pytorch/optim/quantization/helpers.py
bfineran: see other comment

src/sparseml/pytorch/optim/quantization/helpers.py
natuan: NIT: add some comment as to why nni won't fail to import

natuan (dismissed):

markurtz (approved):

src/sparseml/pytorch/optim/quantization/helpers.py
```diff
@@ -37,15 +39,40 @@
...
+        nni.ConvReLU2d,
+        nni.ConvReLU3d,
+        nni.ConvBnReLU1d,
+        nni.ConvBnReLU2d,
+        nni.ConvBnReLU3d,
+        # Linear Layers
+        torch.nn.Linear,
+        nni.LinearReLU,
+    }
+    if nni
```
bfineran, line 61:
```suggestion
    if nni  # nni will always import if torch.quantization is available
```

mgoin (approved):

markurtz (approved):

PR #69, opened 2021-02-08
Serialize/deserialize MaskedLayer

natuan:
Support serialization and deserialization for models constructed for pruning using MaskedLayer layers.

### DIFF
--- a/src/sparseml/keras/optim/mask_pruning.py
+++ b/src/sparseml/keras/optim/mask_pruning.py
@@ -14,15 +14,25 @@
 
 import abc
 import collections
+import functools
 import inspect
-from typing import List
+from typing import List, Union
 
 import tensorflow as tf
 
-from sparseml.keras.optim.mask_pruning_creator import PruningMaskCreator
+from sparseml.keras.optim.mask_pruning_creator import (
+    PruningMaskCreator,
+    load_mask_creator,
+)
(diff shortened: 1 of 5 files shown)


src/sparseml/keras/optim/mask_pruning.py
```diff
@@ -51,6 +59,54 @@ def target_sparsity(self, step: int, **kwargs) -> float:
         """
         raise NotImplementedError("Not implemented")
 
+    @abc.abstractmethod
+    def get_config(self):
+        raise NotImplementedError("Not implemented")
+
+
+class SchedulerRegistry:
```
markurtz, line 67: We should be able to get rid of this class entirely and avoid the potential confusion that comes with a more complicated setup. Specifically, from_config would be a class method on the PruningScheduler and an internal registry of the derived classes can be created by overriding the __new__ definition

markurtz (changes requested):

src/sparseml/keras/optim/mask_pruning.py
natuan: Agreed that decorator is not a good solution here for registering classes. I've removed it and fixed using __init_subclass__.

markurtz (approved):

bfineran (approved):

PR #70, opened 2021-02-10
rwightman/pytorch-image-models integration

bfineran:


examples/timm-sparseml/main.py
```diff
@@ -0,0 +1,901 @@
...
+####################################################################################
+parser.add_argument(
+    "--sparseml-recipe-path",
+    required=True,
+    type=str,
+    help="path to a SparseML recipe file or a SparseZoo model stub for a recipe to load. "
+    "SparseZoo stubs should be preceded by 'zoo:'. i.e. '/path/to/local/recipe.yaml', "
+    "'zoo:zoo/model/stub'"
+)
+parser.add_argument(
```
markurtz, line 100: This argument seems to only have an effect for zoo recipes, so conditional on how sparseml-recipe-path is used. Also is restrictive if we come up with other recipe sub types other than transfer learning.

Can we potentially change it to be something like zoo-recipe-type and default it to the baseline one?

examples/timm-sparseml/main.py
```diff
@@ -0,0 +1,901 @@
...
+
+
+parser = argparse.ArgumentParser(description='PyTorch ImageNet Training')
+
+
+####################################################################################
+# Start SparseML arguments
+####################################################################################
+parser.add_argument(
+    "--sparseml-recipe-path",
```
markurtz, line 93: NIT: remove path from the arg since this supports both paths and stubs

examples/timm-sparseml/main.py
```diff
@@ -0,0 +1,901 @@
...
+parser.add_argument('--weight-decay', type=float, default=0.0001,
+                    help='weight decay (default: 0.0001)')
+parser.add_argument('--clip-grad', type=float, default=None, metavar='NORM',
+                    help='Clip gradient norm (default: None, no clipping)')
+
+
+
+# Learning rate schedule parameters
+####################################################################################
+# SparseML Integration, hide lr args, they will be overridden by SparseML
```
markurtz, line 173: Let's leave these conditionally enabled, using the LR args depends on if the loaded recipe has them or not. I would argue we don't want to force the use of modifying the LR through the recipe especially since this code has more built-in features for that right now

markurtz: To expand on it, if the loaded recipe has LR modifiers in it, then we should disable whatever the settings were given here and log a warning that the LR was overridden by the recipe

examples/timm-sparseml/main.py
```diff
@@ -0,0 +1,901 @@
...
+    )
+    optimizer = ScheduledOptimizer(
+        optimizer,
+        model,
+        manager,
+        steps_per_epoch=len(loader_train),
+        loggers=sparseml_loggers
+    )
+    start_epoch = manager.min_epochs  # override min_epochs
+    num_epochs = manager.max_epochs or num_epochs  # override num_epochs
```
markurtz, line 669: we should log out what we're overriding the epochs to

examples/timm-sparseml/main.py
```diff
@@ -0,0 +1,901 @@
...
+        args.sparseml_recipe_path, zoo_recipe_type=zoo_recipe_type
+    )
+    optimizer = ScheduledOptimizer(
+        optimizer,
+        model,
+        manager,
+        steps_per_epoch=len(loader_train),
+        loggers=sparseml_loggers
+    )
+    start_epoch = manager.min_epochs  # override min_epochs
```
markurtz, line 668: I'm not sure if users would expect this. For example, if we train for an epoch without doing anything to modify the process and then start pruning, this would improperly start at 1 and not 0 which the user had intended.

Additionally, we would like the user to be able to start in the middle of a recipe when restoring from a checkpoint

markurtz (changes requested):
Left a few comments. Additionally, can we migrate it over to your suggestion of having a folder for integrations? And let's name the root just timm since we're already in the sparseml repo

markurtz:
Oh also, can we get the exact checkpoint of the integrated repo this was tested with in the even people have issues with newer versions?

examples/timm-sparseml/main.py
bfineran: updated to include these parameters by default, but ignore the lr scheduler if any "LearningRate" modifiers are present

examples/timm-sparseml/main.py
bfineran: good point, removing

bfineran (commented):
moved this integrations to a new `integrations` directory. will move the existing integrations in a follow up PR

bfineran:
> Oh also, can we get the exact checkpoint of the integrated repo this was tested with in the even people have issues with newer versions?

I include the commit hash in the doc string

integrations/timm/main.py
```diff
@@ -0,0 +1,879 @@
...
+
+"""
+Integration between https://github.com/rwightman/pytorch-image-models and SparseML
+
+This script is adapted from https://github.com/rwightman/pytorch-image-models/blob/master/train.py
+to apply a SparseML recipe from the required `--sparseml-recipe` argument.
+Integration lines are preceded by comment blocks.  Run with `--help` for help printout,
+more information can be found in the readme file.
+
+Latest pytorch-image-models commit this script is based on: aaa715b
```
markurtz, line 16: NIT: can we add a link to the repo with the commit hash?

markurtz (changes requested):

markurtz (approved):

integrations/timm/README.md
```diff
@@ -0,0 +1,123 @@
...
+and installed from their respective source repositories.
+
+```bash
+pip install git+https://github.com/rwightman/pytorch-image-models.git
+pip install sparseml[torchvision]
+```
+
+
+## Script
+`integrations/timm/main.py` modifies
```
markurtz, line 42: one more small change, can we get the main file renamed to train.py? That way we have a one to one mapping for the file(s) from the integration with sparseml

markurtz (changes requested):

markurtz (dismissed):

mgoin (dismissed):

integrations/timm/README.md
```diff
@@ -0,0 +1,123 @@
...
+## Script
+`integrations/timm/train.py` modifies
+[`train.py`](https://github.com/rwightman/pytorch-image-models/blob/master/train.py)
+from pytorch-image-models to include a `sparseml-recipe` argument
+to run SparseML optimizations with.  This can be a file path to a local
+SparseML recipe or a SparseZoo model stub prefixed by `zoo:` such as
+`zoo:cv-classification/resnet_v1-50/pytorch-rwightman/imagenet-augmented/pruned_quant-aggressive`.
+
+Additionally, to run sparse transfer learning with a SparseZoo model that has
+a transfer learning recipe, add `?recipe_type=transfer` as part of the model stub.
```
bfineran, line 50:
```suggestion
a transfer learning recipe, add `?recipe_type=transfer_learn` as part of the model stub.
```

integrations/timm/README.md
```diff
@@ -0,0 +1,123 @@
...
+  --output models/optimized \
+  --model resnet50 \
+  --workers 8 \
+```  
+
+Training from a SparseZoo recipe and checkpoint with sparse transfer learning enabled
+```bash
+python integrations/timm/train.py \
+  /PATH/TO/DATASET/imagenet/ \
+  --sparseml-recipe zoo:model/stub/path?recipe_type=transfer \
```
bfineran, line 115:
```suggestion
  --sparseml-recipe zoo:model/stub/path?recipe_type=transfer_learn \
```

markurtz (approved):

mgoin (approved):

PR #72, opened 2021-02-13
load recipes directly from SparseZoo

bfineran:
i.e. `manager = ScheduledModifierManager.from_yaml("zoo:/model/stub/path?recipe_type=<recipe-type>")`

### DIFF
--- a/src/sparseml/utils/helpers.py
+++ b/src/sparseml/utils/helpers.py
@@ -30,6 +30,7 @@ from urllib.parse import urlparse
 import numpy
 
 from sparsezoo import Zoo
+from sparsezoo.objects import OptimizationRecipe
 from sparsezoo.utils import load_numpy_list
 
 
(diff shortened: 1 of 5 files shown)


mgoin (dismissed):
haha, this explains what i discovered in the yolov5 PR. i will echo that i think this is a great idea

markurtz (approved):

src/sparseml/utils/helpers.py
```diff
@@ -763,16 +764,32 @@ def _tensors_export_batch(
     )
 
 
-def load_recipe_yaml_str(file_path: str) -> str:
+def load_recipe_yaml_str(
+    file_path: str, zoo_recipe_type: Union[str, None] = None
```
markurtz, line 768: I'm generally not a fan of arguments that are conditional use based on the value of another arg, can lead to unfound bugs in other pipelines since there is effectively a planned non-tested, non-supported pathway. Is there another way we can find to simplify it such that the conditional logic is handled elsewhere such as putting the zoo_recipe_type as a conditional piece of the stub?

Also, we should take in a SparseZoo Recipe object here as well. That could also be another route for supplying the zoo_recipe_type that we don't need to explicitly call out here.

markurtz (changes requested):
Small complaint that I noticed afterwards

src/sparseml/utils/helpers.py
markurtz: I do like the query string approach and definitely seems very flexible with something like a sparse_zoo_stub function. Is there any reason to not have a Recipe constructor act as the sparse_zoo_stub function or have the Recipe constructor call into the sparse_zoo_stub? I think those combined with allowing a Recipe class to be passed in would cover all basis and give us enough flexibility to document an easy flow around. Additionally can easily identify each recipe by the stub through the UI's with easy copy and paste and additionally document flows in the recipe markdowns themselves for use cases

bfineran: updated these changes so Managers can accept stubs or `sparsezoo.objects.OptimizationRecipe` objects. additionally, all logic for recipe types has been moved to SparseZoo by supporting query strings as we discussed.

markurtz (approved):

natuan (approved):

PR #74, opened 2021-02-15
Native Keras support

natuan:
Allow to use native Keras. Only Keras >= 2.4.3 is supported due to the following reasons:
- for Keras < 2.4.0, the `clone_model` function does not have the param `clone_layer` required by the modifiers;
- the model's `fit` functions in 2.4.0/2.4.1/2.4.2 do not work with the current custom callbacks such as `LRModifierCallback` w/ error `AttributeError: 'LRModifierCallback' object has no attribute '_implements_train_batch_hooks'`

For TF2.0, model save hit an assertion inside TF, and therefore we only enable >=TF2.1.

### DIFF
--- a/src/sparseml/keras/__init__.py
+++ b/src/sparseml/keras/__init__.py
@@ -22,11 +22,24 @@ editing models for performance in the Neural Magic System
 try:
     import tensorflow
 
-    version = [int(v) for v in tensorflow.__version__.split(".")]
-    if version[0] != 2 or version[1] < 2:
-        raise Exception
+    if tensorflow.__version__ < "2.1.0":
+        raise RuntimeError("TensorFlow >= 2.1.0 is required, found {}".format(version))
 except:
     raise RuntimeError(
-        "Unable to import tensorflow. tensorflow>=2.2 is required"
+        "Unable to import tensorflow. TensorFlow>=2.1.0 is required"
         " to use sparseml.keras."
     )
+
+
+try:
+    import keras
+
+    v = keras.__version__
+    if v < "2.4.3":
+        raise RuntimeError(
+            "Native keras is found and will be used, but required >= 2.4.3, found {}".format(
+                v
+            )
+        )
+except:
+    pass
--- a/src/sparseml/keras/optim/manager.py
+++ b/src/sparseml/keras/optim/manager.py
@@ -23,6 +23,7 @@ from typing import List, Union
 import tensorflow as tf
 
 from sparseml.keras.optim.modifier import Modifier, ScheduledModifier
+from sparseml.keras.utils.compat import keras
 from sparseml.keras.utils.logger import KerasLogger
 from sparseml.optim import BaseManager
 from sparseml.utils import load_recipe_yaml_str
@@ -71,8 +72,8 @@ class ScheduledModifierManager(BaseManager, Modifier):
 
     def modify(
         self,
-        model: Union[tf.keras.Model, tf.keras.Sequential],
-        optimizer: tf.keras.optimizers.Optimizer,
+        model: Union[keras.Model, keras.Sequential],
+        optimizer: keras.optimizers.Optimizer,
         steps_per_epoch: int,
         loggers: Union[KerasLogger, List[KerasLogger]] = None,
         input_tensors: tf.Tensor = None,
@@ -106,14 +107,14 @@ class ScheduledModifierManager(BaseManager, Modifier):
                 continue
             if isinstance(callback, list):
                 callbacks = callbacks + callback
-            elif isinstance(callback, tf.keras.callbacks.Callback):
+            elif isinstance(callback, keras.callbacks.Callback):
                 callbacks.append(callback)
             else:
                 raise RuntimeError("Invalid callback type")
         self._optimizer = optimizer
         return model, optimizer, callbacks
 
-    def finalize(self, model: tf.keras.Model):
+    def finalize(self, model: keras.Model):
         """
         Remove extra information related to the modifier from the model that is
         not necessary for exporting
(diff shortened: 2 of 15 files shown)


src/sparseml/keras/optim/manager.py
```diff
@@ -22,6 +22,12 @@
 
 import tensorflow as tf
 
+
+try:
```
markurtz, line 26: We shouldn't be repeating this everywhere. Please move this to a utils function and import keras from that utility like we did for tf_compat

src/sparseml/keras/optim/manager.py
```diff
@@ -22,6 +22,12 @@
...
+    import keras
+except ModuleNotFoundError:
```
markurtz, line 28: This catch needs to be more generic like with Exception. We also should have some logging built-in around this at least at the debug level

src/sparseml/keras/__init__.py
```diff
@@ -24,9 +24,23 @@
...
+        "Unable to import tensorflow. TensorFlow>=2.2 is required"
         " to use sparseml.keras."
     )
+
+
+try:
+    import keras
+
+    v = keras.__version__
+    if v < "2.4.3":
```
markurtz, line 39: Why are we only supporting the very latest release and the very latest bug release at that? Let's expand this support, otherwise there really isn't much of a point to supporting Keras here since it was pushed to the TF package as the main flow before this

src/sparseml/keras/__init__.py
```diff
@@ -24,9 +24,23 @@
 
     version = [int(v) for v in tensorflow.__version__.split(".")]
     if version[0] != 2 or version[1] < 2:
-        raise Exception
+        raise RuntimeError("TensorFlow >= 2.2 is required, found {}".format(version))
```
markurtz, line 27: Why are we still restricted on 2.2 right now?

markurtz (changes requested):

src/sparseml/keras/optim/modifier_lr.py
```diff
@@ -19,8 +19,16 @@
 from typing import Dict, List, Union
 
 import tensorflow as tf
-from tensorflow.keras import backend as K
-from tensorflow.keras.optimizers.schedules import LearningRateSchedule
+
+
+try:
```
markurtz, line 24: need to change this over to our shared conditional keras import

src/sparseml/keras/optim/modifier_params.py
```diff
@@ -19,7 +19,13 @@
 
 from typing import List, Union
 
-import tensorflow
+import tensorflow as tf
```
markurtz, line 22: NIT: remove the abbreviation for tensorflow

src/sparseml/keras/optim/modifier_pruning.py
```diff
@@ -19,7 +19,13 @@
 
 from typing import Dict, List, Union
 
-import tensorflow
+import tensorflow as tf
+
+
+try:
```
markurtz, line 25: need to change this over to our shared conditional keras import

src/sparseml/keras/optim/modifier_pruning.py
```diff
@@ -19,7 +19,13 @@
 
 from typing import Dict, List, Union
 
-import tensorflow
+import tensorflow as tf
```
markurtz, line 22: NIT: remove abbreviation for tensorflow

src/sparseml/keras/optim/modifier_pruning.py
```diff
@@ -222,10 +228,8 @@ def target_sparsity(self, step: int, tensor=None) -> float:
         if tensor is None:
             raise ValueError("Invalid empty tensor")
         if self._start_step <= step < self._end_step:
-            mask = tensorflow.cast(tensorflow.not_equal(tensor, 0.0), tensor.dtype)
-            sparsity = tensorflow.math.reduce_sum(1.0 - mask).numpy() / tensorflow.size(
-                tensor
-            )
+            mask = tf.cast(tf.not_equal(tensor, 0.0), tensor.dtype)
+            sparsity = tf.math.reduce_sum(1.0 - mask).numpy() / tf.size(tensor)
```
markurtz, line 232: this is going to return a numpy float and not a python float. Need to update the types or use .item.

Additionally, the sparsity is set to None and returned and the return type does not reflect that for this function

src/sparseml/keras/utils/exporter.py
```diff
@@ -19,7 +19,11 @@
 import os
 from typing import Any, List
 
-from tensorflow.keras import Model
+
+try:
```
markurtz, line 23: need to change this over to our shared conditional keras import

tests/sparseml/keras/optim/mock.py
```diff
@@ -16,7 +16,12 @@
 
 import numpy as np
 import tensorflow as tf
-from tensorflow.keras.models import Model
+
+
+try:
```
markurtz, line 21: need to change this over to our shared conditional keras import

tests/sparseml/keras/optim/test_modifier_pruning.py
```diff
@@ -16,6 +16,12 @@
 import pytest
 import tensorflow as tf
 
+
+try:
```
markurtz, line 20: need to change this over to our shared conditional keras import

markurtz (changes requested):

src/sparseml/keras/__init__.py
natuan: Updated to support from TF 2.1: for TF2.0 there's a bug in TF model save.

src/sparseml/keras/__init__.py
markurtz: Resolved, we can't support clone_model for previous versions. 2.4.3 is when they did the merge with tensorflow which is the reason why it works well and is an exact copy of the 2.2 implementation

markurtz (approved):

bfineran (approved):

PR #96, opened 2021-03-08
Keras: train, prune, evaluate and export scripts

natuan:


natuan:
Blocked by: 
https://github.com/neuralmagic/sparseml/pull/95
https://github.com/neuralmagic/sparseml/pull/97

scripts/keras_classification.py
```diff
@@ -0,0 +1,738 @@
...
+                        given epoch completes until the end of training
+  --save-epochs SAVE_EPOCHS [SAVE_EPOCHS ...]
+                        epochs to save checkpoints at
+  --init-lr INIT_LR     The initial learning rate to use while training, the
+                        actual initial value used should be set by the
+                        sparseml recipe
+  --optim-args OPTIM_ARGS
+                        Additional args to be passed to the optimizer passed
+                        in as a json object
+#############
```
markurtz, line 118: For this help doc, please match the style setup in pytorch_vision.py and other scripts. Specifically the use of spaces and the use of Example command rather than Example:

scripts/keras_classification.py
```diff
@@ -0,0 +1,738 @@
...
+        )
+        par.add_argument(
+            "--save-dir",
+            type=str,
+            default="keras_classification",
+            help="The path to the directory for saving results",
+        )
+
+        # task specific arguments
+        if par == train_parser:
```
markurtz, line 312: why not add these outside of the for loop if they're only being applied to one?

scripts/keras_classification.py
```diff
@@ -0,0 +1,738 @@
...
+                " as a json object",
+            )
+            par.add_argument(
+                "--run-eagerly",
+                type=bool,
+                default=True,
+                help="Run training in eager execution mode",
+            )
+
+        if par == eval_parser:
```
markurtz, line 411: why not add these outside of the for loop if they're only being applied to one?

scripts/keras_classification.py
```diff
@@ -0,0 +1,738 @@
...
+                help="the number of parallel workers for dataset loading",
+            )
+            par.add_argument(
+                "--test-prefetch-buffer-size",
+                type=int,
+                default=None,
+                help="Prefetch buffer size for test dataset loading",
+            )
+
+        if par == export_parser:
```
markurtz, line 432: why not add these outside of the for loop if they're only being applied to one?

scripts/keras_classification.py
```diff
@@ -0,0 +1,738 @@
...
+"""
+
+import argparse
+import json
+import math
+import os
+from typing import Optional, Tuple
+
+import numpy as np
+import tensorflow as tf
```
markurtz, line 181: remove the abbreviation of tensorflow

scripts/keras_classification.py
```diff
@@ -0,0 +1,738 @@
...
+from tensorflow import keras
```
markurtz, line 182: to make this properly compatible we should use our keras import that checks between tensorflow.keras and native keras

scripts/keras_classification.py
```diff
@@ -0,0 +1,738 @@
...
+            **kwargs,
+        )
+
+    if (
+        args.pretrained_dataset is not None
+        and args.dataset is not None
+        and args.pretrained_dataset != args.dataset
+    ):
+        # Set up the model for fine-tuning
+        model.trainable = False
```
markurtz, line 577: We should not restrict the whole model from being trainable. For transfer learning, we want to keep the whole model trainable. If the desire is to turn off layers from being trained, that should be done through a recipe and not hardcoded

scripts/keras_classification.py
```diff
@@ -0,0 +1,738 @@
...
+        new_output = keras.layers.Dense(num_classes, activation="softmax")(
```
markurtz, line 578: We can't force use of softmax here, need to figure out a more generic way to allow different activations such as sigmoid for multi class

scripts/keras_classification.py
```diff
@@ -0,0 +1,738 @@
...
+            model.layers[-2].output
+        )
+        model = keras.Model(model.input, new_output)
+
+    return model
+
+
+def create_optimizer(args):
+    optim_const = {
```
markurtz, line 587: exact mapping of names to properties, let's use a lookup from keras.optimizers

scripts/keras_classification.py
```diff
@@ -0,0 +1,738 @@
...
+    input_shape = ModelRegistry.input_shape(args.arch_key)
+    image_size = (input_shape[0], input_shape[1])
+
+    train_dataset, num_classes = create_dataset(args, train=True, image_size=image_size)
+    train_dataset = train_dataset.build(args.num_samples)
+
+    model = create_model(args, input_shape, num_classes=num_classes)
+    model = remove_pruning_masks(model)
+    exporter = ModelExporter(model, args.save_dir)
+    exporter.export_onnx(opset=args.onnx_opset)
```
markurtz, line 716: let's change over the default for debug mode or pass that in here. Currently as exported, the models can't be run performantly in ORT or deepsparse

markurtz (changes requested):
small comments, overall structure looks good

scripts/keras_classification.py
natuan: fixed

scripts/keras_classification.py
natuan: Add an argument for the transfer class_type, and optionally the name of the layer whose outputs to be connected to the new layer for transfer learning (default None for model.layers[-2])

markurtz (approved):

bfineran (approved):

PR #98, opened 2021-03-10
[WIP] PyTorch Movement Pruning

bfineran:
Based on [this paper](https://arxiv.org/pdf/2005.07683.pdf)

Changes to existing APIs:
* movement pruning does not take the absolute value of its scores, so to accommodate this, pytorch mask creators no longer run absolute value, and this is now called outside of the mask creator


bfineran:
TODO:
* test against BERT models
* add `GradualMagnitudePruningModifier` and `GradualMovementPruningModifier` ease of use classes that inherit from `GMPruningModifier`
* unit tests

bfineran:
Closing in favor of #216 which integrates with latest `main` branch and is compatible with more generic pipelines

PR #99, opened 2021-03-11
Convert models from Pytorch to Keras

natuan:


src/sparseml/keras/utils/pytorch2keras.py
```diff
@@ -0,0 +1,347 @@
+import re
```
markurtz, line 1: let's move this script over to the utils folder under root since it's a utility script and not meant to be imported anywhere

src/sparseml/keras/utils/pytorch2keras.py
```diff
@@ -0,0 +1,347 @@
...
+        "resnet152",
+        "pruned-moderate",
+        "single",
+        resnet_mapping,
+        output_dir,
+        IMAGENET_DIR,
+    )
+
+
+def main():
```
markurtz, line 342: can we add in some args and make the function callable for the key and weights to use? This would be instead of hardcoding the conversions into the script as that won't scale long term

src/sparseml/keras/utils/pytorch2keras.py
markurtz, line 1: Script needs to be copy righted as well, be sure to run make style

markurtz, line 1: Finally let's add some docs at the top to go over what the script is for and give examples for the command to run with the args mentioned in another diff

markurtz (changes requested):
Left a few comments and will re-review once those are addressed other diffs land so this can be rebased

utils/pytorch2keras.py
```diff
@@ -0,0 +1,463 @@
...
+        os.makedirs(model_dir)
+    model_file_path = os.path.join(model_dir, "model.h5")
+
+    # Convert and save
+    kr_model = _convert_model(
+        arch_key, type_, class_type, layer_mapping, test_image_file_path
+    )
+    kr_model.save(model_file_path)
+
+    # verify_keras_model(kr_model, imagenet_dir)
```
kevinaer, line 334: Small nit but remove commented code

utils/pytorch2keras.py
```diff
@@ -0,0 +1,463 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Script to convert Pytorch models to Keras
+
```
kevinaer, line 17: Nit but can you add the results of `python utils/pytorch2keras.py --help` so we can keep documentation content consistent with other scripts

kevinaer (commented):
Looks good to me, left a few small nits

markurtz (approved):

bfineran (approved):
great job getting this working @natuan!

kevinaer (approved):

PR #100, opened 2021-03-12
improve runtime for constant to initializer conversion in torch qat->…

bfineran:
…quant flow

minimized the number of graph traversals needed, and got rid of calling many O(n) `remove()` operations

side effect changes:
* some non-quantizations constants such as constants for global pools may be now converted to initializers, this is to avoid checking all quantization nodes for their children, is still compatible with ONNX spec

Tested that graph structure is correct and it runs in ORT and DeepSparse with the same accuracy for resnet50.

Speedup on rn50 (seconds):
```
old: 73.66600489616394
new: 0.0065975189208984375
```

### DIFF
--- a/src/sparseml/pytorch/optim/quantization/quantize_qat_export.py
+++ b/src/sparseml/pytorch/optim/quantization/quantize_qat_export.py
@@ -205,11 +205,16 @@ def _convert_single_constants_to_initializers(model: ModelProto):
         if node.op_type != "Constant" or len(node.attribute) != 1:
             non_single_constant_nodes.append(node)
             continue  # skip non-constants, and constants with multiple tensors
-        # extract tensor and name it with the appropriate edge
-        const_tensor = node.attribute[0].t
-        const_tensor.name = node.output[0]
+
+        # create initializer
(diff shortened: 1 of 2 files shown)


mgoin (dismissed):
yeah looks good to me, maybe a bit fragile if protobuf format changes but we should definitely do more batching of operations like this (i.e. finding all the nodes to delete and then deleting them)

bfineran:
Noticing a small (<1%) accuracy discrepancy. Investigating.

bfineran:
accuracy error fixed by keeping in the int8 -> uint8 conversion. this adds some negligible execution time.  from what I understand, this error is likely caused by ORT expecting int8 weights to be calibrated for a range of [-127,127] (a reduction of 1 from the actual range), but our torch QAT calibrating for [-128,127].  It's possible to change this, but in the meantime uint8 is fine since ORT uses a range of [0,255] which maps correctly to the [-128,127] that our models are calibrated for.

mgoin (approved):

markurtz (approved):

PR #101, opened 2021-03-15
option to use LeakyReLU activations in yolo integration

bfineran:
default activations in ultralytics/yolov5 were switched to SiLU in january, this flag replaces those activations with LeakyReLU, which was previously used for yolo v3 and is friendlier for onnx export + quantization

### DIFF
--- a/integrations/ultralytics/train.py
+++ b/integrations/ultralytics/train.py
@@ -50,6 +50,7 @@ from utils.loss import ComputeLoss
 from utils.plots import plot_images, plot_labels, plot_results, plot_evolution
 from utils.torch_utils import ModelEMA, select_device, intersect_dicts, torch_distributed_zero_first
 
+from sparseml.pytorch.nn import replace_activations
 from sparseml.pytorch.optim import ScheduledModifierManager, ScheduledOptimizer
 from sparseml.pytorch.utils import ModuleExporter, PythonLogger, TensorBoardLogger
 from sparsezoo import Zoo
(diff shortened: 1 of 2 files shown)


integrations/ultralytics/train.py
```diff
@@ -50,6 +50,7 @@
 from utils.plots import plot_images, plot_labels, plot_results, plot_evolution
 from utils.torch_utils import ModelEMA, select_device, intersect_dicts, torch_distributed_zero_first
 
+from sparseml.pytorch.nn import replace_activations
```
markurtz, line 53: Quick thought, would it make sense to put the imports under our commented blocks/where the code is used? Ideally, that way it makes it easier to swap out -- I'm not sold on needing to do that, though. Just a thought

integrations/ultralytics/train.py
```diff
@@ -535,6 +539,11 @@ def train(hyp, opt, device, tb_writer=None, wandb=None):
         help="set flag to enable Automatic Mixed Precision (AMP). disabled by default "
         "in SparseML integration"
     )
+    parser.add_argument(
+        "--leaky-relu",
+        action="store_true",
```
markurtz, line 544: given that our default is LeakyRelu for now, would it make sense to change this to store_false along with a negative on the name?

markurtz (commented):
Double checking, all of ultralytics model defs assign an attribute for activations instead of using torch functional? Also left some other comments

integrations/ultralytics/train.py
bfineran: good idea. done.

integrations/ultralytics/train.py
bfineran: Looks good to me from a usability perspective.  I changed it to something a little more verbose than `no-leaky-relu` because there are still configs that can generate a model that uses leakyrelu that this arg wouldn't affect.

markurtz (approved):

mgoin (approved):

PR #102, opened 2021-03-16
generic onnx quantized residual add fusor graph optimization

bfineran:
To fuse quantized adds, the DeepSparse engine needs both inputs to be near dequantize nodes.  this is not the immediate structure of common target models (ie Yolo-v3, BERT).  This change adds an optimization pass to the QAT->Quantized ONNX export flow that adds an additional dequantize node to the identity brach of quantized residual blocks, enabling the DeepSparse engine to fuse the residual adds as quantized operations.

Additionally, this PR adds the `ONNXGraph` object that caches the edges in an ONNX model graph, enabling common lookups and edits to run an order of magnitude or two quicker.  This will be used in the general effort to overhaul and improve the QAT export flow.  Using this object, this optimization runs in less than 10ms on yolo-v3.

Example residual before optimization:
Screen Shot 2021-03-16 at 1 31 59 PM

After:
Screen Shot 2021-03-16 at 3 50 32 PM

### DIFF
--- a/src/sparseml/onnx/utils/graph_editor.py
+++ b/src/sparseml/onnx/utils/graph_editor.py
@@ -16,16 +16,18 @@
 Helper functions to edit ONNX Graphs.
 """
 
-from typing import Iterable, List, Union
+from collections import defaultdict
+from typing import Iterable, List, Optional, Union
 
 import numpy
 import onnx
-from onnx import ModelProto, NodeProto, numpy_helper
+from onnx import ModelProto, NodeProto, TensorProto, numpy_helper
 
 from sparseml.onnx.utils.helpers import get_node_params
 
 
 __all__ = [
+    "ONNXGraph",
     "update_model_param",
     "swap_node_output",
     "remove_node_and_params_from_graph",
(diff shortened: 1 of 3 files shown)


src/sparseml/onnx/utils/graph_editor.py
```diff
@@ -36,6 +38,110 @@
...
+        self.update()
+
+    def update(self, model: Optional[ModelProto] = None):
+        """
+        Update the graph state based on the model this graph represents or
+        the given model.
+
+        :param model: model to represent. defaults to current loaded model state
+        """
+        model = model or self._model
```
mgoin, line 64: this should just update `self._model`

mgoin (commented):
i think this already looks pretty nice, thanks Ben. 

one thing that would be nice would be if a function call like `get_init_by_name` failed (as in we couldn't find the initializer), it would call `update()` and try again? maybe a decorator trick? you don't have to do this

mgoin (dismissed):

src/sparseml/onnx/utils/graph_editor.py
bfineran: good catch, updated

bfineran:
> i think this already looks pretty nice, thanks Ben.
> 
> one thing that would be nice would be if a function call like `get_init_by_name` failed (as in we couldn't find the initializer), it would call `update()` and try again? maybe a decorator trick? you don't have to do this

Thanks, I thought about something like that for refreshing the cache, but there's also situations where we wouldn't want to refresh, like if we aren't sure if the initializer exists (could be a constant, or another node in the graph etc).  We could add an option to update or not on fail

bfineran:
@mgoin I think for now (maybe use cases will change), it would be best for users to know when the underlying graph changes outside of calls to `ONNXGraph` and have to acknowledge with `update()`.  i do like the idea of a decorator to handle this though.

mgoin (dismissed):

bfineran:
adding in an extra optimization to quantize all outputs of adds that we have the information for. this is relevant for the upsampling/concats in yolo v3 models

before:
Screen Shot 2021-03-16 at 6 02 51 PM

after:
Screen Shot 2021-03-16 at 6 03 01 PM

src/sparseml/onnx/utils/graph_editor.py
```diff
@@ -36,6 +38,110 @@
...
+        :return: list of nodes that include this node as an output
+        """
+        children = []
+        for output_id in node.output:
+            children.extend(self._input_id_to_nodes[output_id])
+        return children
+
+    def add_node(self, node: NodeProto):
+        """
+        Adds the given node to the model and graph state
```
markurtz, line 121: add space below for documentation

src/sparseml/onnx/utils/graph_editor.py
```diff
@@ -36,6 +38,110 @@
...
+        self._input_id_to_nodes = defaultdict(list)
+        for node in self._model.graph.node:
+            self._store_node_edges(node)
+
+        # initializers
+        self._name_to_initializer = {
+            init.name: init for init in self._model.graph.initializer
+        }
+
+    def _store_node_edges(self, node: NodeProto):
```
markurtz, line 77: move private method declaration to after all public methods

markurtz (commented):
Small NITs, otherwise looks good

markurtz (dismissed):

markurtz (approved):

mgoin (approved):

PR #103, opened 2021-03-17
Yolo integration QAT compatibility fixes

bfineran:
PyTorch does not support QAT pickling or half precision. These fixes address those issues


markurtz (approved):
Just noting that I am getting worried about the complexity of the integration now and migrations in the future so we're being conscious of it, but I don't see any way around the additions right now

bfineran:
@markurtz agreed. The minimal integration for pruning is still a few lines. A lot of the changes introduced here are due to unsupported features in QAT that could be simplified based on how future pytorch releases look.

mgoin (approved):

PR #104, opened 2021-03-18
example Yolo deployment server

bfineran:
credits to @mgoin for the server infrastructure in his DeepSparse work

- [x] planned changes include E2E testing when sparsezoo models are available, quantized models, accuracy checks, etc

### DIFF
--- a/integrations/ultralytics/server/server.py
+++ b/integrations/ultralytics/server/server.py
@@ -1,0 +1,223 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Example script for hosting a Yolo ONNX model as a Flask server
+using the DeepSparse Engine as the inference backend
+
+##########
+Command help:
+usage: server.py [-h] [-s BATCH_SIZE] [-j NUM_CORES] [-a ADDRESS] [-p PORT]
+ onnx_filepath
+
+Host a Yolo ONNX model as a server, using the DeepSparse Engine and Flask
+
+positional arguments:
+  onnx_filepath         The full filepath of the ONNX model file or SparseZoo stub
+                        to the model
+
+optional arguments:
+  -h, --help            show this help message and exit
+  -s BATCH_SIZE, --batch_size BATCH_SIZE
+                        The batch size to run the analysis for
+  -j NUM_CORES, --num_cores NUM_CORES
+                        The number of physical cores to run the analysis on,
+                        defaults to all physical cores available on the system
+  -a ADDRESS, --address ADDRESS
+                        The IP address of the hosted model
+  -p PORT, --port PORT  The port that the model is hosted on
+
+##########
+Example command for running:
+python server.py \
+    ~/models/yolo-v3-pruned_quant.onnx
+"""
+
+import argparse
+import time
+from typing import List, Tuple
+
+import numpy
+import torch
+
+import cv2
+import flask
+from deepsparse import compile_model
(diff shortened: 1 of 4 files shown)


bfineran:
Server / Client sanity checked using latest unstructured sparse yolo models.  Client was able to retrieve accurate annotations for its image.

integrations/ultralytics/server/server.py
```diff
@@ -0,0 +1,267 @@
...
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description=(
+            "Host a Yolo ONNX model as a server, using the DeepSparse Engine and Flask"
+        )
+    )
+
+    parser.add_argument(
+        "onnx_filepath",
```
markurtz, line 71: for our standard args in other scripts, we use "-" as the separator, can we update for consistency along with the other params?

integrations/ultralytics/server/server.py
```diff
@@ -0,0 +1,267 @@
...
+        type=str,
+        help="The full filepath of the ONNX model file or SparseZoo stub to the model",
+    )
+
+    parser.add_argument(
+        "-s",
```
markurtz, line 77: I think -b may be more easily understood as the batch size, additionally it matches our shorthand

integrations/ultralytics/server/server.py
```diff
@@ -0,0 +1,267 @@
...
+        "--batch_size",
+        type=int,
+        default=1,
+        help="The batch size to run the analysis for",
+    )
+    parser.add_argument(
+        "-j",
```
markurtz, line 84: I think -c may be more easily understood as num cores, additionally it matches our shorthand

integrations/ultralytics/server/server.py
```diff
@@ -0,0 +1,267 @@
...
+    torch.Tensor([[10, 13], [16, 30], [33, 23]]),
+    torch.Tensor([[30, 61], [62, 45], [59, 119]]),
+    torch.Tensor([[116, 90], [156, 198], [373, 326]]),
+]
+_YOLO_V3_ANCHOR_GRIDS = [t.clone().view(1, -1, 1, 1, 2) for t in _YOLO_V3_ANCHORS]
+_YOLO_V3_OUTPUT_SHAPES = [80, 40, 20]
+_YOLO_V3_GRIDS = [_get_grid(grid_size) for grid_size in _YOLO_V3_OUTPUT_SHAPES]
+
+
+def arrays_to_bytes(arrays: List[numpy.array]) -> str:
```
markurtz, line 129: is this something more general we should add as a utility to sparseml?

integrations/ultralytics/server/server.py
```diff
@@ -0,0 +1,267 @@
...
+    for arr in arrays:
+        arr_dtype = bytearray(str(arr.dtype), "utf-8")
+        arr_shape = bytearray(",".join([str(a) for a in arr.shape]), "utf-8")
+        sep = bytearray("|", "utf-8")
+        arr_bytes = arr.ravel().tobytes()
+        to_return += arr_dtype + sep + arr_shape + sep + arr_bytes
+    return to_return
+
+
+def bytes_to_arrays(serialized_arr: str) -> List[numpy.array]:
```
markurtz, line 140: same for this one, I think it would be good to add a a utility to sparseml

integrations/ultralytics/server/server.py
```diff
@@ -0,0 +1,267 @@
...
+
+    batch = numpy.stack(images, image_size)  # shape: (batch_size, 3, (image_size))
+    return numpy.ascontiguousarray(batch)
+
+
+def _pre_nms_postprocess(outputs: List[numpy.ndarray]) -> torch.Tensor:
+    # postprocess and transform raw outputs into single torch tensor
+    processed_outputs = []
+    for idx, pred in enumerate(outputs):
+        pred = torch.from_numpy(pred)
```
markurtz, line 183: do we take a decent performance hit at all when loading torch tensors from numpy? Wondering if this is a ref copy or a memory copy ultimately and therefore if we should adapt the code to run in numpy rather than torch

markurtz: Ah, I guess we're limited to torch because of the import for non_max_suppression

markurtz (commented):
Small comments and questions

integrations/ultralytics/server/server.py
bfineran: added here: https://github.com/neuralmagic/deepsparse/pull/62

markurtz (approved):

mgoin (approved):

PR #108, opened 2021-03-19
Yolo benchmarking

bfineran:


markurtz (dismissed):

integrations/ultralytics/deepsparse/benchmark.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+def _load_file_paths(dataset_path: str) -> List[str]:
+    # from yolov5/utils/datasets.py
+    path = str(Path(dataset_path).absolute())  # os-agnostic absolute path
+    if '*' in path:
+        files = sorted(glob.glob(path, recursive=True))  # glob
+    elif os.path.isdir(path):
+        files = sorted(glob.glob(os.path.join(path, '*.*')))  # dir
+    elif os.path.isfile(v):
```
bfineran, line 158:
```suggestion
    elif os.path.isfile(path):
```

integrations/ultralytics/deepsparse/benchmark.py
```diff
@@ -0,0 +1,250 @@
...
+        files = [path]  # files
+    else:
+        raise Exception(f"ERROR: {p} does not exist")
```
bfineran, line 161:
```suggestion
        raise Exception(f"ERROR: {path} does not exist")
```

markurtz (dismissed):

integrations/ultralytics/deepsparse/benchmark.py
```diff
@@ -0,0 +1,248 @@
...
+    )
+
+    results = BenchmarkResults()
+    progress_bar = tqdm(total=args.num_iterations)
+    for iteration, batch in enumerate(data_loader):
+        iter_start = time.time()
+
+        # pre-processing
+        batch = [cv2.imread(filepath) for filepath in batch]
+        inputs = [preprocess_images(batch)]
```
bfineran, line 208:
```suggestion
        inputs = [preprocess_images(batch, fp32=not args.quantized_inputs)]
```

kevinaer (dismissed):

markurtz (dismissed):

bfineran:
Planned changes:
* move batching to dataloader
* torch GPU and ORT backend support

integrations/ultralytics/deepsparse/client.py
```diff
@@ -43,21 +43,19 @@ def detect(
...
+            run the detection model on. Number of images should equal model batch size
         :return: list of post-processed object detection results for each image including
             class label, likelihood, and bounding box coordinates
         """
         if not isinstance(images, List):
             images = [images]
-        images = [
-            cv2.imread(image) if isinstance(image, str) else image for image in images
-        ]
+        images = numpy.stack([load_image(image) for image in images])
```
mgoin, line 53: does `len` work on numpy arrays?

mgoin (dismissed):
LGTM just one possible nit

integrations/ultralytics/deepsparse/client.py
bfineran: yes, where does `len` factor in here?

mgoin: sorry, i made this on the line below and it selected this instead
```
print(f"Sending batch of {len(images)} images for detection to {self._url}")
```

integrations/ultralytics/deepsparse/benchmark.py
```diff
@@ -0,0 +1,430 @@
...
+from pathlib import Path
+from typing import Any, Iterable, List, Optional, Tuple, Union
+
+import numpy
+import onnx
+import onnxruntime
+import torch
+from tqdm.auto import tqdm
+
+import cv2
```
markurtz, line 108: Need to add cv2 to known third party for setup config and then run make style again

integrations/ultralytics/deepsparse/benchmark.py
```diff
@@ -0,0 +1,430 @@
...
+        "--fp16",
+        help=("Set flag to execute torch benchmark in half precision (fp16)"),
+        action="store_true",
+    )
+    parser.add_argument(
+        "--device",
+        type=_parse_device,
+        default="cpu",
+        help=(
+            "Device to benchmark the model with. Default is cpu. Non cpu benchmarking "
```
markurtz, line 220: Can we give an example for the torch use case here? Also, looks like this is the device id going off of the parse device function, can we support cpu and cuda for torch and default to first device? They could still control device outside of this through standard commands and I feel most users will spin this up on single gpu machines

markurtz (commented):
Left a few comments

integrations/ultralytics/deepsparse/client.py
bfineran: ah, thanks.  images is a list at this point so should be fine.

integrations/ultralytics/deepsparse/benchmark.py
bfineran: sounds good.  will make the torch default `0` if cuda is available and add some documentation

mgoin (approved):

markurtz (approved):

PR #110, opened 2021-03-23
quantized inputs optimization for qat exports

bfineran:
there are instances where we want quantized graphs to take uint8 inputs instead of regular preprocessed fp32 inputs.  this optimization pass enables this by updating the input tensor of a quantized graph and deleting initial `QuantizeLinear` node.

As an example, this saves redundant operations in YOLOv3 preprocessing since inputs don't need to be converted from int to float and then back in the first step of the model.

Before:
Screen Shot 2021-03-22 at 9 47 20 PM

After:
Screen Shot 2021-03-22 at 9 47 36 PM

### DIFF
--- a/src/sparseml/pytorch/utils/quantization/quantize_qat_export.py
+++ b/src/sparseml/pytorch/utils/quantization/quantize_qat_export.py
@@ -18,6 +18,7 @@ quantization aware training.
 """
 
 
+import logging
 from collections import defaultdict
 from copy import deepcopy
 from typing import Any, NamedTuple, Optional, Union
@@ -47,6 +48,9 @@ __all__ = [
     "quantize_torch_qat_export",
     "skip_onnx_input_quantize",
 ]
+
+
+_LOGGER = logging.getLogger(__name__)
 
 
 """
@@ -636,10 +640,20 @@ def _skip_input_quantize(model: ModelProto) -> Optional[str]:
                 continue
             input_idx = input_idx[0]
             graph.update_node_input(child_node, input_node.name, input_idx)
+            _LOGGER.debug(
+                f"set node with output id {child_node.output[0]} as initial node in "
+                "graph"
+            )
 
+    _LOGGER.debug(
+        f"deleting QuantizeLinear node(s) with output id(s): "
+        f"{[n.output for n in input_children]}"
+    )
     graph.delete_nodes(input_children)  # only contains references to the Quantize nodes
     graph.delete_unused_initializers()  # cleanup
     input_node.type.tensor_type.elem_type = 2  # fp32 -> uint8
+    _LOGGER.info("Model initial QuantizeLinear node(s) deleted and inputs set to uint8")
+
     return None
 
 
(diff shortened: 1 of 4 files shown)


src/sparseml/pytorch/utils/quantization/quantize_qat_export.py
```diff
@@ -593,3 +599,65 @@ def quantize_torch_qat_export(
         onnx.save(model, output_file_path)
 
     return model
+
+
+def _skip_input_quantize(model: ModelProto) -> bool:
+    if len(model.graph.input) != 1:
```
markurtz, line 605: NIT: remove if block, already covered by later one

src/sparseml/pytorch/utils/quantization/quantize_qat_export.py
```diff
@@ -593,3 +599,65 @@ def quantize_torch_qat_export(
...
+    ):
+        # more than 1 input or input is not FP32
+        return False
+
+    input_node = model.graph.input[0]
+    input_children = [
+        node for node in model.graph.node if input_node.name in node.input
+    ]
+    if not all(node.op_type == "QuantizeLinear" for node in input_children):
+        return False
```
markurtz, line 619: NIT: can we put a quick note here that we're failing because the next part after an input must be a quantized block

src/sparseml/pytorch/utils/quantization/quantize_qat_export.py
```diff
@@ -593,3 +599,65 @@ def quantize_torch_qat_export(
...
+    deleted. This enables quantize graphs to take quantized inputs instead of floats.
+
+    :param model: The model to convert, or a file path to it
+    :param output_file_path: File path to save the converted model to
+    :return: True if a modification was made. False otherwise
+    """
+    if isinstance(model, str):
+        model = onnx.load(model)
+
+    mod_made = _skip_input_quantize(model)
```
markurtz, line 658: I think it might be safer to raise an exception here that the onnx graph couldn't be converted rather than returning false. I'm worried people might miss checking that for if changes were made or not

markurtz (changes requested):
Left some comments, also can we add some logging into this at the debug and info levels? Finally, is there anything easy we can do here to enable some quick tests for this code to make sure it has code coverage?

src/sparseml/pytorch/utils/quantization/quantize_qat_export.py
bfineran: yes, that's fine for the way we have it implemented now, as a one off optimization that is only applied in cases where we know it is both wanted and will work (ie yolov3), if this is ever added to more generic flows or something that does multiple passes I'd avoid anything that would fail if there is an issue with the graph. same generally goes for why I didn't originally add warnings.

bfineran:
@markurtz added logging and two simple tests for this optimization

markurtz (approved):

natuan (approved):

PR #116, opened 2021-03-26
Show error when using keras2onnx with tf>=2.3

natuan:


src/sparseml/keras/utils/exporter.py
```diff
@@ -74,6 +75,16 @@ def export_onnx(
         if keras2onnx_import_error is not None:
             raise keras2onnx_import_error
 
+        if warn_on_tf_support:
+            import tensorflow
+            v = tensorflow.__version__
+            if v >= "2.3.0":
+                raise ValueError(
```
bfineran, line 82: nit: do we want to warn or raise an exception here?

markurtz: The flows currently don't work for most models in these versions, so preference is to raise an error to make sure the user sees that first and their first course of action to fix is ideally downgrading the tf version. We want to make sure they can consciously override it, though, to have it go through.

@natuan can we change the name from warn to raise to better signify what will happen?

Additionally in the error message let's change it to something along the lines of:
"Tensorflow version {} is greater than the currently supported version for keras2onnx. Please downgrade the Tensorflow version <2.3.0 or set raise_on_tf_support to False to continue."

src/sparseml/keras/utils/exporter.py
```diff
@@ -74,6 +75,16 @@ def export_onnx(
...
+                    "keras2onnx is known to cause errors for Tensorflow >= 2.3.0, "
+                    "and you have {}. Consider downgrade Tensorflow for model export "
+                    "to ONNX.".format(v)
```
markurtz, line 85: remove .format notation in place of f strings

markurtz (changes requested):

src/sparseml/keras/utils/exporter.py
```diff
@@ -74,6 +75,16 @@ def export_onnx(
         if keras2onnx_import_error is not None:
             raise keras2onnx_import_error
 
+        if raise_on_tf_support:
+            import tensorflow
+            v = tensorflow.__version__
+            if v >= "2.3.0":
+                raise ValueError(
+                    "Tensorflow version {} is greater than the currently supported "
```
markurtz, line 83: @natuan please migrate this to an f string instead of using .format

markurtz (changes requested):

markurtz (approved):

kevinaer (approved):

bfineran (approved):

PR #127, opened 2021-03-31
Retrain ResNet20 model to work with older TF versions

natuan:
Pending model upload before testing

### DIFF
--- a/integrations/keras/prune_resnet20.py
+++ b/integrations/keras/prune_resnet20.py
@@ -65,7 +65,29 @@ def download_model_and_recipe(root_dir: str):
     Download pretrained model and a pruning recipe
     """
     model_dir = os.path.join(root_dir, "resnet20_v1")
-    zoo_model = Zoo.load_model(
+
+    # Load base model to prune
+    base_zoo_model = Zoo.load_model(
+        domain="cv",
+        sub_domain="classification",
+        architecture="resnet_v1",
+        sub_architecture=20,
+        framework="keras",
+        repo="sparseml",
+        dataset="cifar_10",
+        training_scheme=None,
+        optim_name="base",
+        optim_category=None,
+        optim_target=None,
+        override_parent_path=model_dir,
+    )
+    base_zoo_model.download()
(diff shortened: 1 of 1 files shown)


bfineran (dismissed):

markurtz (dismissed):

bfineran (approved):

integrations/keras/prune_resnet20.py
```diff
@@ -74,16 +76,33 @@ def download_model_and_recipe(root_dir: str):
...
-    zoo_model.download()
-    model_file_path = zoo_model.framework_files[0].downloaded_path()
+    base_zoo_model.download()
+    model_file_path = base_zoo_model.framework_files[0].downloaded_path()
     if not os.path.exists(model_file_path) or not model_file_path.endswith(".h5"):
         raise RuntimeError("Model file not found: {}".format(model_file_path))
-    recipe_file_path = zoo_model.recipes[0].downloaded_path()
+
+    # Load recipe to use
+    pruned_zoo_model = Zoo.load_model(
```
bfineran, line 90: can we use `Zoo.download_recipe`

bfineran (changes requested):

integrations/keras/prune_resnet20.py
```diff
@@ -74,16 +76,33 @@ def download_model_and_recipe(root_dir: str):
...
+        framework="keras",
+        repo="sparseml",
+        dataset="cifar_10",
+        training_scheme=None,
+        optim_name="pruned",
+        optim_category="conservative",
+        optim_target=None,
+        override_parent_path=model_dir,
+    )
+    pruned_zoo_model.download()
```
markurtz, line 104: Why are we downloading the whole model if we're only using the recipe? Also, @kevinaer / @bfineran we added in a way to get the base model from a pruned model, right?

markurtz (changes requested):

markurtz (approved):

bfineran (approved):

PR #135, opened 2021-04-01
Fix incorrect loss scaling in DDP setup

eldarkurtic:
According to the PyTorch documentation, DDP syncs models by calculating mean of gradients across all processes.
Since the default reduction used by almost all PyTorch loss functions is `mean`, this loss scaling would result with `world_size` times larger gradients and the DDP experiment won't be mathematically equivalent to the corresponding one-process experiment.


bfineran (approved):
thanks for the catch @eldarkurtic

markurtz (approved):

PR #138, opened 2021-04-02
Fix minor typos

eldarkurtic:


bfineran (changes requested):
Thank you for the PR @eldarkurtic.  The docstring at the top of this file is generated from running the `--help` option for each of the comands.  To that end, could you also update the `argparse` texts [here](https://github.com/neuralmagic/sparseml/blob/main/scripts/pytorch_vision.py#L565) and [here](https://github.com/neuralmagic/sparseml/blob/main/scripts/pytorch_vision.py#L590) to match your updates to the docstring?

eldarkurtic:
Sorry, completely missed those.

bfineran (approved):
No problem. LGTM!

kevinaer (approved):

PR #142, opened 2021-04-03
PyTorch Global Magnitude Pruning

bfineran:
enable global magnitude pruning by adding `global_sparsity: True` to GMPruningModifier configs

* [x] refactor mask creators and masking objects to process lists of tensors instead of single tensors
* [x] add option for global masking in mask creators and pruning modifier
* [x] testing

### DIFF
--- a/tests/sparseml/pytorch/optim/test_modifier_pruning.py
+++ b/tests/sparseml/pytorch/optim/test_modifier_pruning.py
@@ -17,7 +17,12 @@ import os
 import pytest
 import torch
 
-from sparseml.pytorch.optim import ConstantPruningModifier, GMPruningModifier
+from sparseml.pytorch.optim import (
+    ConstantPruningModifier,
+    GlobalMagnitudePruningModifier,
+    GMPruningModifier,
+    MagnitudePruningModifier,
+)
 from tests.sparseml.pytorch.helpers import LinearNet
 from tests.sparseml.pytorch.optim.test_modifier import (
     ScheduledModifierTest,
@@ -214,6 +219,15 @@ def test_constant_pruning_yaml():
             inter_func="cubic",
             global_sparsity=True,
         ),
+        lambda: GlobalMagnitudePruningModifier(
+            params="__ALL_PRUNABLE__",
+            init_sparsity=0.05,
+            final_sparsity=0.95,
+            start_epoch=10.0,
+            end_epoch=25.0,
+            update_frequency=1.0,
+            inter_func="cubic",
+        ),
         lambda: GMPruningModifier(
             params=["seq.fc1.weight", "seq.fc2.weight"],
             init_sparsity=0.05,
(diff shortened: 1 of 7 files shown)


bfineran:
Current status is this PR will be on hold until other algorithms are landed on main

bfineran:
Putting this back up for review

kevinaer (dismissed):

markurtz (dismissed):

markurtz:
Overall looks good, 2 notes for the future to keep an eye on:

- I'm worried about the memory hit we're going to take on creating a new tensor the size of all the params that are used to just calculate the threshold. This may be something we need to do on the CPU, or at least have an option for, instead of keeping the tensor on the current device
- Also worried about the discoverability of global magnitude. As we're doing the refactor and renaming, it may make sense to break out GradualMagnitude and GlobalMagnitude into their own root level classes that extend the same base class so it's more discoverable and readily available in docs and tutorials

natuan (dismissed):
LGTM. For future iteration, I wonder if the `create_sparsity_masks` could be simplified by always create mask with the defined sparsity on top of all tensors passed in. The distinction between global and layer-wise would be made from the call-sites by either passing a tensor list, or calling this func on separate tensors (there's a burden for layer wise here, but might worth it to simplify the create mask code)

bfineran:
@markurtz updated to include child classes for pruning modifiers, note that we will have to be careful about `ModifierProp` serialization when inheriting from a class with more props.  Agreed on the memory issue, definitely something to keep an eye out for.

@natuan That implementation would move logic for when to group globally or not to two different places (the call site would have to pass in tensors separately), as implemented, the only place in the stack that cares about global or not, other than propagating, is the mask creator.  I do think adding support for a single tensor input would be good though.

markurtz (dismissed):

tests/sparseml/pytorch/optim/test_modifier_pruning.py
```diff
@@ -324,11 +344,83 @@ def test_gm_pruning_yaml():
...
+        str(yaml_modifier.global_sparsity)
+        == str(serialized_modifier.global_sparsity)
+        == str(obj_modifier.global_sparsity)
+    )
+
+
+@pytest.mark.skipif(
+    os.getenv("NM_ML_SKIP_PYTORCH_TESTS", False),
+    reason="Skipping pytorch tests",
+)
```
kevinaer, line 397: Can you remove the skips? I am not sure if it is being used in some legacy job but environment based skipping is no longer being used in favor of make targets.

tests/sparseml/pytorch/optim/test_modifier_pruning.py
```diff
@@ -340,7 +432,85 @@ def test_gm_pruning_yaml():
...
+        str(yaml_modifier.mask_type)
+        == str(serialized_modifier.mask_type)
+        == str(obj_modifier.mask_type)
+    )
+
+
+@pytest.mark.skipif(
+    os.getenv("NM_ML_SKIP_PYTORCH_TESTS", False),
+    reason="Skipping pytorch tests",
+)
```
kevinaer, line 477: Can you remove the skips? I am not sure if it is being used in some legacy job but environment based skipping is no longer being used in favor of make targets.

kevinaer (changes requested):

bfineran:
@kevinaer resolved, thanks!

kevinaer (approved):

markurtz (approved):

Issue #152, opened 2021-04-06
Memory leak when using debug-steps


eldarkurtic:
**Describe the bug**
There is a memory leak when using `--debug-steps`. RAM fills up quickly, especially when using larger number of workers (for example 20). 

**Expected behavior**
Having a run without out of memory issues.

**Environment**
Include all relevant environment information:
1. OS: Debian 4.19.171-2 (2021-01-30) x86_64
2. Python version: 3.8
3. SparseML version or commit hash: [b82577](https://github.com/neuralmagic/sparseml/commit/b82577205f0832dc5f9d42bafc050ec46515af24)
4. ML framework version(s): torch 1.7.1

**To Reproduce**
Run:
```bash
CUDA_VISIBLE_DEVICES=0,1,2,3 python scripts/pytorch_vision.py train --recipe-path ./dummy.yaml --arch-key resnet50 --dataset imagenet --dataset-path ~/ILSVRC --train-batch-size 256 --test-batch-size 512 --loader-num-workers 20 --debug-steps 1
```
where `dummy.yaml` is:
```yaml
modifiers:
  - !EpochRangeModifier
    start_epoch: 0.0
    end_epoch: 90.0
```

**Errors**
RuntimeError: DataLoader worker (pid <worker_pid>) is killed by signal: Killed.

PR #153, opened 2021-04-06
Fix memory leak when using debug-steps

eldarkurtic:
Memory leak seems to be related to the way how tqdm is wrapping data_loader.
More info at: https://github.com/tqdm/tqdm/issues/746
This PR fixes https://github.com/neuralmagic/sparseml/issues/152


bfineran (approved):
Thanks for the fix @eldarkurtic

markurtz (approved):

PR #159, opened 2021-04-07
Fix incorrect epoch in the checkpoint

eldarkurtic:
The checkpoint created at the end of training is saved with the wrong epoch number because of the `epoch += 1` in the training loop.


bfineran (approved):
great catch @eldarkurtic!

markurtz (approved):

PR #164, opened 2021-04-08
option to enable QAT before training begins

bfineran:
With `enable_on_initialize` set in a quantization modifier, QAT ops will be injected into the graph as soon as the recipe is initialized making it easier to load weights to/from QAT models.

An additional nicety is that the flow for applying QAT to any model can now be simplified to initializing a `QuantizationModifier`. torchvision ResNet50 example:

```python

# load model
model = resnet50(False)

# apply quantization
qmod = QuantizationModifier(enable_on_initialize=True)
qmod.initialize(model, None)
```

### DIFF
--- a/src/sparseml/pytorch/optim/modifier_quantization.py
+++ b/src/sparseml/pytorch/optim/modifier_quantization.py
@@ -77,6 +77,9 @@ class QuantizationModifier(ScheduledModifier):
         exception. For compatibility with YAML serialization only.
     :param model_fuse_fn_kwargs: dictionary of keyword argument values to be passed
         to the model fusing function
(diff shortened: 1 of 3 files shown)


natuan (dismissed):

markurtz:
@bfineran I'm wondering if there's something more generic we can build into the framework to enable an initialize and finalize on the modifier/manager. My thoughts on this are to push a generic setup that any modifier can support. In addition, something like finalize might allow us to apply to an EMA model afterward if it were to optionally take in a state dict

markurtz (approved):

kevinaer (approved):

natuan (approved):

PR #168, opened 2021-04-10
Base framework and sparsification implementation for phase 2

markurtz:
Base functionality for phase2 to enable loading the information available for training, sparsifying, and exporting for a given ML framework


kevinaer (dismissed):

src/sparseml/sparsification/info.py
```diff
@@ -0,0 +1,304 @@
...
+        create_parent_dirs(path)
+
+        with open(path, "w") as file:
+            file.write(info.json())
+
+        _LOGGER.info(
+            "saved sparsification info for framework %s in file at %s", framework, path
+        ),
+    else:
+        print(info.json())
```
kevinaer, line 244: Same as framework

src/sparseml/framework/info.py
```diff
@@ -0,0 +1,299 @@
...
+        create_parent_dirs(path)
+
+        with open(path, "w") as file:
+            file.write(info.json())
+
+        _LOGGER.info(
+            "saved framework info for framework %s in file at %s", framework, path
+        ),
+    else:
+        print(info.json())
```
kevinaer, line 240: While functionality wise this works would it be worth using pretty print so the output is more readable?

bfineran (approved):
LGTM. the pydantic models look like they will be good for solidifying the Modifiers

kevinaer (approved):

PR #175, opened 2021-04-13
fix pickling in torch pruning hooks

bfineran:
Pruning hook functions were recently moved to be nested functions that would be returned for a given parameter idx to support global pruning.  This causes issues with python pickling because nested functions cannot be pickled.  This causes issues when `torch.save` is directly called on a pruned model.

Note that best practice is not to pickle entire pytorch models, but to just serialize their state dicts.

PyTorch tests passing. Some sample code to directly test pickling (fails without this change):

```python
import torch
from sparseml.pytorch.optim import GMPruningModifier

model = resnet50(False)
mod = GMPruningModifier(0.5, 0.9, 0.0, 10.0, 0.1, "__ALL_PRUNABLE__", leave_enabled=True)
mod.initialize(model, None)
mod.scheduled_update(model, None, 1, 1)
torch.save(model, "tmp.pt")  # fails without this PR
```


markurtz (dismissed):
Looks good, I like the use of partial here

dalistarh:
Looks good to me.

markurtz (approved):

kevinaer (approved):

PR #177, opened 2021-04-15
topologically sort ONNX nodes

bfineran:
@mgoin correctly noted that our quantized ONNX graph exports do not have topologically sorted nodes which violates the ONNX model proto spec and will cause an ONNX checker error. This PR adds a helper function to correct models after optimizations are made.


kevinaer (approved):

mgoin (approved):
very nice, this'll be useful. thanks for the test

PR #178, opened 2021-04-19
Fix issues with nighly tests failing due to different package names for sparseml

markurtz:


bfineran (approved):

mgoin (approved):
LGTM but I still am not happy with `:type`, there has to be a better way for docs

PR #179, opened 2021-04-19
Github Actions for tests

kevinaer:
- Created github action for pull requests that runs if sparseml source code was changed as well as different framework targets if the directories were changed. Will run all tests if setup.py is changed.
- Created github action for merge that runs all sparseml tests. Uses the sparsezoo branch with the matching name for installation

Note: onnx tests are currently failing because MNIST download link is down.


markurtz (approved):

mgoin (approved):
looking good

PR #183, opened 2021-04-20
Change version creation code for nighlies so that package version and version within the code can match

markurtz:


src/sparseml/version.py
```diff
@@ -16,16 +16,34 @@
 Functionality for storing and setting the version info for SparseML
 """
 
+from datetime import date
+
+
+_BASE_VERSION = "0.2.0"
+_NIGHTLY = True  # change to False to turn off the build version for releases
```
bfineran, line 23: if this is only used with `make build`, can we propagate the nightly build param here?

bfineran (commented):
if installed from source, could we see any inconsistencies with the date in `sparseml.__version__`?

src/sparseml/version.py
dhuangnm: Similar to Ben's question, currently we build nightly with following command:
    make build BUILD_ARGS=nightly

Is the change here to remove the need of BUILD_ARGS=nightly and only relying on changing _NIGHTLY in the file? I'd prefer keeping the old way though so it'll be able to build both regular and nightly builds without changing any code.

markurtz: The nightly is still needed in the setup.py because we change the package name based on that. 

This file is compiled into the source code as-is and therefore cannot rely on build args. A user running the code locally wouldn't have that nightly flag set for any of the code running and we would create the wrong version. In that way, we need to change this file in the source code that goes into the package so it matches what is distributed as the nightly version. The only two ways I think we have for going this is either changing it in git for the release branches or having jenkins override before it creates the package.

Another benefit of setting this arg to True by default is that if a user runs into a problem locally we know it's local or nightly and not on a release by checking the version string. 

We can rename the arg if we feel there is a more appropriate name.

dhuangnm: So if I understand correctly, after this change:
  ==>to build nightly builds: 
               make build BUILD_ARGS=nightly
  ==> to build regular builds:
               modify _NIGHTLY to False in version.py
               make build
Is this correct?

markurtz: @dhuangnm I pushed up another change to make this step a little bit smoother and remove nightly altogether. Now you can set is_release in version.py to True for release builds and set to False for nightly (the default)

setup.py
```diff
@@ -13,27 +13,19 @@
...
 
 import os
-import sys
-from datetime import date
 from typing import Dict, List, Tuple
 
 from setuptools import find_packages, setup
 
 
+is_release = False
```
dhuangnm, line 21: Mark, can is_release here read from version.py so that we don't keep it in two places? i.e. we only need to modify it once in version.py for nightly or regular builds.

markurtz: this does read from version.py, I just put a default in here so that code editors don't complain about an unamed variable. Once the later exec version.py file runs it overrides the is_release variable

markurtz: @dhuangnm, I've added a bit more documentation in setup.py to make this apparent

dhuangnm (commented):
Thanks Mark. So now by default "make build" builds a nightly version and if to build regular builds we just modify is_release to True in version.py and still run "make build" command, right? If so it looks good to me.

markurtz:
> Thanks Mark. So now by default "make build" builds a nightly version and if to build regular builds we just modify is_release to True in version.py and still run "make build" command, right? If so it looks good to me.

Yep, this is correct with this implementation now

bfineran (approved):

dhuangnm (approved):
Great, thanks.

PR #196, opened 2021-04-27
Fix non-existing stubs

eldarkurtic:
The fixed ones are taken from here: [https://docs.neuralmagic.com/sparsezoo/source/models.html](https://docs.neuralmagic.com/sparsezoo/source/models.html)

### DIFF
--- a/integrations/timm/README.md
+++ b/integrations/timm/README.md
@@ -44,11 +44,11 @@ pip install sparseml[torchvision]
 from pytorch-image-models to include a `sparseml-recipe` argument
 to run SparseML optimizations with.  This can be a file path to a local
 SparseML recipe or a SparseZoo model stub prefixed by `zoo:` such as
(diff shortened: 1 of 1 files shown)


eldarkurtic:
I have just noticed that SparseZoo state_dicts are not compatible with the expected state_dicts for `torchvision` models, so runs with these stubs will crash with `RuntimeError: Error(s) in loading state_dict for ResNet`. 
`torchvision` models expect something like this:

- conv1.weight
- layer1.0.conv1.weight
- fc.weight

and SparseZoo models have something like this:

- input.conv.weight
- sections.0.0.conv1.weight
- classifier.fc.weight

bfineran (commented):
The rwightman models are still under development. Thanks for pointing this out. Unfortunately the sparseml pytorch stub weights will not load into the rwightman model

PR #199, opened 2021-04-27
Update annotate.py

jeanniefinks:
minor edits, thx


bfineran (changes requested):
thanks @jeanniefinks for these fixes.  The changes to the `help=` strings should also be included in the help text at the top of the file since the help text is generated from these `help=` strings

jeanniefinks (commented):
Made updates; think I got them all.

NMsasa (approved):

bfineran (approved):

PR #203, opened 2021-04-28
setter function for quant enable_on_initialize

bfineran:
required for overriding quant recipe model load


natuan (dismissed):

src/sparseml/pytorch/optim/modifier_quantization.py
```diff
@@ -218,6 +218,14 @@ def enable_on_initialize(self) -> bool:
         """
         return self._enable_on_initialize
 
+    @enable_on_initialize.setter
+    def enable_on_initialize(self, value: bool):
+        """
+        :params value: Epoch to stop the tracking of batch norm stats. Set
+        None to not stop tracking batch norm stats during QAT
```
markurtz, line 225: need an indent on this next line for docs to compile correctly

markurtz (changes requested):

src/sparseml/pytorch/optim/modifier_quantization.py
bfineran, line 225:
```suggestion
            None to not stop tracking batch norm stats during QAT
```

bfineran:
@markurtz updated

src/sparseml/pytorch/optim/modifier_quantization.py
```diff
@@ -218,6 +218,14 @@ def enable_on_initialize(self) -> bool:
         """
         return self._enable_on_initialize
 
+    @enable_on_initialize.setter
+    def enable_on_initialize(self, value: bool):
+        """
+        :params value: Epoch to stop the tracking of batch norm stats. Set
+            None to not stop tracking batch norm stats during QAT
```
bfineran, line 225:
```suggestion
        :params value: True if QAT should be enabled when this modifier is initialized,
            False otherwise. If true, start_epoch will not be respected
```

markurtz (approved):

mgoin (approved):

PR #206, opened 2021-04-29
Fix training args passed into load optim

natuan:
Currently in `integrations/transformers` we passed a TrainingArgs default instance into `load_optimizer`, making argument such as --learning_rate always use the default value instead of a custom value. This change is to fix that.


bfineran (approved):
good catch!

markurtz (approved):

spacemanidol (approved):

PR #210, opened 2021-05-01
Fixing usage of deepsparse vnni field

kevinaer:


bfineran (changes requested):
the error here may actually be that the field we're looking for should be `vnni` instead of `vni`.

https://github.com/neuralmagic/deepsparse/blob/main/src/deepsparse/cpu.py#L51

if this is the case we'd also have to fix this line in the deepsparse framework info: https://github.com/neuralmagic/sparseml/blob/main/src/sparseml/deepsparse/framework/info.py#L119

kevinaer:
Great catch Ben, completely missed the spelling issue as well!

bfineran (approved):

markurtz (approved):

PR #221, opened 2021-05-06
[cherry-pick] fix propagation of mask creator objects in PT pruning modifier (#220)

bfineran:
fixes bug that defaults pytorch mask creators to be 'unstructured' in certain training flows

updates release version to 0.3.1 for this hotfix

### DIFF
--- a/src/sparseml/pytorch/optim/modifier_pruning.py
+++ b/src/sparseml/pytorch/optim/modifier_pruning.py
@@ -608,6 +608,7 @@ class GMPruningModifier(ScheduledUpdateModifier):
             layers,
             param_names,
             layer_names=layer_names,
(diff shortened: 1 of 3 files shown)


dhuangnm:
Can we also add a test for it so we can verify things are working?

bfineran:
@dhuangnm the test file is updated in this PR to check that the correct mask creator is set

dhuangnm:
> @dhuangnm the test file is updated in this PR to check that the correct mask creator is set

oh ok, cool thanks.

dhuangnm:
Can we update the version.py to 0.3.1 for the cherry-pick?

markurtz (approved):

kevinaer (approved):

PR #230, opened 2021-05-11
Update New Slack

jeanniefinks:
added better community link


bnellnm (approved):
Seems ok to me.  Are those special links ever expected to change?

jeanniefinks (commented):
Good question @bnellnm. I had set it to never expire.
And it works with external users who may already have set up an account with us on our Slack workspace (e.g., logged out and need to log in again)

bfineran (approved):

PR #231, opened 2021-05-12
Refactor to enable support for Amp in PyTorch and better support future phase2 work

markurtz:


src/sparseml/pytorch/optim/manager.py
```diff
@@ -87,67 +265,147 @@ def from_yaml(
...
+        """
+        modifiers_index = {mod.identifier(): mod for mod in self.modifiers}
+
+        if strict:
+            modifier_keys = {key for key in modifiers_index.keys()}
+            state_dict_keys = {key for key in state_dict.keys()}
+            diff = modifier_keys.symmetric_difference(state_dict_keys)
+            if diff:
+                raise IndexError(
+                    f"Found extra keys: {state_dict_keys - modifier_keys} "
```
bfineran, line 292: nice

src/sparseml/optim/modifier.py
```diff
@@ -342,6 +348,99 @@ def yaml_key(clazz, framework: Union[str, None] = None):
 
         return "!{}.{}".format(framework, clazz.__name__)
 
+    @staticmethod
+    def comparator(one, two) -> int:
```
bfineran, line 347: should get a test for this in. am ok with later as tech debt since this req doesn't seem immediately breaking

src/sparseml/pytorch/optim/modifier_pruning.py
```diff
@@ -77,117 +79,95 @@ def _log_sparsity(
...
-                )
-        for name in mask_names:
-            if name not in state_dict:
-                raise RuntimeError(
-                    f"Mask parameter name {name} not found in state dict"
-                )
+        mask_names = {key for key in self._module_masks.names}
+        state_dict_keys = {key for key in state_dict.keys()}
+        diff = mask_names.symmetric_difference(state_dict_keys)
+        if diff:
```
bfineran, line 162: should check `strict` here

src/sparseml/pytorch/optim/modifier.py
```diff
@@ -115,18 +127,80 @@ def loggers(self):
...
-        :param module: module to modify
-        :param optimizer: optimizer to modify
+        :param state_dict: dictionary object as generated by this object's state_dict
+            function
+        :param strict: True to raise a KeyError for any missing or extra information in
+            the state dict, False to ignore
+        :raises IndexError: If any keys in the state dict do not correspond to a valid
+            index for this manager and strict=True
+        """
+        pass
```
bfineran, line 147: instead of `pass` I think the minimal functionality here should be checking for `strict` and raising an index error if anything is present.

src/sparseml/pytorch/optim/manager.py
```diff
@@ -32,7 +31,185 @@
 from sparsezoo.objects import Recipe
 
 
-__all__ = ["ScheduledModifierManager", "load_manager"]
+__all__ = ["ModifierManagerOptimWrapper", "ScheduledModifierManager"]
+
+
+class ModifierManagerOptimWrapper(object):
```
bfineran, line 37: nit: not sure if this name makes it clear what the class' purpose is. could we change the name to something like `TrainingStepRecipeWrapper` to singal that we're wrapping things that aren't just solo optimizers?

src/sparseml/pytorch/optim/manager.py
```diff
@@ -87,67 +265,147 @@ def from_yaml(
...
+            in place of the original optimizer since it doesn't always call into
+            the optimizer.step() function.
+        :param epoch: Optional epoch that can be passed in to start modifying at.
+            Defaults to the epoch that was supplied to the initialize function.
+        :return: A wrapped optimizer object. The wrapped object makes all the
+            original properties for the wrapped object available so it can be
+            used without any additional code changes.
+        """
+        if not self.initialized:
+            raise RuntimeError("manager must be initialized first")
```
bfineran, line 368: can we call into initialize instead of failing here? an implicit call to initialize would save an extra line in standard integrations.

src/sparseml/pytorch/optim/modifier_pruning.py
```diff
@@ -77,117 +79,95 @@ def _log_sparsity(
             )
 
 
-@PyTorchModifierYAML()
-class ConstantPruningModifier(ScheduledModifier):
+class _PruningParamsModifier(ScheduledUpdateModifier):
```
bfineran, line 82: would it make sense to add a `finalize` method that checks if the pruning masks are enabled, and if so, calls `apply` one last time and then disables them?

bfineran: looks like you read my mind

src/sparseml/pytorch/optim/modifier_pruning.py
```diff
@@ -212,48 +192,74 @@ def params(self, value: Union[str, List[str]]):
...
+
+        :param module: The model/module to finalize the modifier for.
+            Marked optional so state can still be cleaned up on delete,
+            but generally should always be passed in.
+        :param reset_loggers: True to remove any currently attached loggers (default),
+            False to keep the loggers attached.
+        :param kwargs: Optional kwargs to support specific arguments
+            for individual modifiers.
+        """
+        super().finalize(module, reset_loggers, **kwargs)
```
bfineran, line 259: let's call `apply()` one more time if the masks are enabled

bfineran (changes requested):
Looks overall good to me. Thanks for getting this bug resolved while also improving the codebase. Since `ScheduledOptimizer` is deprecated, could we update one of our examples to use these changes or update this PR with a brief example of the new flow for AMP?

Also are there any docs that need updating?

src/sparseml/optim/modifier.py
markurtz: Agree on the test, but let's push that off for later -- there are some integration tests we'll need to do between pruning and quantization to make sure they order correctly based on this

src/sparseml/pytorch/optim/manager.py
markurtz: Will rename to RecipeManagerStepWrapper in prep for renaming the Managers to RecipeManager for phase2

src/sparseml/pytorch/optim/manager.py
markurtz: yes, will default to calling into initialize if not called. Will also work out well for TF and keras

src/sparseml/pytorch/optim/modifier.py
markurtz: good point, fixing

src/sparseml/pytorch/optim/modifier_pruning.py
markurtz: :)

src/sparseml/pytorch/optim/modifier_pruning.py
markurtz: Because of the way we load the masks, strict is the only option, otherwise, we'll have to fail. I put a note in the docs for this

src/sparseml/pytorch/optim/modifier_pruning.py
markurtz: done

markurtz:
> Looks overall good to me. Thanks for getting this bug resolved while also improving the codebase. Since `ScheduledOptimizer` is deprecated, could we update one of our examples to use these changes or update this PR with a brief example of the new flow for AMP?
> 
> Also are there any docs that need updating?

I'm holding off on changing over examples/docs for a minor point release. So, trying to keep everything originally functioning and enabling this silently on the backend for the ultralytics integrations to unblock those for now. On 0.4 we'll update examples and docs

bfineran (dismissed):

src/sparseml/pytorch/optim/manager.py
```diff
@@ -32,7 +31,185 @@
...
+                self._wrapped_steps_per_epoch,
+            )
+            self._wrapped_manager.optimizer_pre_step(
+                self._wrapped_module,
+                self._wrapped_optimizer,
+                self._wrapped_epoch,
+                self._wrapped_steps_per_epoch,
+            )
+
+        ret = self._wrapped.step(*args, **kwargs)
```
natuan, line 175: Can you clarify what the difference between the self._wrapped and the self._wrapped_optimizer? In the context of amp support, what would we pass in for the _wrapped?

src/sparseml/optim/modifier.py
```diff
@@ -342,6 +343,99 @@ def yaml_key(clazz, framework: Union[str, None] = None):
...
+            return 1
+        if one_start > two_start:
+            # if one starts after two, return after
+            return 1
+        return -1
+
+    @staticmethod
+    def comparator_identifiers(one, two) -> int:
+        """
+        Comparator implementation for Modifiers based on end_epoch.
```
natuan, line 423: based on identifiers

src/sparseml/pytorch/optim/manager.py
markurtz: Wrapped is what we override the step call for. Wrapped optimizer is the pytorch optimizer implementation needed to be passed through to the modifiers

natuan: I saw we passed the same optimizer into both the _wrapped and _wrapped_optimizer (in optimizer.py). This is not always the case, right? (otherwise, we could just use _wrapped_optimizer)

natuan (approved):

PR #245, opened 2021-05-18
Fix mismatched versions

### LINKED ISSUE #244: Mismatched lib versions
**Describe the bug**
Because of the `sparsezoo~=_VERSION_MAJOR_MINOR` pip installation picks the latest minor version, instead of the latest patch version. This causes some incompatibilities between installed versions of `sparseml` and `sparsezoo`. 

**Expected behavior**
When `__version__ = "0.2.0"` is specified, to get `sparsezoo==0.2.0` and not `sparsezoo==0.3.1`.

eldarkurtic:
This PR fixes https://github.com/neuralmagic/sparseml/issues/244 by adding the patch version in the `setup.py` installation process. 
Before this fix, specifying `version_base = "0.2.0"` would call `pip install sparsezoo~=0.2`, and this would pick the latest minor version, i.e. `sparsezoo==0.3.1`. The installed libs would be in the broken state, since `sparseml==0.2.0` and `sparsezoo==0.3.1` are not compatible.
With this fix, `pip install sparsezoo~=0.2.0` is called and the latest patch version, i.e. `>=0.2.0` and `<0.3`, is installed.


markurtz:
Thanks @eldarkurtic, this was a planned fix for the 0.4 release we're coming out with at the end of this month. I'm going to make adjustments to this diff as there are changes we want to do in addition. Specifically, we don't want to remove major_minor completely and we should not be using major_minor_bug for the docs

markurtz (approved):

kevinaer (approved):

PR #263, opened 2021-05-28
PyTorch refactor pruning algorithms with pruning scorer object

bfineran:
Introduces the `PruningParamsScorer` class to handle logic specific to the ranking of parameters to prune.  With this change, the `ModuleParamPruningMask` now makes calls into the scorer to determine which scores should be passed into the mask creator.  This change will make future implementations of new pruning algorithms more modular and simple to implement.  Scorers are implemented for magnitude, movement, and M-FAC pruning.


markurtz (dismissed):
New structure looks good

markurtz (dismissed):

markurtz (approved):

natuan (approved):

PR #283, opened 2021-06-11
non-parameterized torch QAT MatMul Module

bfineran:
### QATMatMul Module
To help in efforts of quantizing more parts of transformer graphs, this module is to be used in place of `torch.matmul` and will export to ONNX (before quantization conversion) as follows:
Screen Shot 2021-06-11 at 4 47 08 PM

This graph will be able to be converted to a single `QLinearMatMul` op with additional post processing to be added in a future PR.


natuan (dismissed):

markurtz:
Talked with @bfineran and @natuan on this, I'm concerned about two points on this direction:
- requiring sparseml for the model definition now which we've been avoiding as much as possible previously
- a design pattern that is more flexible for the future of including more functional operators for quantization

markurtz (changes requested):
As discussed for next release (0.7) we'll go through and implement a more generic version

bfineran:
closing in favor of #353

PR #292, opened 2021-06-16
define final_sparsity by param in torch pruning modifiers

bfineran:
this change will enable a user to define multiple target sparsities for different parameters within a single pruning modifier. the specific use for this is so a single modifier can prune to multiple levels for parameters while sharing state between them in that same modifier object. This is important for pruning methods such as WoodFisher where multiple sparsity targets are necessary and sharing state between all parameters saves computational overhead.

example:
```yaml
- !GMPruningModifier
  params: []  # to be read from final_sparsity
  final_sparsity: {
      0.8: ['param_1', 'param_2'],
      0.9: ['param_3', ...],
      ...
    }
  ...
```

### DIFF
--- a/src/sparseml/pytorch/optim/mask_creator_pruning.py
+++ b/src/sparseml/pytorch/optim/mask_creator_pruning.py
@@ -61,13 +61,18 @@ class PruningMaskCreator(ABC):
 
     @abstractmethod
     def create_sparsity_masks(
-        self, tensors: List[Tensor], sparsity: float, global_sparsity: bool = False
+        self,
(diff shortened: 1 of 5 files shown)


markurtz:
@bfineran overall changes look good, we'll need to get it resolved with main, though. Specifically the phased pruning impl that was landed

bfineran:
merged into main - tests for phased pruning and multi-final sparsity both passing

rahul-tuli (approved):
Looks good!

markurtz (approved):

PR #294, opened 2021-06-17
separate layer-wise and global WoodFisher pruning modifiers

bfineran:
#292 adds support for a single modifier to prune layers to different levels. This makes it simple to add layer-wise support for WoodFisher/MFAC pruning. This PR sets `global_sparsity` to False in the default `MFACPruningModifier` so that layer wise sparsities may be defined and adds a specific `MFACGlobalPruningModifier` as a convenience for global WF pruning.

Note that all parameters should still be under a single `MFACPruningModifier` when using layer wise, and sparsity groups should be defined for each parameter within the `final_sparsity` arg.

### DIFF
--- a/src/sparseml/pytorch/optim/modifier_pruning.py
+++ b/src/sparseml/pytorch/optim/modifier_pruning.py
@@ -825,6 +825,7 @@ class GMPruningModifier(_PruningParamsModifier):
         self._pre_step_completed = True
 
         if started:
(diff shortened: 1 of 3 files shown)


markurtz (dismissed):

rahul-tuli (dismissed):

markurtz (approved):

rahul-tuli (approved):
LGTM!

PR #305, opened 2021-06-23
One shot MFAC pruning support

bfineran:
One shot pruning support for WoodFisher / MFAC. This requires extra support because the method needs a certain number of gradient samples to prune with.  Support is added by offering `dataloader` and `loss_fn` kwargs that will be used to obtain the required samples.

Testing with dummy data/loss function passes through the MFAC pipeline:
Recipe contains - 
```yaml
  - !MFACGlobalPruningModifier
    params: "__ALL_PRUNABLE__"
    init_sparsity: 0.0
    final_sparsity: *target_sparsity
    start_epoch: *pruning_start_epoch
    end_epoch: *pruning_end_epoch
    update_frequency: *pruning_update_frequency
    mask_type: *pruning_mask_type
```

Sample code - 
```python
from sparseml.pytorch.optim import ScheduledModifierManager
from sparseml.pytorch.utils import GradSampler
import torch

RECIPE_PATH = ""

r = resnet50(False)
manager = ScheduledModifierManager.from_yaml(RECIPE_PATH)

dataloader = ((torch.randn(10,3,224,224), 0) for _ in range(200))
loss_fn = lambda x, _: torch.mean(x)
grad_sampler = GradSampler(dataloader, loss_fn)

manager.apply(r, grad_sampler=grad_sampler)
```

### DIFF
--- a/src/sparseml/pytorch/utils/mfac_helpers.py
+++ b/src/sparseml/pytorch/utils/mfac_helpers.py
@@ -26,6 +26,7 @@ from typing import (
     Callable,
     Dict,
     Generator,
+    Iterable,
     Iterator,
     List,
     Optional,
@@ -57,7 +58,7 @@ class GradSampler:
     Class for computing gradient samples for a Model given a sample data loader and
     loss function.
 
-    :param dataloader: iterator of data samples to use as model inputs and their loss
+    :param data_loader: iterator of data samples to use as model inputs and their loss
         targets. Samples can either be single tensors as model input or a list of
         inputs and should be iterated in tuples with their targets
     :param loss_fn: function to be called on model outputs to compute the loss at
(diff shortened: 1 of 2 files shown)


src/sparseml/pytorch/utils/mfac_helpers.py
```diff
@@ -38,6 +50,75 @@
...
+
+    def __init__(
+        self,
+        dataloader: Iterator[Tuple[Union[Tensor, List[Tensor]], Any]],
+        loss_fn: Callable[[Tensor], Tensor],
+    ):
+        self._dataloader = dataloader
+        self._loss_fn = loss_fn
+
+    @staticmethod
```
markurtz, line 73: static method won't allow the implementer to override -- should probably change out for class or instance if we expect there's any chance they'll need to access state

src/sparseml/pytorch/utils/mfac_helpers.py
```diff
@@ -38,6 +50,75 @@
...
+    def module_forward(module: Module, data: Union[Tensor, List[Tensor]]) -> Any:
+        """
+        :param module: module to perform forward pass with
+        :param data: single data sample to pass to module
+        :return: output(s) of the module forward pass
+        """
+        if isinstance(data, Tensor):
```
markurtz, line 80: would it make sense to use the tensors_module_forward function here ideally making it more robust and generic?

src/sparseml/pytorch/utils/mfac_helpers.py
```diff
@@ -38,6 +50,75 @@
...
+        each step
+    """
+
+    def __init__(
+        self,
+        dataloader: Iterator[Tuple[Union[Tensor, List[Tensor]], Any]],
+        loss_fn: Callable[[Tensor], Tensor],
+    ):
+        self._dataloader = dataloader
+        self._loss_fn = loss_fn
```
markurtz, line 71: NIT: check if loss_fn is callable and dataloader is iterable

markurtz (changes requested):

markurtz (approved):
LGTM

rahul-tuli (approved):
LGTM!

PR #308, opened 2021-06-27
update:pytorch example notebooks to use manager.modify(...)

rahul-tuli:
Update pytorch example notebooks to use the manager.modify(...) instead of older ScheduledOptimizer(...)
fix epoch initialization to 0 instead of None


bfineran (commented):
Looks great @rahul-tuli! Thanks for adding in the change to vision.py as well. Could you update the notebook(s) documentation to point out that `finalize()` should be called after training?

rahul-tuli:
> Looks great @rahul-tuli! Thanks for adding in the change to vision.py as well. Could you update the notebook(s) documentation to point out that `finalize()` should be called after training?

Sure! Coming right up.

rahul-tuli:
@bfineran Does this look okay?

bfineran (commented):
looks good @rahul-tuli. let's also surround the finalize(...)s in backticks `` `finalize(...)` so they will be rendered as code

rahul-tuli:
@bfineran, added backticks

bfineran (approved):

markurtz (approved):

PR #310, opened 2021-06-27
update:LearningRateModifier(...) and LearningRateFunctionModifier(...)

rahul-tuli:
to add verbosity level


src/sparseml/pytorch/optim/modifier_lr.py
```diff
@@ -273,6 +297,7 @@ def __init__(
         param_groups: Optional[List[int]] = None,
         update_frequency: float = -1.0,
         log_types: Union[str, List[str]] = ALL_TOKEN,
+        constant_logging: int = 1,
```
markurtz, line 300: constant_logging name doesn't seem to fit with what this is doing. Can we update to something like log_level

src/sparseml/pytorch/utils/verbosity_helper.py
```diff
@@ -0,0 +1,35 @@
...
+from enum import Enum
+
+
+__all__ = [
+    "Verbosity",
+]
+
+
+class Verbosity(Enum):
+    DEFAULT = 1
```
markurtz, line 24: making these integers isn't going to be easy for anyone but advanced users to know which levels are which. Let's switch these over to string constants so the yaml mapping matches what the user is trying to input. Additionally should support non truthy such as False and None and have that evaluate to OFF

src/sparseml/pytorch/utils/verbosity_helper.py
```diff
@@ -0,0 +1,35 @@
...
+
+from enum import Enum
+
+
+__all__ = [
+    "Verbosity",
+]
+
+
+class Verbosity(Enum):
```
markurtz, line 23: missing class docs

src/sparseml/pytorch/utils/verbosity_helper.py
```diff
@@ -0,0 +1,35 @@
...
+    DEFAULT = 1
+    ON_LR_CHANGE = 2
+    ON_EPOCH_CHANGE = 3
+    ON_LR_OR_EPOCH_CHANGE = 4
+    OFF = 0
+
+    @staticmethod
+    def convert_int_to_verbosity(value):
```
markurtz, line 31: missing public function docs

tests/sparseml/pytorch/utils/test_verbosity_helper.py
```diff
@@ -0,0 +1,46 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
```
markurtz, line 1: We'll need to additionally add an integration test within the modifiers to make sure that the logging is doing what we intend

markurtz (changes requested):

markurtz:
This is a good direction. I would say let's hold off on this and I'm going to close it out for now. I think there's more we could do with this to build a concept of logging level into the base modifiers along with the loggers that are used with them. We could build this in as part of the phase 2 refactor. Ex we could have the loggers have a concept of level where level is broken down into "step", "epoch", "change", etc. Each log is then done with a level from the modifiers to pass the info for what the significance of the log is. Additionally, we could allow the modifiers to override the logging level for a per instance basis

markurtz:
(@rahul-tuli @bfineran )

PR #311, opened 2021-06-27
Phased pruning

markurtz:
Implementation for allowing for phased pruning where the imposed sparsity alternates between the normal function for sparsity and no sparsity. Pruning starts on and alternates with the update_frequency. All weight pruning modifiers are supported with this. Note, if the sparsity ends up not being imposed at the final_epoch, it will be at the very end. Example:
```yaml
!GMPruningModifier
    init_sparsity: 0.9
    final_sparsity: 0.9
    start_epoch: 0.0
    end_epoch: 10.0
    update_frequency: 2.0
    params: ["re:.*weight"]
    phased: True
```


bfineran (approved):
really slick implementation. LGTM. Will we want to enable variable lengths for pruning/non-pruning phases in the future?

rahul-tuli (approved):
Looks Good to me! :)

markurtz:
> really slick implementation. LGTM. Will we want to enable variable lengths for pruning/non-pruning phases in the future?

Good question, I could potentially see that happening. For now, though, I've only heard about equivalent phases from the research. Shouldn't be too much in the future to overload the phased param to support this if we need it

PR #314, opened 2021-07-01
Replace quant/dequant stubs around matmul with QLinearMatmul

natuan:
This PR needs to land with/after https://github.com/neuralmagic/sparseml/pull/283

It turns the following graph:

before

to:

after


bfineran (approved):
great job @natuan. LGTM.

Could you see if the produced graph passes the ONNX checker?

we should really get in some integration tests for this file some day (more so on me).

mgoin (approved):

PR #317, opened 2021-07-08
skip initial QuantizeLinear node optimization for YOLOv5

bfineran:
this optimization will allow for significant execution savings in the initial operations of YOLOv5 by allowing the model to take quantized (uint8) inputs.

Before:
Screen Shot 2021-07-08 at 6 59 50 PM

After:
Screen Shot 2021-07-08 at 6 59 39 PM


markurtz (approved):

mgoin:
@bfineran does the input to the model need to have the same scale and zero_point as the QuantizeLinear that is being removed?

bfineran:
> @bfineran does the input to the model need to have the same scale and zero_point as the QuantizeLinear that is being removed?

@mgoin for best results likely, yes. the reason we can do this optimization for models trained in the YOLO pipeline is that the raw images (stored as RGBs 0-255) are just transformed to float and fed to the convs as `img / 255`.  (the model used in the images I provided are from a toy model so the scale isn't finely tuned, but usually for these YOLO models, the initial scale is approx. `1/255`.

So this wouldn't work well for something like ImageNet training where normalization is usually used.

mgoin (approved):

PR #324, opened 2021-07-14
Allow for export ONNX model with dict inputs

natuan:
This allows for passing in inputs of dict type to the export_onnx, enabling exported model with the original inp/out names.

### DIFF
--- a/src/sparseml/pytorch/utils/exporter.py
+++ b/src/sparseml/pytorch/utils/exporter.py
@@ -15,11 +15,12 @@
 """
 Export PyTorch models to the local device
 """
-
+import collections
 import logging
 import os
+import warnings
 from copy import deepcopy
-from typing import Any, Callable, Iterable, List, Optional, Tuple
+from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple
 
 import numpy
 import onnx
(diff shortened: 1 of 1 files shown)


src/sparseml/pytorch/utils/exporter.py
```diff
@@ -151,6 +151,22 @@ def export_to_zoo(
             sample_originals=sample_originals,
         )
 
+    def get_output_names(self, out: Any):
```
bfineran, line 154: nit: method should be static

bfineran (commented):
LGTM overall, I'm worried that for models with multiple inputs, the proper input order will not be preserved by a `Dict` unless it is an `OrderedDict`. 

It might be safer to have `input_names` as a separate arg whose names should match the order of the sample input. Having a warning if the `Dict` is not ordered might also be ok though. If we keep `Dict` the transformers integration should be updated to use an ordered one.

src/sparseml/pytorch/utils/exporter.py
natuan: Fixed

natuan:
> LGTM overall, I'm worried that for models with multiple inputs, the proper input order will not be preserved by a `Dict` unless it is an `OrderedDict`.
> 
> It might be safer to have `input_names` as a separate arg whose names should match the order of the sample input. Having a warning if the `Dict` is not ordered might also be ok though. If we keep `Dict` the transformers integration should be updated to use an ordered one.

Added the followings:

- Warning if sample batch passed in is a dict but not an ordered dict
- From transformers side, the sample batch constructed as OrderedDict

bfineran (dismissed):
LGTM, thanks @natuan!

bfineran:
looks like quality needs to run @natuan

bfineran (approved):

markurtz (approved):

PR #352, opened 2021-08-03
Generic Benchmark API

kevinaer:
Generic benchmarking APIs for use with different frameworks.

Includes:
- BenchmarkResults class representing benchmark results, based on class in deepsparse
- Abstract BenchmarkRunner, which would be implemented using different SparseML frameworks
- Serialization via BenchmarkInfo
- Ability to run benchmarks via CLI
- Ability to run benchmarks via python API
- Ability to reload a BenchmarkInfo json and create a new run with the same conditions


src/sparseml/benchmark/info.py
```diff
@@ -0,0 +1,520 @@
...
+        progress_steps = self.warmup_iterations + self.iterations
+        _LOGGER.debug("running {} items through model".format(progress_steps))
+        data_iter = (
+            enumerate(data)
+            if not show_progress
+            else auto.tqdm(enumerate(data), desc=desc, total=progress_steps)
+        )
+        for index, batch in data_iter:
+            if index < self.warmup_iterations:
+                continue
```
markurtz, line 177: we should still be running the batch for warmup here, correct?

markurtz (commented):
I'm a bit confused on the separation of the pydantic schemas and the result objects. Is there any reason we can implement the result objects as pydantic models and override accordingly? Ideally would make the flow a lot easier so serialization is built into every return type we have (the original intention with everything being added in this new phase 2 step)

src/sparseml/benchmark/info.py
kevinaer: Ah sorry was supposed to run self.run_batch then continue, will fix

markurtz (approved):

rahul-tuli (approved):
GG!

PR #353, opened 2021-08-03
torch QATWrapper class, custom QConfig support

bfineran:
goal of this PR is to enable a workflow that allows users to create Modules that do not require sparseml and normally are not quantized in default torch/sparseml quantization flows, but with no further edits, could be converted to a fully quantizable QAT graph using a recipe.

`QATWrapper` achieves this by allowing users to wrap arbitrary operations with Modules that can later be wrapped as long as `wrap_qat` is set to True in the module.

This PR also introduces support that allows users to override the default QConfigs assigned to Modules if a module provides its own `configure_qconfig` method.

## QAT MatMul Example
The following can be used to Quantize the non-parameterized MatMul in the attention layers of transformer models. `sparseml` is not a dependency, and until it is processed by SparseML, the following module behaves just like `torch.matmul`
```python
import torch

class QATMatMul(torch.nn.Module):
    def __init__(self):
        super().__init__()

        self.wrap_qat = True
        self.qat_wrapper_kwargs = {
            "num_inputs": 2,
            "input_qconfigs": ["asymmetric", "symmetric"],
        }

    def forward(self, a: torch.Tensor, b: torch.Tensor):
        return torch.matmul(a, b)
```

#### After ONNX export:
Screen Shot 2021-08-03 at 5 00 49 PM

#### After QAT -> Quant conversion:
(notice the `b` tensor has a symmetric zero point)
Screen Shot 2021-08-03 at 5 00 57 PM

### DIFF
--- a/src/sparseml/pytorch/utils/quantization/helpers.py
+++ b/src/sparseml/pytorch/utils/quantization/helpers.py
@@ -77,6 +77,8 @@ class QATWrapper(Module):
     :param num_inputs: number of inputs of the forward function to add a QuantStub
         to. Will wrap the first num_inputs ordered inputs of the function. Default
         is 1
+    :param kwarg_input_names: list of names of key word arguments to the forward pass
+        that should be wrapped with a fake quantize operation. Defaults to empty
     :param num_outputs: number of outputs of the forward function to add a QuantStub
         to. Will wrap the first num_inputs ordered outputs of the function. Default
         is 1. Will also add a DeQuantStub for FP32 conversion if
@@ -101,17 +103,11 @@ class QATWrapper(Module):
             function. Will attempt to find any other named parameter of the QATWrapper
             constructor from the attributes of the given Module
         """
-        qat_wrapper_param_names = [
-            "num_inputs",
-            "num_outputs",
-            "input_qconfigs",
-            "output_qconfigs",
-        ]
-        qat_wrapper_kwargs = {
-            param_name: getattr(module, param_name)
-            for param_name in qat_wrapper_param_names
-            if hasattr(module, param_name)
-        }
+        qat_wrapper_kwargs = (
+            module.qat_wrapper_kwargs or {}
+            if hasattr(module, "qat_wrapper_kwargs")
+            else {}
+        )
 
         return QATWrapper(forward_fn=module, **qat_wrapper_kwargs)
 
@@ -119,6 +115,7 @@ class QATWrapper(Module):
         self,
         forward_fn: Callable[[Any], Any],
         num_inputs: int = 1,
+        kwarg_input_names: List[str] = None,
         num_outputs: int = 1,
         input_qconfigs: Union[
             "torch.quantization.QConfig", str, List["torch.quantization.QConfig"]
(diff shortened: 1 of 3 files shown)


src/sparseml/pytorch/utils/quantization/helpers.py
```diff
@@ -64,6 +67,214 @@
...
+            function. Will attempt to find any other named parameter of the QATWrapper
+            constructor from the attributes of the given Module
+        """
+        qat_wrapper_param_names = [
+            "num_inputs",
+            "num_outputs",
+            "input_qconfigs",
+            "output_qconfigs",
+        ]
+        qat_wrapper_kwargs = {
```
markurtz, line 110: could we nest these one level deeper into a specifically named dictionary and then pull these attributes to ensure there's little chance for namespace collision on the original module? Mainly worried about the num_inputs and num_outputs as those can be fairly generic names

src/sparseml/pytorch/utils/quantization/helpers.py
```diff
@@ -64,6 +67,214 @@
...
+        output_qconfigs: Union[
+            "torch.quantization.QConfig", str, List["torch.quantization.QConfig"]
+        ] = "asymmetric",
+    ):
+        super().__init__()
+
+        if torch_quantization is None:
+            raise RuntimeError(
+                "Unable to import package torch.quantization. "
+                "Try upgrading your PyTorch version."
```
markurtz, line 135: can we add in the min PyTorch version here of 1.7 to give a bit more info

src/sparseml/pytorch/utils/quantization/helpers.py
```diff
@@ -64,6 +67,214 @@
...
+        :param args: arguments to forward function; the first num_inputs of these args
+            will be wrapped by a QuantStub
+        :param kwargs: key word arguments to pass to the wrapped forward function
+        :return: outputs of the forward function with a QuantStub applied to the first
+            num_outputs outputs
+        """
+
+        qat_args = []
+
+        for idx, arg in enumerate(args):
```
markurtz, line 173: It's less of an edge case based on how most PyTorch code is implemented currently, but if someone does module(inp=tensor) we won't currently catch it. We should note this in the docs that only positional args are supported currently or try and update to be more generic so we can go through named kwargs by key and positional by index

src/sparseml/pytorch/utils/quantization/helpers.py
```diff
@@ -64,6 +67,214 @@
...
+                arg = self.input_quant_stubs[idx](arg)
+            qat_args.append(arg)
+
+        outputs = self.forward_fn(*qat_args, **kwargs)
+
+        if len(self.output_quant_stubs) == 0:
+            # no output wrapping
+            return outputs
+
+        if isinstance(outputs, torch.Tensor):
```
markurtz, line 184: could we also check for this case that the setup was told to only create one output_quant_stub and otherwise raise an exception?

markurtz (changes requested):

src/sparseml/pytorch/utils/quantization/helpers.py
bfineran: good point, updating

src/sparseml/pytorch/utils/quantization/helpers.py
bfineran: will add support for key word args

bfineran:
thanks for the review @markurtz added your suggestions

mgoin (approved):
nice lgtm

markurtz (approved):

PR #354, opened 2021-08-04
Onnx benchmark

kevinaer:
Changes:
- Added `check_onnxruntime_gpu_install` and `require_onnxruntime_gpu` annotation
- Added ORT Benchmark runners for CPU, CUDA, and generic ORT with user provided execution provider
- Updated DeepSparseModelRunner to take in all inputs that `compile_model` takes in
- Added `create_benchmark_runner` and `run_benchmark` to call into from `sparseml.benchmark` command


setup.py
```diff
@@ -79,6 +79,7 @@
     "sphinx-rtd-theme>=0.5.0",
     "wheel>=0.36.2",
     "pytest>=6.0.0",
+    "pytest-mock>=3.6.1",
```
markurtz, line 82: 🥑

src/sparseml/onnx/base.py
```diff
@@ -31,20 +31,24 @@
...
     onnxruntime_err = err
-
+    onnxruntime_gpu_installed = False
 
 __all__ = [
     "onnx",
     "onnx_err",
     "onnxruntime",
     "onnxruntime_err",
+    "onnxruntime_gpu_installed",
```
markurtz, line 45: I'm not sold on the additional checks for a specific provider type. ORT has a lot of providers that we may want to support in the future, can we make something more generic here for checking for a given provider or list of providers?

src/sparseml/onnx/benchmark/info.py
```diff
@@ -0,0 +1,599 @@
...
+
+    @property
+    def model(self) -> ModelProto:
+        """
+        :return: the model as an ONNX ModelProto
+        """
+        return self._model
+
+
+class ORTCudaBenchmarkRunner(ORTBenchmarkRunner):
```
markurtz, line 212: I'm worried about this scaling as well across all of the different providers we may want to support in the future for ORT. Is there anyway to make this more generic and set a provider or list of providers arg and then run checks based on that?

markurtz (changes requested):

markurtz (dismissed):
LGTM, left a small comment

markurtz (approved):

bfineran (approved):

PR #359, opened 2021-08-16
fix issues in configure_module_default_qconfigs

bfineran:
this PR fixes a logic error (reference to wrong module) and infinite looping edge case in `configure_module_default_qconfigs`


markurtz:
@bfineran is there anything we can add as a test case with this diff?

bfineran:
@markurtz yup, just added an extra layer to the test module which triggers the bugs that this fix addresses

markurtz (approved):

rahul-tuli (approved):
LGTM! Good work @bfineran

PR #361, opened 2021-08-17
Support for simple arithmetic in recipes

bfineran:
## Summary of features:
* support for simple arithmetic (`+, -, *, /`) operations in recipes over metadata variables, and variables defined as top level numeric values in a recipe (variables can also be string equations based on metadata variables)
* in order to be evaluated, a string must be wrapped by `eval(...)`. Metadata may not include eval values

## Example Recipe:
recipe:
```yaml
num_epochs: 10.0

pruning_start_epoch: eval(num_epochs * 0.2)
init_sparsity: 0.2

modifiers:
    - !EpochRangeModifier
        end_epoch: 1.0
        start_epoch: 0.0

    - !GMPruningModifier
        end_epoch: 8
        final_sparsity: eval(init_sparsity + 0.7)
        init_sparsity: eval(init_sparsity)
        inter_func: cubic
        leave_enabled: True
        log_types: __ALL__
        mask_type: [1, 4]
        params: __ALL_PRUNABLE__
        start_epoch: pruning_start_epoch
        update_frequency: 0.01
```

when loaded to a `Manager` and printed in python evaluates to:
```yaml
modifiers:
    - !EpochRangeModifier
        end_epoch: 1.0
        start_epoch: 0.0

    - !GMPruningModifier
        end_epoch: 8
        final_sparsity: 0.8999999999999999
        global_sparsity: False
        init_sparsity: 0.2
        inter_func: cubic
        leave_enabled: True
        log_types: __ALL__
        mask_type: [1, 4]
        params: __ALL_PRUNABLE__
        phased: False
        score_type: magnitude
        start_epoch: 2.0
        update_frequency: 0.01
```


markurtz:
Thanks, Ben! Obviously generally supporting an eval call for Python is too dangerous, but can we look around and see if we can find a general math parsing library that we can plugin? Ideally, we can pick up more functionality and also have less to maintain in the future.

bfineran:
updated - now we run through a private eval function that goes through `ast.parse`, additionally, `eval(...)` is required for all recipe eval expressions and variable eval functions may reference each other as long as they don't create a loop

markurtz:
@bfineran let's discuss in one on one, I'm concerned about the limitations on metadata currently so would like to brain storm what we can do

bfineran:
@markurtz updated based on our discussion. formal metadata fields removed

markurtz (approved):
Looks great Ben!

rahul-tuli (approved):

PR #362, opened 2021-08-20
Unwrap DP/DDP torch modules on manager.modify

bfineran:
proposed change to help simplify when managers need to be initialized in existing training scripts.  this can come up when dataloaders are created after other DDP artifacts are made meaning for a simple integration, managers would have to come after DDP initialization. this PR will require further testing before landing


markurtz:
I'm nervous about defaulting to this. Could we add in a warning for the user if a parallel model is detected? Additionally can add in an argument that will unwrap but default it to the current functionality then have that warning point to the arg that will allow unwrapping. If we notice no warnings across our tests over the next release, then we can change the default functionality

bfineran:
thanks for the review @markurtz, completely agree with your sentiments, I've updated the functionality to match them

markurtz (approved):

rahul-tuli (approved):

PR #365, opened 2021-08-23
Add:Support for ONNX versions 1.8.*, 1.9.* upto 1.10.1

rahul-tuli:
* Replaced usage of onnx.helpers.make_model(model.graph) to deepcopy(model) in [file](https://github.com/neuralmagic/sparseml/blob/main/src/sparseml/onnx/utils/helpers.py) because opset and ir_versions were lost in the process initially which broke Tests when using onnx versions>=1.8.0 
* Added explicit IR_VERSION=6 in usages of onnx.helpers.make_model(...) in [file](https://github.com/neuralmagic/sparseml/blob/main/tests/sparseml/onnx/optim/quantization/helpers.py)
* Updated setup.py to install latest onnx version(at the time of this PR) when installing using pip -e ...

Tested on `Python 3.6.9`, and Linux/Debian system,
Tests ran:`make test TARGETS=onnx`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ _deps = [
     "numpy>=1.0.0",
     "matplotlib>=3.0.0",
     "merge-args>=0.1.0",
-    "onnx>=1.5.0,<1.8.0",
+    "onnx>=1.5.0,<=1.10.1",
     "onnxruntime>=1.0.0",
     "pandas>=0.25.0",
     "packaging>=20.0",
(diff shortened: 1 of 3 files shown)


bfineran (approved):
@rahul-tuli LGTM, before landing, could you add a comment about which tests you ran and with what environments installed?

rahul-tuli:
> @rahul-tuli LGTM, before landing, could you add a comment about which tests you ran and with what environments installed?

kevinaer (approved):

PR #367, opened 2021-08-24
Mark test_one_shot_ks_perf_sensitivity as flaky

dbarbuzzi:
Marking the below test as flaky per a previous conversation:

`tests/sparseml/onnx/optim/test_sensitivity_ks.py::test_one_shot_ks_perf_sensitivity`


kevinaer (approved):

bfineran (approved):
do we know what part of this test is flaky?

markurtz:
This is for performance sensitivity, which we should not be testing exact results for because those will change across environments. This is fine to land for now, but the proper, long-term fix is to refactor the test such that it isn't testing for exact numbers.

markurtz (approved):

PR #381, opened 2021-09-08
PyTorch Distillation modifier

bfineran:
adds a `DistillationModifier` for PyTorch recipes. a `distillation_teacher` module must be provided during the `initialize` phase and the `loss_update` phase of the SparseML lifecycle must also be triggered

### DIFF
--- a/src/sparseml/pytorch/optim/modifier_distillation.py
+++ b/src/sparseml/pytorch/optim/modifier_distillation.py
@@ -190,6 +190,28 @@ class DistillationModifier(ScheduledModifier):
         super().update(module, optimizer, epoch, steps_per_epoch)
         self._check_distillation_update(module, epoch, steps_per_epoch)
 
+    def update_ready(self, epoch: float, steps_per_epoch: int) -> bool:
+        """
+
+        :param epoch: current epoch and progress within the current epoch
+        :param steps_per_epoch: number of steps taken within each epoch
+            (calculate batch number using this and epoch)
+        :return: True if the modifier is pending an update and update() should be called
+        """
+        if not self._initialized:
+            raise RuntimeError("modifier must be initialized first")
+
+        if not self._enabled:
+            return False
+
+        pending = (
+            self.start_pending(epoch, steps_per_epoch)
+            or self.end_pending(epoch, steps_per_epoch)
+            or (not self._distillation_enabled and self._is_distillation_epoch(epoch))
+        )
+
+        return pending
+
     def loss_update(
         self,
         loss: Tensor,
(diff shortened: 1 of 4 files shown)


src/sparseml/pytorch/optim/modifier_distillation.py
```diff
@@ -0,0 +1,326 @@
...
+class DistillationModifier(ScheduledModifier):
+    """
+    Adds a knowledge distillation loss based on a teacher model during the
+    loss_update phase of the SparseML lifecycle. A distillation_teacher
+    module must be provided as a kwarg to the Manager initialization and
+    loss_update(loss) must be called before any backwards pass in the integrated
+    training flow
+
+    | Sample yaml:
+    |   !DistillationModifier
```
spacemanidol, line 47: @bfineran just for the logability and tracability of our recipes it would be useful to require the model path for the teacher.

spacemanidol (commented):
looks good but as mentioned having some traceability of what the teacher model was would be useful

src/sparseml/pytorch/optim/modifier_distillation.py
markurtz: Model path is local state that would break the recipes for anything outside of the environment it was originally used on. State for how to define the teacher should be managed externally to this with the caller code and passed in so we can keep this extendable and usable across systems

markurtz (changes requested):
Adding in the convo from slack:

Benjamin Fineran  32 minutes ago
@Mark Kurtz this will break currently flows with manager.apply since the initialize won't have the teacher. thoughts on having the one-shot apply function skip over DistillationModifiers?

Mark Kurtz  7 minutes ago
Yep, so there's more functionality I want to get in with this:
- default to self distillation and create a model from the given one at the start_epoch
- allow override of self distillation with another model passed in through the kwargs
- allow the override to specify not to use distillation at all (in case a user wants to disable distillation for a run and we won't require them to edit the recipe)
-- apply could use this functionality
- options for what to compare for the outputs in the event that a model has multiple outputs such as our classification models which have a tuple of the logits and softmax results

src/sparseml/pytorch/optim/modifier_distillation.py
markurtz: To clarify a bit more, the integration within hugging face would be responsible for receiving the path to the teacher from the user and passing that into the modifiers

bfineran: resolving this - we should make a point to log the teacher model in integration scripts when possible to your point @spacemanidol

bfineran:
thanks @markurtz @spacemanidol. @markurtz these changes should now be reflected on this branch

markurtz (dismissed):

markurtz:
@bfineran can we get logs in there to log under a DistillationModifier group for the following:

- original loss
- teacher loss
- distillation loss

rahul-tuli (dismissed):
LGTM!

markurtz (approved):

rahul-tuli (approved):

PR #384, opened 2021-09-10
Force conversion of LR variables to float due to type coercion from manager serialization

markurtz:


bfineran (approved):
LGTM. We should consider something more generic later

rahul-tuli (approved):
Lgtm!

PR #393, opened 2021-09-20
No need to toss conversion because of branching inputs.

vixadd:
I've had a consistent problem with multiple ONNX models where (despite different names for layers) I've had to skirt around this assertion error using this workaround. I think this is valueable for the project. However, please correct me if I'm wrong. If there's something I'm missing as to why this assertion error needs to be here, please explain it to me. I'd like to understand more about re-quantization of models. However, my understanding is that if the input bias hasn't changed from a previous calculation, you can just re-use it.


bfineran (approved):
Hi @vixadd thanks for this update, LGTM!

markurtz (approved):

PR #398, opened 2021-10-04
Revised distillation modifier

natuan:
I made the following changes to the distillation modifier:
- Removed the use of forward hook to capture the latest inputs/outputs because it ignores the kwargs dict passed in to model forward (which is the type of model inputs, e.g. in QA model)
- Removed the computation of teacher outputs inside the modifier's loss_update. Passing in the entire inputs dict into the teacher forward causes cross entropy loss to be calculated, which we don't need; it also creates an issue downstream from autograd when doing loss.backward.

### DIFF
--- a/src/sparseml/pytorch/optim/modifier_distillation.py
+++ b/src/sparseml/pytorch/optim/modifier_distillation.py
@@ -21,6 +21,7 @@ import logging
 from copy import deepcopy
 from typing import Any, Dict, Iterable, List, Optional, Union
 
+import torch
 import torch.nn.functional as TF
 from torch import Tensor
 from torch.nn import Module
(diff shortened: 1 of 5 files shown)


src/sparseml/pytorch/optim/modifier_distillation.py
```diff
@@ -233,69 +236,45 @@ def loss_update(
...
-                for tens in self._student_inputs.values()
-                if isinstance(tens, Tensor)
-            ][0]
-        )
-        self._teacher.to(target_device)
-
-        teacher_outputs = tensors_module_forward(
-            self._student_inputs, self._teacher, check_feat_lab_inp=False
-        )
+        if student_outputs is None or teacher_outputs is None:
```
markurtz, line 246: We shouldn't silently ignore here, if the user passed in something wrong it's impossible to tell that it isn't working correctly

markurtz (changes requested):
@natuan our goal is to remove as much complexity for the user with modifiers as possible. This implementation is adding more complexity back to the user, specifically needing to manage the complete lifecycle of the teacher model themselves now. Additionally, the old code, as set up, would work for other pipelines and reduce simplicity for users.

markurtz (approved):

bfineran (approved):
LGTM

PR #400, opened 2021-10-07
sparsification oracle base classes

bfineran:
This PR introduces classes that form a base to perform auto sparsification recipe generation with.


markurtz (dismissed):
@bfineran I forgot to mention that we are moving to pydantic for all data classes in the sparsification packages. This is to enable consistent serialization and potential storage in a DB for future work in Sparsify. Given release, let's include this after. Additionally, we'll want to add in a run_iter function for the analysis classes to enable progress bars in the future which we can add in a later state

rahul-tuli (approved):

markurtz (approved):

PR #402, opened 2021-10-07
Update: Save Strategy

rahul-tuli:
- Update the train script to use a save strategy using command line argument `--save-strategy`.
  The save strategy can be one of the following:
        - `no`: No save is done during training.
        - `epoch`: Save is done at the end of each epoch.
        - `steps`: Save is done every :obj:`save_steps`
 - Add support to delete older checkpoints using `--save-total-limit` argument
 - Specify save steps using `--save-steps`

### DIFF
--- a/integrations/pytorch/recipes/resnet50-imagenette-pruned.md
+++ b/integrations/pytorch/recipes/resnet50-imagenette-pruned.md
@@ -133,7 +133,7 @@ Alternatively, a full walk-through notebook is located at `sparseml/integrations
 *script command:*
 
 ```
-python integrations/pytorch/scripts/train.py \
+python integrations/pytorch/train.py \
     --recipe-path zoo:cv/classification/resnet_v1-50/pytorch/sparseml/imagenette/pruned-conservative?recipe_type=original \
     --pretrained True \
     --arch-key resnet50 \
--- a/integrations/pytorch/recipes/resnet50-imagenette-pruned.md
+++ b/integrations/pytorch/recipes/resnet50-imagenette-pruned.md
@@ -127,7 +127,7 @@ using a SGD optimizer.
 When running, adjust hyperparameters based on training environment and dataset.
 
 ## Training
-The training script can be found at `sparseml/integrations/pytorch/scripts/train.py`. 
+The training script can be found at `sparseml/integrations/pytorch/train.py`. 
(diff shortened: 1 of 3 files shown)


lucaslie (changes requested):
I get this error when trying to run it with the flag:

```bash
(base) root@bc608c919dc1:/src# python integrations/pytorch/train.py     --recipe-path zoo:cv/classification/resnet_v1-50/pytorch/sparseml/imagenette/pruned-conservative?recipe_type=original     --pretrained True     --arch-key resnet50     --dataset imagenette     --dataset-path /data/imagenette      --train-batch-size 128 --test-batch-size 256     --loader-num-workers 8 --save_last_epoch True     --model-tag resnet50-imagenette-pruned-conservative                       
Traceback (most recent call last):
  File "integrations/pytorch/train.py", line 696, in <module>
    main()
  File "integrations/pytorch/train.py", line 670, in main
    training_args, _ = _parser.parse_args_into_dataclasses()
  File "/src/integrations/pytorch/argparser_/nm_argparser_.py", line 250, in parse_args_into_dataclasses
    raise ValueError(
ValueError: Some specified arguments are not used by the NmArgumentParser: ['--save_last_epoch', 'True']
```

So I couldn't try it out yet.

rahul-tuli:
[this](https://github.com/neuralmagic/sparseml/pull/402#pullrequestreview-774384598) resolved offline, need to use hyphens instead of underscores in `--save-last-epoch`

markurtz (changes requested):
@rahul-tuli it might be beneficial to look into the transformers and ultralytics code to come up with better save strategies. I'm worried about overloading with parameters in the future as we add more functionality here. Additionally, let's be sure to keep similar args grouped together in the code base. This makes discoverability and readability quite a bit easier

lucaslie (changes requested):
Looks like a very nice way to handle checkpoint saving. I just have two comments:

1. Naming convention for `save_steps`, see comment. 

2. With regards to my comment [here](https://github.com/neuralmagic/sparseml/pull/402#issuecomment-938802960): I now saw in the code (see [here](https://github.com/neuralmagic/sparseml/blob/5d1fc4bf83e0c3076bfd36a54dc020d09934f0f4/integrations/pytorch/utils.py#L356)) that we currently do not support checkpoint loading for continuing to train. Maybe we can raise a warning to make people aware that we currently lack support for that.

PR #413, opened 2021-10-15
Add framework fallback ability to execute in sparseml

markurtz:


bfineran (changes requested):
good catch on this necessary functionality. I'm a little worried about how coupled the `attempt_num` logic is in `detect_framework` with the execute function.  Is there a specific reason for doing it this way? Any thoughts on having a detect function that returns all possible frameworks for an item, and then having execute return the first valid in-framework run of the function between the possible frameworks?

bfineran (dismissed):
LGTM. If we see performance issues when trying to load large models in the wrong framework, let's definitely update the `detect_frameworks` function to use what you had before, or maybe a generator if that would work

bfineran (approved):

xkazuka (approved):

PR #417, opened 2021-10-18
Update BERT tutorial and recipes with distillation modifiers

natuan:
This change updates the recipes and tutorials in hugging face integration to use the new distillation modifiers.


bfineran (dismissed):
LGTM, let's eventually migrate these to the new recipe format and add a recipe kwargs option to the integration scripts so that the distillation hyperparams can be controlled via command line, but tracked via recipe

spacemanidol (dismissed):
looks good

bfineran (approved):

markurtz (approved):

PR #438, opened 2021-10-29
update deepsparse perf analysis onnx test files for latest DS build

bfineran:


dhuangnm:
Can we get this PR approved so it can land and fix the test failures?

markurtz (approved):

markurtz:
I'm going to go ahead and force-land this so we can clear the test errors on main @dhuangnm @bfineran

PR #439, opened 2021-10-29
Update: Documentation

rahul-tuli:
- Documentation updated to reflect
`ScheduledModifierManager.from_yaml(...)` can also take in
a raw `yaml` `str` as input


src/sparseml/optim/helpers.py
```diff
@@ -57,7 +57,7 @@ def load_recipe_yaml_str(
         '?recipe_type=<type>' parameter or include a `/<type>` subpath. Can also
         be a SparseZoo Recipe object. i.e. '/path/to/local/recipe.yaml',
         'zoo:model/stub/path', 'zoo:model/stub/path?recipe_type=transfer_learn',
-        'zoo:model/stub/path/transfer_learn'
+        'zoo:model/stub/path/transfer_learn'. Can also be a yaml str
```
markurtz, line 60: @rahul-tuli let's keep the two updates consistent for the docs. Additionally, let's format it a bit better to something like "Additionally, a raw yaml str is supported in place of a file path."

markurtz (changes requested):

src/sparseml/optim/helpers.py
rahul-tuli:
```suggestion
        'zoo:model/stub/path/transfer_learn'. Additionally, a raw yaml str is supported in place of a file path.
```

markurtz (approved):

PR #445, opened 2021-11-03
Add: Sparsification + Quantization Recipes for YOLACT

rahul-tuli:
- yolact.pruned.md
- yolact.pruned.perf.md
- yolact.quantized.md


bfineran (changes requested):
* the YAML recipe section should be surrounded by dashes `---` instead of underscores `___`
* where is the `module.` coming from in the recipes? is this part of the core yolact model or the DDP wrapper?

rahul-tuli:
> * the YAML recipe section should be surrounded by dashes `---` instead of underscores `___`
> * where is the `module.` coming from in the recipes? is this part of the core yolact model or the DDP wrapper?

Good catch @bfineran; Corrected!

markurtz:
@bfineran can you work with @rahul-tuli to paramaterize these recipes where if a user enters a different number of epochs it will shift the pruning and quantization start and ends appropriately?

rahul-tuli:
> @bfineran can you work with @rahul-tuli to paramaterize these recipes where if a user enters a different number of epochs it will shift the pruning and quantization start and ends appropriately?

Done!

bfineran (dismissed):
Great job on the updates, they look structurally correct to me. Let's make sure they can all be loaded into a `ScheduledModifierManager`. For the recipe composition, the pruned perf recipe is actually a block pruning recipe, so let's merge that with the quantization recipe to make a single pruned quant recipe. Looks good to land after that.

also noting that `--resume` and `--recipe` arguments should be updated with proper sparsezoo stubs

markurtz (dismissed):

PR #448, opened 2021-11-12
FourBlockMaskCreator - block masks with no divisibility requirement

bfineran:
The normal `[1, 4]` `BlockPruningMaskCreator` requires that the input dimension of a parameter be divisible by 4. This causes failures after structured model thinning when input dimension shapes may change and no longer be divisible by four. This PR adds a `FourBlockMaskCreator` that does not have this restriction. It can be used in a pruning recipe with `mask_type: block`.


markurtz (approved):
LGTM

xkazuka (approved):

PR #450, opened 2021-11-15
Fix Quantization and ONNX export for stand alone top/classifier layer

natuan:
This PR fixes the quantization and ONNX export for models where the classifier layer does not have submodules. In that case, there're 3 issues: (1) the add_quant_dequant func when applied to this layer (assuming it's specified in the quantization recipe) leaves it unchanged, instead of being wrapped by quant and dequant constructs; (2) the onnx export disconnects the model output from the output of the classifier's GEMM; (3) the name of the model output is set to dummy strings "output_xyz".

An example of the correct output of the classifier:
after_fix


bfineran (dismissed):
LGTM, thanks for these fixes @natuan. small nit, can we change the name of `Mod2Quant` to something without the `2`?

src/sparseml/pytorch/optim/modifier_quantization.py
```diff
@@ -49,6 +50,8 @@
     "QuantizationModifier",
 ]
 
+Mod2Quant = collections.namedtuple("Mod2Quant", ["name", "module"])
```
markurtz, line 53: @natuan for consistency, please change over to NamedTuple from typings. Additionally, we need to add types for the attributes in this collection. Finally, if this is meant to be a private, internal variable please mark it as such by prefacing with an underscore: https://github.com/neuralmagic/sparseml/blob/a574de1dbac844080e986783ef7c88858c9a3a29/src/sparseml/onnx/utils/helpers.py#L226

Also, I agree with Ben that we want to stay away from putting numbers in our variable names. We've kept names verbose and clean for our current implementations and don't want to move away from that

src/sparseml/pytorch/optim/modifier_quantization.py
```diff
@@ -364,15 +367,17 @@ def _enable_module_qat(self, module: Module):
 
         # prepare each module / submodule for quantization
         qconfig = get_qat_qconfig()
-        for quant_module in self._modules_to_quantize:
+        for mod2quant in self._modules_to_quantize:
```
markurtz, line 370: please change out this variable name as well

markurtz (changes requested):

natuan:
> LGTM, thanks for these fixes @natuan. small nit, can we change the name of `Mod2Quant` to something without the `2`?

Changed it to ModuleToQuantize

src/sparseml/pytorch/optim/modifier_quantization.py
natuan: Changed to _ModuleToQuantize

src/sparseml/pytorch/optim/modifier_quantization.py
natuan: Fixed

markurtz (approved):

bfineran (approved):

PR #462, opened 2021-11-18
Throw if variable updating in recipe might not take effect

natuan:
I tried to update variables that are not defined inside the recipe, and the warning happened too fast that I didn't notice to fix. As a consequence, the model was being trained w/o me knowing I'm using the right value I would like. I think it would be better if we throw for the user to fix the recipe.


bfineran (approved):
good call @natuan

markurtz (approved):

PR #481, opened 2021-11-23
save torch recipes with state dict and apply on load

bfineran:
this PR adds support to save a torch sparsification recipe if a wrapped object is given to `save_model` and apply it on `load_model` if it is present in the state dict (must be under key `recipe`). This will support flows such as loading a pre-trained thinned or quantized torch model

### DIFF
--- a/src/sparseml/pytorch/utils/model.py
+++ b/src/sparseml/pytorch/utils/model.py
@@ -85,6 +85,9 @@ def load_model(
     model_dict = torch.load(path, map_location="cpu")
     current_dict = model.state_dict()
 
+    checkpoint_epoch = model_dict.get("epoch", float("inf"))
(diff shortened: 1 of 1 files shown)


src/sparseml/pytorch/utils/model.py
```diff
@@ -85,6 +85,9 @@ def load_model(
     model_dict = torch.load(path, map_location="cpu")
     current_dict = model.state_dict()
 
+    checkpoint_epoch = model_dict.get("epoch", float("inf"))
+    checkpoint_recipe = model_dict.get("recipe", None)
```
rahul-tuli, line 89: nit: can remove `None` in `model_dict.get(...)`

rahul-tuli (approved):
GG!

bfineran:
closing as functionality added by @dbogunowicz

PR #488, opened 2021-12-03
[MFAC] training aware implementation for GradSampler

bfineran:
* updates the `GradSampler` class to allow for more flexible inputs to models with `args` and `kwargs`
* allows for `GradSampler`s to be passed on recipe initialization so that they may be run before pruning steps to avoid the need for gradient buffering during training. if no grad samplers are passed, normal gradient buffering will occur
* will need to do some small live tests, and then get an integration with our BERT flows

### DIFF
--- a/research/mfac/tutorials/one_shot_pruning_with_mfac.md
+++ b/research/mfac/tutorials/one_shot_pruning_with_mfac.md
@@ -106,6 +106,19 @@ to extract gradients from the model to prune with. The code block below
 creates a `GradSampler` that has a data loader that loads the MNIST dataset to the correct
 device at batch size 16 and has a cross entropy loss function.
 
+The data loader must yield tuples of 
+`(forward_args: List, forward_kwargs: Dict, loss_targets: Any)`
+so that the forward and backwards passes may be run with the loss function as:
+```python
+forward_args, forward_kwargs, loss_targets = next(data_loader)
+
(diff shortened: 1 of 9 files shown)


research/mfac/tutorials/one_shot_pruning_with_mfac.md
```diff
@@ -118,11 +131,12 @@ data_loader = DataLoader(dataset, shuffle=True, batch_size=16)
...
     for sample in data_loader:
-        yield [t.to(device) if isinstance(t, torch.Tensor) else t for t in sample]
+        img, target = [t.to(device) if isinstance(t, torch.Tensor) else t for t in sample]
+        yield [img], {}, target
 
 # wrap cross entropy loss function to use correct output
-def loss_function(x, y):
-    return torch.nn.functional.cross_entropy(x[0], y)
+def loss_function(model_outputs, loss_target):
+    return torch.nn.functional.cross_entropy(model_outputs[0], loss_target)
```
spacemanidol, line 139: shouldn't we have an assert here that model_outputs is a tuple? Also worth adding a check that the shapes of loss target and model outputs match right?

spacemanidol (approved):
Looks good. Minor comment about adding an assert on outputs vs target outputs but feel free to ignore

rahul-tuli (approved):

PR #491, opened 2021-12-17
add transformers trainer integration and scripts to package

bfineran:
* sparseml-transformers integration can now be installed with `pip install sparseml[transformers]` this will pip install from neural magic's transformers fork
* related code stored in `sparseml.transformers`
* `SparseMLTrainer` class refactored to be pluggable with other subclasses of `Trainer`
* migrated training scripts for question answering, text classification, token classification, and masked language modeling (mlm not tested yet due to time)
* scripts added as command line executable through `$ sparseml.transformers.train.<task-name>`
* ONNX export flow removed from training script, to be later added as a standalone flow
* script arguments outside of ONNX export ones largely unchanged for ease of migration, subject to change for ease of use later

original integration authored by @natuan and @spacemanidol.


src/sparseml/transformers/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+    if transformers_import_error is None:
+        return
+    raise ImportError(
+        "No installation of transformers found. It is recommended to use the "
+        "sparseml fork of transformers which can be installed under "
+        "sparseml[transformers] or git+https://github.com/neuralmagic/transformers.git"
+    )
+
+
+_check_transformers_install()
```
markurtz, line 39: This isn't a release blocker, but could we migrate this to the new standards where the package is importable without transformers being installed but will fail when code is executed that requires the dependencies? Makes docs compilation a lot easier and also allows for more expansion in the future for capabilities in the case of optional dependencies for different code pathways

markurtz (approved):
Looks good overall. I would like to see if there's anything we could do in the future to move the sparseml.transformers package to fail after code execution for requirements checks rather than fail on import

KSGulin (approved):

PR #495, opened 2021-12-27
reset optimizer momentum during phased pruning steps

bfineran:


markurtz (changes requested):
I'm not sure if we should force this as this is likely a hyperparam we'd like to control for and have it default to this functionality. Let's revisit next sprint when we go through the image classification and pruning algorithms overhauls

PR #498, opened 2022-01-10
Implementation of BinaryCrossEntropy/CrossEntropy/InceptionLoss using label smoothing.

dbogunowicz:
CrossEntropy losses (Categorical) now can take `label smoothing` are as parameter.


dbogunowicz:
Just tested the current commit using command:

`python integrations/pytorch/train.py \
    --recipe-path ~/pruning_resnet50.yaml \
    --arch-key resnet50 --dataset imagenette \
    --train-batch-size 128 --test-batch-size 512`
  
  (pruning resnet50 on imagenette)
  
  ...and just for fun checked whether the loss graph looks aight:

image

dbogunowicz:
Will not be needed if we move repo to be compatible with pytorch 1.10.x (where label smoothing is explicitly implemented).

PR #499, opened 2022-01-10
Cyclic linear drop LR schedule

natuan:
This LR schedule drops the LR from an initial to a final value every pre-specified number of epochs. It can be defined in a recipe as following:

```
!LearningRateFunctionModifier
      start_epoch: 2
      end_epoch: 26
      lr_func: cyclic_linear
      init_lr: 8e-5
      final_lr: 8e-6
      cycle_epochs: 4 
```
The graph of this LR looks like:
cyclic_linear_drop

### DIFF
--- a/src/sparseml/pytorch/optim/modifier_lr.py
+++ b/src/sparseml/pytorch/optim/modifier_lr.py
@@ -257,6 +257,8 @@ class LearningRateFunctionModifier(ScheduledUpdateModifier):
         (set to -1.0 so it starts immediately)
     :param end_epoch: The epoch to end the modifier at,
         (set to -1.0 so it doesn't end)
+    :param cycle_epochs: The number of epochs between two consecutive LR rewinding;
+        used for cyclic_linear schedule only.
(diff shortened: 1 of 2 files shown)


spacemanidol (dismissed):

src/sparseml/pytorch/optim/modifier_lr.py
```diff
@@ -273,6 +273,7 @@ def __init__(
         final_lr: float,
         start_epoch: float,
         end_epoch: float,
+        cycle_epochs: float = -1.0,
```
mgoin, line 276: this default value of -1 (also for update_frequency) seems to be malformed based on how the value is used. should it even have a default value?

natuan: Set default value for cycle_epochs to 1; also verify it's positive if the schedule is cyclic_linear. For update_frequency, it's a dummy value for an unused argument.

mgoin (approved):
lgtm!

markurtz (approved):

PR #505, opened 2022-01-14
[transformers] trainers to prioritize recipe num_epochs + 0.10 fixes

bfineran:
main change:
* breaks out `recipes` in SparseMLTrainer into a sparsification recipe and list of checkpoint recipes to be applied before loading weights
* moves override of num_train_epochs to earlier in the training flow (previous location was after number of steps were calculated)

fixes:
* kwarg error on CSV data load for QA training
* support for underscores in task names in ONNX export script


KSGulin (dismissed):
Looks good to me and runs fine from the container side!

markurtz (approved):

KSGulin (approved):

rahul-tuli (approved):

PR #510, opened 2022-01-20
refactoring code with Dictionary Comprehension

zjzh:
refactoring code with Dictionary Comprehension which is more pythonic, concise and efficient; how do you think this change which has practical value？


markurtz (approved):
Thanks for the contribution @zjzh, looks great!

bfineran (approved):

markurtz:
Hi @zjzh, we had an issue where our quality checks we're not running properly. That has since been fixed and it looks like this PR is failing those now. Could you run `make quality` locally and then push up?

Thanks,
Mark

bfineran (approved):

bfineran:
thanks for the contribution @zjzh! pushed a small update to conform with our style checks and applied the same change to the `get_linear_layers` helper.

markurtz (approved):

PR #511, opened 2022-01-21
Install Sparseml-Transformers deps dyanmically on import

bfineran:
removes the `sparseml[transformers]` install pathway and instead installs the dependencies on any import from `sparseml.transformers`. Auto install can be prevented by setting the environment variable `SPARSEML_NO_AUTOINSTALL_TRANSFORMERS`.

Example:
From a new virtual environment:
```bash
cd sparseml
pip install -e ./[torch]

# will trigger sparseml-transformers install
sparseml.transformers.export_onnx -h
```

### DIFF
--- a/src/sparseml/transformers/__init__.py
+++ b/src/sparseml/transformers/__init__.py
@@ -47,24 +47,25 @@ def _install_transformers_and_deps():
         f"@{transformers_branch}"
     )
 
-    _pip.main(
-        [
-            "install",
-            transformers_requirement,
-            "datasets",
-            "sklearn",
-            "seqeval",
-        ]
-    )
-
     try:
+        _pip.main(
+            [
+                "install",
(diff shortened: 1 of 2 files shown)


src/sparseml/transformers/__init__.py
```diff
@@ -18,23 +18,86 @@
...
+        [
+            "install",
+            transformers_requirement,
+            "datasets",
+            "sklearn",
+            "seqeval",
+        ]
     )
 
+    try:
```
markurtz, line 60: I think we might want to extend this try catch to encompass the earlier portion as well in the event that pip doesn't have correct user access or something along those lines. This way we can give a good error to the user for the direction to fix

rahul-tuli (approved):

markurtz (approved):

PR #519, opened 2022-01-25
update transformers install to point to tarball + small fixes

bfineran:
* transformers install now points to a tarball from releases of neuralmagic/transformers
* update logic flow so previous installs of transformers are overwritten if they are not from the NM fork
* manager num_steps calculation bug fix from @eldarkurtic 
* convert_qat flag improvement
* pin datasets version to `1<.18.0` due to a new issue with the squad dataset
* use a subprocess call to perform pip install instead of pip API due to an issue found by @dbarbuzzi

### DIFF
--- a/src/sparseml/transformers/__init__.py
+++ b/src/sparseml/transformers/__init__.py
@@ -57,7 +57,7 @@ def _install_transformers_and_deps():
             [
                 "install",
                 transformers_requirement,
-                "datasets",
+                "datasets<1.18.0",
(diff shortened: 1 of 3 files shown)


KSGulin (dismissed):
Works fine in AutoSparse. LGTM!

markurtz (dismissed):

rahul-tuli (approved):

KSGulin (approved):

PR #526, opened 2022-01-25
Add: generic calibration dataloader

rahul-tuli:
This pull request adds a generic `calibration_dataloader` to `QuantizationModifier`

Steps to Test:
```python
import torch


def random_dataloader(n=5):
    """
    A random generator
    """
    for _ in range(n):
        print(f"Trying to yield batch {_}")
        yield [torch.rand(32, 3, 224, 224)], {}


# Pretrained Model

# Modifier init
modifier = QuantizationModifier(start_epoch=0)

# Modifier Application
modifier.apply(
    module=model,
    calibration_dataloader=random_dataloader(),
)

print(model)

``` 

Observer weights post modifier application and after first forward pass should look like 
```
(activation_post_process): MovingAverageMinMaxObserver(min_val=-4.431874752044678, max_val=6.029556751251221)
```

### DIFF
--- a/src/sparseml/pytorch/models/classification/resnet.py
+++ b/src/sparseml/pytorch/models/classification/resnet.py
@@ -42,6 +42,12 @@ from sparseml.pytorch.models.registry import ModelRegistry
 from sparseml.pytorch.nn import ReLU
 
 
+try:
+    from torch.nn.quantized import FloatFunctional
+except Exception:
+    FloatFunctional = None
+
+
 __all__ = [
     "ResNetSectionSettings",
     "ResNet",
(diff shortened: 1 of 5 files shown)


markurtz (changes requested):
We should build out a more generic approach here. The user currently doesn't have any control over what get's passed to the model other than hacking up the dataloader which may be outside the scope of what they work on. Dan working with YOLO, for example, would still take an ML engineer to reformat the loaders and model processing. Take a look through the ModuleTrainer and Tester classes and let's include some of that generic logic here to enable a user to override the forward and other pathways. CC @bfineran

PR #541, opened 2022-02-03
Add: calibration dataloader + forward func

rahul-tuli:
Adds: A generic calibratiion dataloader and a calibration forward function, for calibrating weights after training, (Post Training Quantization)

```python
import torch


def random_dataloader(n=5):
    """
    A random generator
    """
    for _ in range(n):
        print(f"Trying to yield batch {_}")
        yield [torch.rand(32, 3, 224, 224)], {}


# Pretrained Model

# Modifier init
modifier = QuantizationModifier(start_epoch=0)

# Modifier Application
modifier.apply(
    module=model,
    calibration_dataloader=random_dataloader(),
)

print(model)

```

Can also pass in a Calibration function along with dataloader

```python
import torch
from sparseml.pytorch.utils import  tensors_module_forward

def random_dataloader(n=5):
    """
    A random generator
    """
    for _ in range(n):
        print(f"Trying to yield batch {_}")
        yield [torch.rand(32, 3, 224, 224)], {}


# Pretrained Model

# Modifier init
modifier = QuantizationModifier(start_epoch=0)

# Modifier Application
modifier.apply(
    module=model,
    calibration_dataloader=random_dataloader(),
    calibration_func=tensors_module_forward,
)

print(model)

```

### DIFF
--- a/src/sparseml/pytorch/optim/modifier_quantization.py
+++ b/src/sparseml/pytorch/optim/modifier_quantization.py
@@ -18,7 +18,9 @@ Modifier for models through quantization aware training.
 PyTorch version must support quantization (>=1.2, ONNX export support introduced in 1.7)
 """
 import logging
-from functools import partial
+import warnings
+from contextlib import suppress
+from itertools import cycle
 from typing import (
     Any,
     Callable,
@@ -113,6 +115,8 @@ class QuantizationModifier(ScheduledModifier):
         transformer based models such as BERT where the quantized MatMul outputs
         are kept at 32 bits of precision and fake quantizing the outputs harm training
         recovery. Default is True
+    :param num_steps: Number of steps to run post training calibration for.
+            When None, the entire calibration_dataloader is used
     """
 
     def __init__(
@@ -127,6 +131,7 @@ class QuantizationModifier(ScheduledModifier):
         quantize_embeddings: bool = True,
         reduce_range: bool = False,
         quantize_linear_activations: bool = True,
+        num_calibration_steps: Optional[int] = None,
     ):
         if torch_quantization is None or torch_intrinsic is None:
             raise RuntimeError(
@@ -158,6 +163,7 @@ class QuantizationModifier(ScheduledModifier):
         self._bn_stats_frozen = False
         self._calibration_dataloader = None
         self._calibration_function = None
+        self._num_calibration_steps = num_calibration_steps
 
         if (
             isinstance(self._model_fuse_fn_name, str)
@@ -293,6 +299,14 @@ class QuantizationModifier(ScheduledModifier):
         """
         return self._quantize_linear_activations
 
+    @ModifierProp()
+    def num_calibration_steps(self) -> Optional[int]:
+        """
+        :return: Number of steps to run post training calibration for.
+            When None, the entire calibration_dataloader is used
+        """
+        return self._num_calibration_steps
+
     def initialize(
         self,
         module: Module,
@@ -312,8 +326,10 @@ class QuantizationModifier(ScheduledModifier):
         :param calibration_dataloader: optional dataloader for running post training
             quantization with the given model. if present, calibration will be run
             immediately after quantization is enabled
-        :param calibration_function: optional function to use for calibration of
-            model parameters post training.
+        :param calibration_function: An Optional callable to use for
+            calibration of module parameters post training. Should be able to
+            accept a batch of inputs along with a module.
+            Example: func(batch, module), Defaults to tensors_module_forward
         :param kwargs: Optional kwargs to support specific arguments
             for individual modifiers.
         """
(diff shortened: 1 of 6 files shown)


src/sparseml/pytorch/optim/modifier_quantization.py
```diff
@@ -429,6 +453,36 @@ def _enable_module_qat(self, module: Module):
...
+
+        model_device = next(module.parameters()).device
+
+        for batch in self._calibration_dataloader:
+            batch = tensors_to_device(batch, model_device)
+            with torch.no_grad():
+                try:
+                    forward_fn(batch)
+                except Exception:
+                    module(batch)
```
bfineran, line 482: let's remove the try catch and let the exception occur. especially if a custom function is used, this may be hard to debug

src/sparseml/pytorch/optim/modifier_quantization.py
```diff
@@ -292,11 +309,18 @@ def initialize(
         :param epoch: The epoch to initialize the modifier and module at.
             Defaults to 0 (start of the training process)
         :param loggers: Optional list of loggers to log the modification process to
+        :param calibration_dataloader: optional dataloader for running post training
+            quantization with the given model. if present, calibration will be run
+            immediately after quantization is enabled
+        :param calibration_function: optional function to use for calibration of
```
bfineran, line 315: add information about the default behavior as well as the expected format of the calibration function

bfineran (changes requested):
looks good overall, what happened to `num_calibration_steps`? It's important we have some modifier property that controls calibration so that we can throw an error if no dataloader is provided but calibration should be run

src/sparseml/pytorch/optim/modifier_quantization.py
rahul-tuli: Agreed fixed!

src/sparseml/pytorch/optim/modifier_quantization.py
```diff
@@ -428,6 +468,61 @@ def _enable_module_qat(self, module: Module):
...
+    def _calibrate_if_possible(self, module):
+        if self.num_calibration_steps == 0 and self._calibration_dataloader:
+            warnings.warn(
+                f"num_calibration_steps is {self.num_calibration_steps}."
+                f"Calibration data loader will not be used."
+            )
+        elif self.num_calibration_steps and not self._calibration_dataloader:
+            raise ValueError(
+                f"num_calibration_steps is {self.num_calibration_steps}. "
+                "Calibration data loader is not set."
```
bfineran, line 482: let's specify that it should be passed in as `calibration_dataloader` on initialize

src/sparseml/pytorch/optim/modifier_quantization.py
```diff
@@ -428,6 +468,61 @@ def _enable_module_qat(self, module: Module):
...
+        )
+        with suppress(StopIteration):
+            batch = next(_dataloader)
+            batch_number = 1
+            done = False
+            while not done:
+                batch = tensors_to_device(batch, model_device)
+                with torch.no_grad():
+                    forward_fn(batch, module=module)
+                done = (
```
bfineran, line 517: any reason not to `break`?

tests/sparseml/pytorch/optim/test_modifier_quantization.py
```diff
@@ -63,7 +63,7 @@ def _is_valid_submodule(module_name, submodule_names):
     )
```
bfineran, line 63: let's add `num_calibration_steps` to `test_quantization_modifier_yaml`

bfineran (changes requested):
changes look good

src/sparseml/pytorch/optim/modifier_quantization.py
rahul-tuli: Chose not to put a break cause it wouldn't add any extra optimizations

tests/sparseml/pytorch/optim/test_modifier_quantization.py
rahul-tuli: Added!

src/sparseml/pytorch/optim/modifier_quantization.py
rahul-tuli: Done!

src/sparseml/pytorch/optim/modifier_quantization.py
```diff
@@ -100,6 +115,8 @@ class QuantizationModifier(ScheduledModifier):
         transformer based models such as BERT where the quantized MatMul outputs
         are kept at 32 bits of precision and fake quantizing the outputs harm training
         recovery. Default is True
+    :param num_steps: Number of steps to run post training calibration for.
```
bfineran, line 118: nit: update name here

src/sparseml/pytorch/optim/modifier_quantization.py
bfineran: the done logic is a bit spread out along with while loop. thoughts against
```python
for idx, batch in enumerate(_dataloader):
    forward
    if (...):
        break
```

tests/sparseml/pytorch/optim/test_modifier_quantization.py
```diff
@@ -233,6 +235,7 @@ def test_quantization_modifier_yaml():
         quantize_embeddings=quantize_embeddings,
         reduce_range=reduce_range,
         quantize_linear_activations=quantize_linear_activations,
+        num_calibration_steps=num_calibration_steps,
```
bfineran, line 238: need to add asserts here for `num_calibration_steps` (follow logic below)

bfineran (dismissed):

src/sparseml/pytorch/optim/modifier_quantization.py
rahul-tuli: Done!

tests/sparseml/pytorch/optim/test_modifier_quantization.py
rahul-tuli: Added

src/sparseml/pytorch/optim/modifier_quantization.py
rahul-tuli: Good catch! thanks

src/sparseml/pytorch/optim/modifier_quantization.py
```diff
@@ -507,20 +506,16 @@ def _calibrate(self, module):
...
-                    forward_fn(batch, module=module)
-                done = (
-                    self.num_calibration_steps is not None
-                    and batch_number >= self.num_calibration_steps
-                )
-                batch = next(_dataloader)
-                batch_number += 1
+
+        for batch_idx, batch in enumerate(_dataloader):
+            _calibration_completed = (
```
bfineran, line 511: nit: put directly into if statement

rahul-tuli: Any particular reason why we should prefer that over this?

bfineran: if there's no reason to put it into a separate variable, we shouldn't break it out

rahul-tuli: My opinion is that it serves as a explanatory variable, describing why we are breaking out early, However will make this change if you insist.

PR #542, opened 2022-02-03
Activation Bits Support + qconfig kwargs for QuantizationModifier

rahul-tuli:
```python
import torch


# Pretrained Model

# Modifier init
modifier = QuantizationModifier(start_epoch=0, activation_bits=4)

# Modifier Application
modifier.apply(
    module=model,
)

for param in model.parameters():
    print(param.data.min(), param.data.max())

```

### DIFF
--- a/src/sparseml/pytorch/optim/modifier_quantization.py
+++ b/src/sparseml/pytorch/optim/modifier_quantization.py
@@ -86,7 +86,7 @@ class QuantizationModifier(ScheduledModifier):
     |       disable_quantization_observer_epoch: 2.0
     |       freeze_bn_stats_epoch: 3.0
     |       reduce_range: False
-    |       activation_bits: False
+    |       activation_bits: 4
 
     :param start_epoch: The epoch to start the modifier at
     :param submodules: List of submodule names to perform QAT on. Leave None to quantize
@@ -117,7 +117,7 @@ class QuantizationModifier(ScheduledModifier):
         are kept at 32 bits of precision and fake quantizing the outputs harm training
         recovery. Default is True
     :param activation_bits: Number of bits to use for setting quant min/max values for
-            activations. Default is None, which will quantize activations to 8 bits.
+            activations. Default is  8 bits.
     :param num_calibration_steps: Number of steps to run post training calibration for.
             When None, the entire calibration_dataloader is used
     :param exclude_module_types: optional list of module class names
@@ -140,7 +140,7 @@ class QuantizationModifier(ScheduledModifier):
         quantize_embeddings: bool = True,
         reduce_range: bool = False,
         quantize_linear_activations: bool = True,
-        activation_bits: Optional[int] = None,
+        activation_bits: int = 8,
         num_calibration_steps: Optional[int] = None,
         exclude_module_types: Union[List[str], None] = None,
         activation_qconfig_kwargs: Optional[Dict[str, Any]] = None,
@@ -599,6 +599,12 @@ class QuantizationModifier(ScheduledModifier):
             if self.activation_qconfig_kwargs
             else {}
         )
+
+        if not isinstance(self.activation_bits, int) or self.activation_bits < 0:
+            raise ValueError(
+                f"activation_bits must be a positive integer, "
+                f"but got {self.activation_bits}"
+            )
 
         # update qconfig_kwargs for activation_bits
         if self.activation_bits and (
--- a/src/sparseml/pytorch/optim/modifier_quantization.py
+++ b/src/sparseml/pytorch/optim/modifier_quantization.py
@@ -24,6 +24,7 @@ import torch
 from torch.nn import Module
 from torch.optim.optimizer import Optimizer
 
+
 try:
     from torch import quantization as torch_quantization
     from torch.nn import intrinsic as torch_intrinsic
@@ -32,8 +33,7 @@ except Exception:
     torch_intrinsic = None
 
 from sparseml.optim import BaseModifier, ModifierProp
-from sparseml.pytorch.optim.modifier import PyTorchModifierYAML, \
-    ScheduledModifier
+from sparseml.pytorch.optim.modifier import PyTorchModifierYAML, ScheduledModifier
 from sparseml.pytorch.utils import BaseLogger
 from sparseml.pytorch.utils.quantization import (
     add_quant_dequant,
@@ -46,6 +46,7 @@ from sparseml.pytorch.utils.quantization import (
     remove_activation_qat_by_layer_name,
 )
 from sparseml.sparsification import SparsificationTypes
+
 
 __all__ = [
     "QuantizationModifier",
@@ -71,7 +72,7 @@ class QuantizationModifier(ScheduledModifier):
     |       disable_quantization_observer_epoch: 2.0
     |       freeze_bn_stats_epoch: 3.0
     |       reduce_range: False
-    |       int4_activations: False
+    |       activation_bits: False
 
     :param start_epoch: The epoch to start the modifier at
     :param submodules: List of submodule names to perform QAT on. Leave None to quantize
@@ -96,8 +97,8 @@ class QuantizationModifier(ScheduledModifier):
     :param reduce_range: if True, the quantization range will be reduced by one bit.
         This may prevent overflow issues with model execution on certain hardware
         Default is False
-    :param int4_activations: if True, will enable 4 bit quantization for
-        activations. Default is False.
+    :param activation_bits: Number of bits to use for setting quant min/max values for
+            activations. Default is None, which will quantize activations to 8 bits.
     :param quantize_linear_activations: if False, FakeQuantize ops will not be run
         for activations of fully connected layers. this is important for quantizing
         transformer based models such as BERT where the quantized MatMul outputs
@@ -108,21 +109,21 @@ class QuantizationModifier(ScheduledModifier):
     """
 
     def __init__(
-            self,
-            start_epoch: float = -1.0,
-            submodules: Union[List[str], None] = None,
-            model_fuse_fn_name: Union[str, None] = None,
-            disable_quantization_observer_epoch: Union[float, None] = None,
-            freeze_bn_stats_epoch: Union[float, None] = None,
-            end_epoch: float = -1,
-            model_fuse_fn_kwargs: Dict[str, Any] = None,
-            quantize_embeddings: bool = True,
-            reduce_range: bool = False,
-            quantize_linear_activations: bool = True,
-            int4_activations: bool = False,
-            exclude_module_types: Union[List[str], None] = None,
-            activation_qconfig_kwargs: Dict[str, Any] = {},
-            weight_qconfig_kwargs: Dict[str, Any] = {},
+        self,
+        start_epoch: float = -1.0,
+        submodules: Union[List[str], None] = None,
+        model_fuse_fn_name: Union[str, None] = None,
+        disable_quantization_observer_epoch: Union[float, None] = None,
+        freeze_bn_stats_epoch: Union[float, None] = None,
+        end_epoch: float = -1,
+        model_fuse_fn_kwargs: Dict[str, Any] = None,
+        quantize_embeddings: bool = True,
+        reduce_range: bool = False,
+        quantize_linear_activations: bool = True,
+        activation_bits: Optional[int] = None,
+        exclude_module_types: Union[List[str], None] = None,
+        activation_qconfig_kwargs: Dict[str, Any] = {},
+        weight_qconfig_kwargs: Dict[str, Any] = {},
     ):
         if torch_quantization is None or torch_intrinsic is None:
             raise RuntimeError(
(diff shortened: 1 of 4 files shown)


src/sparseml/pytorch/optim/modifier_quantization.py
```diff
@@ -114,6 +117,7 @@ def __init__(
         quantize_embeddings: bool = True,
         reduce_range: bool = False,
         quantize_linear_activations: bool = True,
+        enable_int4_activations: bool = False,
```
bfineran, line 120: nit: let's remove `enable_` from the name, since it's implied

src/sparseml/pytorch/utils/quantization/helpers.py
```diff
@@ -98,16 +97,24 @@ class QATWrapper(Module):
...
+    :param activation_kwargs: Additional kwargs for quantization of activations.
+        Default is {}
     """
 
     @staticmethod
-    def from_module(module: Module, reduce_range: bool = None) -> "QATWrapper":
+    def from_module(
+        module: Module,
+        reduce_range: bool = None,
+        activation_kwargs: Dict[str, Any] = {},
```
bfineran, line 108: nit: rename to `activation_qconfig_kwargs` since it's just for the qconfigs

bfineran: also need to add a param for the weights

src/sparseml/pytorch/utils/quantization/helpers.py
```diff
@@ -279,13 +303,18 @@ def _load_qconfigs(
...
             )
 
         return qconfigs
 
 
-def configure_module_qat_wrappers(module: Module, reduce_range: bool = False):
+def configure_module_qat_wrappers(
+    module: Module,
+    reduce_range: bool = False,
+    activation_kwargs: Dict[str, Any] = {},
```
bfineran, line 316: also should rename to activation_qconfig_kwargs and add a param for weights

src/sparseml/pytorch/utils/quantization/helpers.py
```diff
@@ -372,6 +411,7 @@ def get_qat_qconfig(
     symmetric_activations: bool = False,
     symmetric_weights: bool = True,
     reduce_range: bool = False,
+    activation_kwargs: Dict[str, Any] = {},
```
bfineran, line 414: should add arg for `weight_kwargs`

src/sparseml/pytorch/utils/quantization/helpers.py
```diff
@@ -501,6 +548,7 @@ def prepare_embeddings_qat(
     module: Module,
     qconfig: "torch.quantization.QConfig" = None,
     reduce_range: bool = False,
+    activation_kwargs: Dict[str, Any] = {},
```
bfineran, line 551: also should rename to activation_qconfig_kwargs and add a param for weights

src/sparseml/pytorch/optim/modifier_quantization.py
bfineran, line 120: We also need to add modifier properties here for `activation_qconfig_kwargs` and `weight_qconfig_kwargs` as discussed.

src/sparseml/pytorch/optim/modifier_quantization.py
```diff
@@ -138,11 +142,15 @@ def __init__(
...
         self._quantize_linear_activations = quantize_linear_activations
+        self._enable_int4_activations = enable_int4_activations
 
         self._modules_to_quantize = None
         self._qat_enabled = False
         self._quantization_observer_disabled = False
         self._bn_stats_frozen = False
+        self._quant_min = 0
+        self._quant_max = 255
+        self._dtype = torch.quint8
```
bfineran, line 153: these shouldn't be stored here. instead the activation_kwargs should be updated conditionally in `_enable_module_qconfig`

src/sparseml/pytorch/optim/modifier_quantization.py
```diff
@@ -278,6 +291,27 @@ def quantize_linear_activations(self) -> bool:
...
+        return self._enable_int4_activations
+
+    @ModifierProp()
+    def qconfig_activation_kwargs(self) -> Dict[str, Any]:
+        """
+        :return: Dictionary with correct quant_min, quant_max, and dtype values
+            for activations
+
+        """
+        return dict(
```
bfineran, line 309: this should not be computed dynamically. modifier props need to be creatable from recipe. (see above comments about adding a modifier prop for this and the weights

bfineran (changes requested):
as discussed we need to have modifier properties to control both the weight and activation qconfig kwargs

src/sparseml/pytorch/optim/modifier_quantization.py
rahul-tuli: Done!

bfineran:
@rahul-tuli pushed the discussed fix for propagating observer quant range

src/sparseml/pytorch/optim/modifier_quantization.py
```diff
@@ -138,11 +143,15 @@ def __init__(
         self._quantize_embeddings = quantize_embeddings
         self._reduce_range = reduce_range
         self._quantize_linear_activations = quantize_linear_activations
+        self._enable_int4_activations = enable_int4_activations
```
natuan, line 146: Does it make sense not to hard code number of bits and the min/max range? Sth like "activation_bits" defining number of bits looks nicer to me for the recipe (also who knows in the future we want to do mix quantization)?

natuan: If we can make it more general, I'm thinking we could also use this field to say that by default we don't quantize activation of a module. What we learned from quantizing BERT is that module/function activations should not be quantized (unless there're good reasons) to avoid rounding errors (this happens when the activation signals go into another module that need to be quantized, and then we have unnecessary double quant/dequant modules).

bfineran: great point on `activation_bits` @natuan. let's save the second suggestion for a future update since this controls quantization of all modules and we do need activation quantization to some level in our models still

src/sparseml/pytorch/optim/modifier_quantization.py
rahul-tuli: Do you mean `_enable_module_qat`?

src/sparseml/pytorch/utils/quantization/helpers.py
rahul-tuli: Added!

rahul-tuli:
> @rahul-tuli pushed the discussed fix for propagating observer quant range

Thanks, incorporated the requested changes!

bfineran (changes requested):

src/sparseml/pytorch/optim/modifier_quantization.py
```diff
@@ -95,6 +97,8 @@ class QuantizationModifier(ScheduledModifier):
     :param reduce_range: if True, the quantization range will be reduced by one bit.
         This may prevent overflow issues with model execution on certain hardware
         Default is False
+    :param activation_bits: Number of bits to use for setting quant min/max values for
```
bfineran, line 100: nit: should go below `quantize_linear_activations`

src/sparseml/pytorch/optim/modifier_quantization.py
```diff
@@ -95,6 +97,8 @@ class QuantizationModifier(ScheduledModifier):
...
+            activations. Default is None, which will quantize activations to 8 bits.
```
bfineran, line 101: need docs for `activation_qconfig_kwargs` and `weight_qconfig_kwargs`

src/sparseml/pytorch/optim/modifier_quantization.py
```diff
@@ -442,10 +489,46 @@ def _enable_module_qat(self, module: Module):
...
+            or self.activation_qconfig_kwargs.get("quant_max")
+        ):
+            raise ValueError(
+                "Cannot override quant_max and quant_min with activation_bits enabled"
+            )
+
+        activation_qconfig_kwargs = self.activation_qconfig_kwargs.copy()
+        quant_min = activation_qconfig_kwargs.get("quant_min", 0)
+        quant_max = activation_qconfig_kwargs.get("quant_max", 2 ** 8 - 1)
+        dtype = activation_qconfig_kwargs.get("dtype", torch.quint8)
```
bfineran, line 517: no need to set a default here when a user does not override activaiton_bits or the kwargs, that will be handled later in our pipeline. let's delete this block and move the dict update inside the if statement

src/sparseml/pytorch/optim/modifier_quantization.py
```diff
@@ -442,10 +489,46 @@ def _enable_module_qat(self, module: Module):
...
+
+        if self.activation_bits:
+            quant_min = 0
+            quant_max = 2 ** self.activation_bits - 1
+
+        activation_qconfig_kwargs.update(
```
bfineran, line 523: move inside if statement so we only update if activation_bits is set

src/sparseml/pytorch/utils/quantization/helpers.py
```diff
@@ -372,6 +436,8 @@ def get_qat_qconfig(
     symmetric_activations: bool = False,
     symmetric_weights: bool = True,
     reduce_range: bool = False,
+    activation_qconfig_kwargs: Dict[str, Any] = {},
+    weight_qconfig_kwargs: Dict[str, Any] = {},
```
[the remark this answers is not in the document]
bfineran: bumping this

tests/sparseml/pytorch/optim/test_modifier_quantization.py
```diff
@@ -88,6 +92,7 @@ def _test_quantizable_module(module, qat_expected, modifier):
         )
```
bfineran, line 92: let's add a test here for `if modifier.activation_bits is not None` and check that the qconfig activation range is correct

bfineran (changes requested):
looks close! left a few comments then good from my end

src/sparseml/pytorch/utils/quantization/helpers.py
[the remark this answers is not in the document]
rahul-tuli: Done!

src/sparseml/pytorch/optim/modifier_quantization.py
rahul-tuli: Done

tests/sparseml/pytorch/optim/test_modifier_quantization.py
rahul-tuli: Done!

tests/sparseml/pytorch/optim/test_modifier_quantization.py
```diff
@@ -211,6 +223,7 @@ def test_quantization_modifier_yaml():
     quantize_linear_activations = False
     num_calibration_steps = 2
     exclude_module_types = ["LayerNorm", "Tanh"]
+    activation_bits = 4
```
bfineran, line 226: let's also test the qconfig kwargs here

bfineran (commented):
one more comment then good to go

tests/sparseml/pytorch/optim/test_modifier_quantization.py
rahul-tuli: Done!

bfineran (approved):

KSGulin (approved):
Good stuff

KSGulin (approved):

bfineran (dismissed):

src/sparseml/pytorch/optim/modifier_quantization.py
```diff
@@ -85,6 +86,7 @@ class QuantizationModifier(ScheduledModifier):
     |       disable_quantization_observer_epoch: 2.0
     |       freeze_bn_stats_epoch: 3.0
     |       reduce_range: False
+    |       activation_bits: False
```
natuan, line 89: False is not in its domain

src/sparseml/pytorch/optim/modifier_quantization.py
```diff
@@ -114,10 +116,16 @@ class QuantizationModifier(ScheduledModifier):
         transformer based models such as BERT where the quantized MatMul outputs
         are kept at 32 bits of precision and fake quantizing the outputs harm training
         recovery. Default is True
+    :param activation_bits: Number of bits to use for setting quant min/max values for
+            activations. Default is None, which will quantize activations to 8 bits.
```
natuan, line 120: I think it would be confusing that None means 8 bits (to me it means you don't want to quantize activations at all); we should make it explicit to see from the recipe. Can't we default to number?

src/sparseml/pytorch/optim/modifier_quantization.py
rahul-tuli: Agree, thanks.

natuan (changes requested):

src/sparseml/pytorch/optim/modifier_quantization.py
```diff
@@ -114,10 +116,16 @@ class QuantizationModifier(ScheduledModifier):
         transformer based models such as BERT where the quantized MatMul outputs
         are kept at 32 bits of precision and fake quantizing the outputs harm training
         recovery. Default is True
+    :param activation_bits: Number of bits to use for setting quant min/max values for
```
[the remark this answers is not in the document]
rahul-tuli: Synced Offline so resolving!

bfineran (approved):

PR #545, opened 2022-02-07
Enforce order on input keys to export

natuan:
Putting back the order of input's keys passing into the pytorch onnx export

### DIFF
--- a/src/sparseml/transformers/export.py
+++ b/src/sparseml/transformers/export.py
@@ -183,8 +183,10 @@ def export_transformer_to_onnx(
         "", return_tensors="pt", padding=PaddingStrategy.MAX_LENGTH.value
     ).data  # Dict[Tensor]
 
-    # Rearrange inputs' keys to match those defined by model foward func
+    # Rearrange inputs' keys to match those defined by model foward func, which
+    # seem to define how the order of inputs is determined in the exported model
     forward_args_spec = inspect.getfullargspec(model.__class__.forward)
+    dropped = [f for f in inputs.keys() if f not in forward_args_spec.args]
     inputs = collections.OrderedDict(
         [
             (f, inputs[f][0].reshape(1, -1))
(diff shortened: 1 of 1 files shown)


bfineran (dismissed):

markurtz (changes requested):
@natuan need to regenerate the input_shapes arg based on the new inputs that will be passed through so we log properly. Additionally, can we add a warning in the event that the keys in the inputs do not match the keys for the forwards.

natuan:
> @natuan need to regenerate the input_shapes arg based on the new inputs that will be passed through so we log properly. Additionally, can we add a warning in the event that the keys in the inputs do not match the keys for the forwards.

I fixed the first one you mentioned. For the warning, the input output by the tokenizer is a Dict so by issuing that warning we imply an ordering over it right? (which is what this fix is) In the pytorch/utils/exporter we did have this type of warning if the input batch is a Dict but not an OrderedDict

markurtz:
> > @natuan need to regenerate the input_shapes arg based on the new inputs that will be passed through so we log properly. Additionally, can we add a warning in the event that the keys in the inputs do not match the keys for the forwards.
> 
> I fixed the first one you mentioned. For the warning, the input output by the tokenizer is a Dict so by issuing that warning we imply an ordering over it right? (which is what this fix is) In the pytorch/utils/exporter we did have this type of warning if the input batch is a Dict but not an OrderedDict

I was mainly concerned that we are now technically filtering out inputs based on whether or not they are defined in the forward function. This is something that might be a good thing to do, but we should have a warning in place to say that the inputs were filtered and not everything was passed through. On the other side, if not all inputs are found that are defined in the forward, it would be good to have this called out as well in the event that some have defaults and therefore would not fail the forward

markurtz (approved):

bfineran (approved):

markurtz (approved):

PR #546, opened 2022-02-07
pruning modifiers refactor

bfineran:
Refactors pruning modifiers to `src/sparseml/pytorch/sparsification/pruning`
Pruning modifiers will now override a base `BasePruningModifier` or `BaseGradualPruningModifier` class to implement pruning algorithms.

The main entry point to implement pruning algorithms will be by overriding `_get_mask_creator` `_get_scorer` and `get_applied_sparsity_for_epoch` to affect mask creation, parameter scoring, and applied sparsity levels respectively.  Entry-points to the training lifecycle are provided in the `BasePruningModifier` and the `PruningParamsScorer` classes.

During migration, some implementations in `src/sparseml/pytorch/optim/pruning` will mirror their pre-refactored forms in `src/sparseml/pytorch/optim`, however, the pruning classes in the `optim` directory will be removed and the `pruning` directory modules will be propagated up in the `pytorch/optim` `__init__.py` file after the migration is completed.  Modifiers in `modifier_pruning.py` will be deleted and imports moved to `optim/pruning` as completed. The existing `GMPruningModifier` which multiple other pruning algorithms use as a parent class has been aliased to `LegacyPruningModifier` during the migration and will be deleted after.

Migrated implementations of `GMPruningModifier` (now only layer-wise magnitude also aliased to `MagnitudePruningModifier`), `GlobalMagnitudePruningModifier`, and `ConstantPruningModifier` are included in this PR.  Migrations of ACDC (currently `phased=True`), `MovementPruningModifier`, and `MFACPruningModifier` will follow.

Example usage of the new base class to implement magnitude pruning is provided below:
```python
class MagnitudePruningParamsScorer(PruningParamsScorer):

    def score_parameters(self) -> List[Tensor]:
        return [torch.abs(param.data) for param in self._params]


@PyTorchModifierYAML()
class GMPruningModifier(BaseGradualPruningModifier, BaseGMPruningModifier):
    def __init__(
        self,
        init_sparsity: float,
        final_sparsity: Union[float, Dict[float, List[str]]],
        start_epoch: float,
        end_epoch: float,
        update_frequency: float,
        params: Union[str, List[str]],
        leave_enabled: bool = True,
        inter_func: str = "cubic",
        log_types: Union[str, List[str]] = ALL_TOKEN,
        mask_type: str = "unstructured",
    ):
        super(GMPruningModifier, self).__init__(...)

    def _get_mask_creator(self) -> PruningMaskCreator:
        if self.mask_type == "unstructured":
            return UnstructuredPruningMaskCreator()
        elif self.mask_type == "block":
            return FourBlockMaskCreator()
        else:
            raise ValueError(
                f"Unknown mask_type {self.mask_type}. Supported mask types include "
                "'unstructured' and 'block'"
            )

    def _get_scorer(self, params: List[Parameter]) -> PruningParamsScorer:
        return MagnitudePruningParamsScorer(params)
```


bfineran:
WIP pending quality checks and tests

markurtz (approved):

src/sparseml/sparsification/modifier_pruning.py
```diff
@@ -27,7 +27,12 @@
     ModifierProp,
 )
 from sparseml.sparsification.types import SparsificationTypes
-from sparseml.utils import ALL_TOKEN, convert_to_bool, validate_str_iterable
+from sparseml.utils import (
+    ALL_TOKEN,
+    FROM_PARAM_TOKEN,
```
dbogunowicz, line 32: What's the function of `FROM PARAM TOKEN`?

bfineran: this is a keyword for `init_sparsity` that will initialize sparsity levels in a pruning modifier to the existing sparsity in the model

dbogunowicz (approved):
Looks good to me.

src/sparseml/pytorch/sparsification/pruning/mask_creator.py
```diff
@@ -0,0 +1,424 @@
...
+        reduced_blocks = GroupedPruningMaskCreator.reduce_tensor(
+            blocked_tensor, 1, self._grouping_fn_name
+        )
+        return reduced_blocks.type(tensor.type())
+
+    def _map_mask_to_tensor(
+        self,
+        grouped_mask: Tensor,
+        original_tensor_shape: torch.Size,
+        tensor_idx: Optional[int] = None,
```
KSGulin, line 385: `tensor_idx` doesn't get used at the moment. Is it included now for the future?

src/sparseml/pytorch/sparsification/pruning/modifier_pruning_constant.py
```diff
@@ -0,0 +1,148 @@
...
+        end_epoch: float = -1.0,
+        update_frequency: float = -1.0,
+        log_types: Union[str, List[str]] = ALL_TOKEN,
+    ):
+        super(ConstantPruningModifier, self).__init__(
+            params=params,
+            start_epoch=start_epoch,
+            end_epoch=end_epoch,
+            end_comparator=-1,
+            update_frequency=-1,
```
KSGulin, line 127: Should this be `update_frequency=update_frequency` since it already defaults to one in `__init__`?

KSGulin (commented):
I'm a big fan of this refactor. Looks really good on the high level, found just a few nits.

markurtz (approved):
Looks good, great job on this Ben!

src/sparseml/pytorch/optim/pruning/modifier_pruning_base.py
```diff
@@ -0,0 +1,755 @@
...
+            # get sparsity level to be applied
+            self._applied_sparsity = self.get_applied_sparsity_for_epoch(
+                epoch, steps_per_epoch
+            )
+
+            self._module_masks.set_param_masks_from_sparsity(self._applied_sparsity)
+            self._sparsity_applied = True
+
+        if self.end_pending(epoch, steps_per_epoch):
+            self._module_masks.pruning_end(self._leave_enabled)
```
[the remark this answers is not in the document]
bfineran: great catch, this was added in an update since when I believe this comment was added. check the latest init function for base pruning modifier

src/sparseml/pytorch/sparsification/pruning/mask_creator.py
bfineran: this can be used for other implementing classes, not currently used for block masks

src/sparseml/pytorch/sparsification/pruning/modifier_pruning_constant.py
bfineran: good catch, there are no updates, so taking this out for ConstantPruningModifier

bfineran: actually this is needed for inheritance reasons, and is documented as dead functionality above

KSGulin (approved):
LGTM!

PR #550, opened 2022-02-10
Fix QATConv3D

KSGulin:
Fix to allow SparseML model import to run If Pytorch version is not recent enough to have QATConv3d (introduced after v1.10, not currently in stable release)


markurtz (commented):
I believe we need to handle this conditional later in the code as well where we check the types for quantized modules

KSGulin:
> I believe we need to handle this conditional later in the code as well where we check the types for quantized modules

If I'm not mistaken, the one issue that could arise from this is that we load in a model with a quantized Conv3D layer with a version of PyTorch which doesn't support QATConv3D. What behavior would we want to default to? To crash with a useful error, or is there a more productive way to handle it?

bfineran:
@KSGulin we'll need a conditional on line 798.
I think it would be ok to raise an exception in `get_quantized_layers` if we detect a `Conv3d` object with a qconfig but `QATConv3d` is not available, since upgrading to a newer version of torch would get that functionality

KSGulin:
@bfineran thanks for pointing that out! I opted for a user warning because it looks like Conv3D only shows up in the latest main (unstable) PyTorch docs and isn't in the docs for the latest stable branch (1.10). Let me know if you'd like me to omit the warning or change it to an exception

rahul-tuli (approved):
LGTM!

markurtz (approved):

PR #556, opened 2022-02-16
Ability to load recipe stages from YAML into dictionary of name -> List[modifiers]

dbogunowicz:
This diff allows the modifiers to additionally digest stage recipes. An example of a staged recipe is:

```
high_level_variable: 0

stage 1:
  training_modifiers:
  - !EpochRangeModifier
    start_epoch: eval(high_level_variable)
    end_epoch: 3
  
  pruning_modifiers:
  - !ACDCPruningModifier
     compression_sparsity: 0.9
     start_epoch: 0
     end_epoch: 3
     update_frequency: 1
     params: ['re:.*conv*', 're:.*fc.weight*']

stage 2:
  modifiers:
    - !EpochRangeModifier
        end_epoch: 3
        start_epoch: 5
    - !GMPruningModifier
        start_epoch: 3
        end_epoch: 5
        final_sparsity: 0.1
        init_sparsity: 0.1  
        update_frequency: 1
        params: ['re:.*conv*', 're:.*fc.weight*']
```
A staged recipe consists of:
- definition of global (main) variables, which trickle down to separate stages.
- definition of multiple stages. Each stage can contain its own set of pruning/training modifiers.


dbogunowicz:
Pushed a stage file `staged_recipe.md` to facilitate communication.
`staged_recipe.md` will get evaluated to the following container:

image

Pretty-printed:

image

src/sparseml/optim/helpers.py
```diff
@@ -219,13 +219,98 @@ def update_recipe_variables(recipe_yaml_str: str, variables: Dict[str, Any]) ->
...
+    stage_signature = "_stage"
+    modifier_signature = "_modifier"
+
+    if not any([stage_signature in key for key in container.keys()]):
+        # if there are no stages in the container, this must be standard recipe, pass
+        return container, not staged_recipe
+
+    else:
+        # parsing the container to create stage_container
+        stage_container = {"global_variables": []}
```
bfineran, line 275: what is the goal for structure and usage of `stage_container` here?

dbogunowicz: Regarding the goal for structure, I am still not sure how do we treat `global_variables` and "`stage_variables`".

From our discussion yesterday I understand that we want to break down the staged recipe into a single list of modifiers. Those modifiers will get activated/deactivated according to e.g. their start/end epoch (their order in the compound sparsity training routine). 

This means that the YAML definitions of a modifier in the list should probably be in its' final stage: by the time it gets passed to the manager, it should have evaluated modifier arguments incl. appropriate treatment of `global_variables` and `stage_variables`. 

The pending question is: given, that our goal structure of `staged_container` is a dictionary {`stage_name`: `List[modifiers]`}, should we include the variables (both global and stage variables) into their respective modifiers? I'd love to sync with you on that one.

dbogunowicz: Let's maybe discuss this more holistically today, so I have bird's-eye view over the mechanism ok?

src/sparseml/optim/helpers.py
```diff
@@ -230,18 +230,63 @@ def evaluate_recipe_yaml_str_equations(recipe_yaml_str: str) -> str:
...
+        non_val_variables = {**non_val_variables, **global_non_val_variables}
+
+        for key, val in container.items():
+            if "modifiers" not in key:
+                continue
+            container[key] = _maybe_evaluate_yaml_object(
+                val, variables, non_val_variables
+            )
+
+        consolidated_container.update(container)
```
bfineran, line 285: instead of consolidated_container update, we should set `container[stage_name] = stage_container` after the `stage_container` updates happen

bfineran (changes requested):
good flow overall, left comments on finalizing some of the logic
also please make sure to rebase onto main now that the manager logic has been updated for stages!

tests/sparseml/optim/test_helpers.py
```diff
@@ -117,21 +163,29 @@ def _test_nested_equality(val, other):
...
-        (RECIPE_SIMPLE_EVAL, TARGET_RECIPE.format(num_epochs=10.0)),
-        (RECIPE_MULTI_EVAL, TARGET_RECIPE.format(num_epochs=10.0)),
+        (
+            TARGET_RECIPE.format(num_epochs=10.0),
+            TARGET_RECIPE.format(num_epochs=10.0),
+            False,
+        ),
+        (RECIPE_SIMPLE_EVAL, TARGET_RECIPE.format(num_epochs=10.0), False),
+        (RECIPE_MULTI_EVAL, TARGET_RECIPE.format(num_epochs=10.0), False),
+        (STAGED_RECIPE, STAGED_RECIPE, True),
```
bfineran, line 175: the given and expected recipe should not be the same, let's make sure that we get the expected recipe added and verify that the proper checks are made so the test only passes when the eval is correct

staged_recipe.md
```diff
@@ -0,0 +1,45 @@
+---
+high_level_variable: 1
```
bfineran, line 2: nit: make sure to remove test file

bfineran (changes requested):

src/sparseml/optim/helpers.py
```diff
@@ -230,18 +230,63 @@ def evaluate_recipe_yaml_str_equations(recipe_yaml_str: str) -> str:
...
+        global_non_val_variables,
+    ) = _evaluate_recipe_variables(main_container)
+
+    for stage_name, stage_container in stages.items():
+        container, variables, non_val_variables = _evaluate_recipe_variables(
+            stage_container
+        )
+
+        variables = {**variables, **global_variables}
+        non_val_variables = {**non_val_variables, **global_non_val_variables}
```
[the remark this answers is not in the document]
dbogunowicz: Yes, this is the dict union as used by python3.5+. But the new one for python 3.9, which is even cooler :)

https://stackoverflow.com/questions/38987/how-do-i-merge-two-dictionaries-in-a-single-expression-take-union-of-dictionari

image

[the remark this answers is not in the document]
bfineran: awesome! just want to note that we should be backwards compatible through python 3.6, so cannot use any 3.10 specific syntax

src/sparseml/optim/helpers.py
```diff
@@ -230,18 +231,82 @@ def evaluate_recipe_yaml_str_equations(recipe_yaml_str: str) -> str:
...
+    - a stage should map to a dict that has at least one key with
+      "modifiers" in its name
+    :param container: a container generated from a YAML string of SparseML recipe
+    :return: True if stage recipe, False if normal recipe
+    """
+
+    if any([key for key in container.keys() if "modifiers" in key]):
+        return False
+    for stage, stage_dict in container.items():
+        if "stage" in stage:
```
bfineran, line 266: for now, "stage" is not a requirement for stage naming. just that the value is a dict with at least one key with `modifiers` in the name

src/sparseml/optim/helpers.py
```diff
@@ -230,18 +231,82 @@ def evaluate_recipe_yaml_str_equations(recipe_yaml_str: str) -> str:
...
+def _evaluate_staged_recipe_yaml_str_equations(container: dict) -> dict:
+    """
+    Consumes a staged container and transforms it into a valid
+    container for the manager and modifiers to consume further.
+
+    :param container: a staged container generated from a staged recipe.
+    :return: transformed container containing evaluated
+            variables, operations and objects.
+    """
+    main_container = {k: v for k, v in container.items() if "stage" not in k}
```
bfineran, line 281: same as above, the requirement is that the value is a dict with at least one key with `modifiers` in the name

tests/sparseml/optim/test_helpers.py
```diff
@@ -15,13 +15,90 @@
...
 from sparseml.optim import (
+    check_if_staged_recipe,
     evaluate_recipe_yaml_str_equations,
     load_recipe_yaml_str,
     load_recipe_yaml_str_no_classes,
     update_recipe_variables,
 )
 
 
+STAGED_RECIPE = """
```
bfineran, line 26: looks like a good test case, let's get another one in with a few more edge cases like stage scope variable names conflicting with each other

bfineran (commented):
looks really close, main comment is to remove the `stage` requirement for a subcontainer to be considered a stage

dbogunowicz:
@bfineran: regarding the dictionary merging function, it is compatible with > python3.6. This is even confirmed in CI/CD pipeline where python 3.6 is being run. All good:

image

tests/sparseml/optim/test_helpers.py
dbogunowicz: I have done so. The test fail to capture the logic which I would expected when variable names conflict. I would like to discuss this logic shortly.

bfineran (approved):

src/sparseml/optim/helpers.py
```diff
@@ -230,18 +230,100 @@ def evaluate_recipe_yaml_str_equations(recipe_yaml_str: str) -> str:
...
 
 
+def check_if_staged_recipe(container: dict) -> bool:
+    """
+    Check whether container pertains to a staged recipe.
+    Such a "staged container" fulfills two conditions:
+    - no top level key in container contains "modifiers" in its name
+    - a stage should map to a dict that has at least one key with
+      "modifiers" in its name
+    :param container: a container generated from a YAML string of SparseML recipe
```
rahul-tuli, line 260: nit: need newline before params for `sphinx`

src/sparseml/optim/helpers.py
```diff
@@ -230,18 +230,100 @@ def evaluate_recipe_yaml_str_equations(recipe_yaml_str: str) -> str:
...
+    :return: True if stage recipe, False if normal recipe
```
rahul-tuli, line 261: I don't like the word `normal` for a recipe without stages, Could we simplify this by saying something like:
`True if staged recipe, False otherwise`

src/sparseml/optim/helpers.py
```diff
@@ -301,26 +387,26 @@ def _evaluate_recipe_variables(
 
             try:
                 val = _maybe_evaluate_recipe_equation(
-                    val, valid_variables, non_evaluatable_variables
+                    val, valid_variables, non_evaluatable_variables, global_container
                 )
             except UnknownVariableException:
                 # dependant variables maybe not evaluated yet
                 continue
```
rahul-tuli, line 394: Is `with suppress(UnknownVariableException)` a cleaner alternative?

rahul-tuli: it can be imported from `contextlib`

rahul-tuli (approved):
Approved with two small suggested nits if they make sense

PR #559, opened 2022-02-16
Update base pruning args

KSGulin:
Came across an issue with multiple inheritance in the pruning refactor. When creating specific pruning implementations (e.g. GMP, ACDC, M-FAC) these implementations can become at odds with each other because of a collision between:
1. The Python multiple inheritance requirement of every parent class's `__init__()` being able to accept the same set of args (usually via `kwargs`)
2. SparseML requirement to not propagate any `kwargs` to `BaseObject`

This fix avoids the issues by allowing pruning implementations to specify a specific subset of `kwargs` to propagate


src/sparseml/pytorch/sparsification/pruning/modifier_pruning_base.py
```diff
@@ -117,10 +117,16 @@ def __init__(
         log_types: Union[str, List[str]] = None,
         global_sparsity: bool = False,
         allow_reintroduction: bool = False,
+        explicit_kwargs: Optional[List[str]] = ["params"],
```
bfineran, line 120: cannot default to a mutable object, let's default to `None` instead. Behavior for `None` should be to not propagate any extra kwargs. users who need `params` propagated should pass that explicitly

bfineran, line 120: `explicit` doesn't give us any extra information here, could we rename to `parent_class_kwarg_names` or something similar

src/sparseml/pytorch/sparsification/pruning/modifier_pruning_magnitude.py
```diff
@@ -135,6 +135,12 @@ def __init__(
             end_comparator=-1,
             global_sparsity=self._use_global_sparsity,
             allow_reintroduction=False,
+            explicit_kwargs=[
```
bfineran, line 138: looks good, should also add `["params"]` for `ConstantPruningModifier`

bfineran (changes requested):

bfineran (approved):

dbogunowicz (approved):
Excellent job Konstantin, thank you!

PR #563, opened 2022-02-18
Add support to total batch size argument for transformers transfer learning

anmarques:
Created the NMTrainingArguments that inherits from HF's TrainingArguments. This class allows one to add arguments to the training script and handling potential conflicts with other arguments. In particular, added "total_train_batch_size" and "total_eval_batch_size" and made sure these arguments cannot be used alongside "per_device_train_batch_size" and "per_device_eval_batch_size."


rahul-tuli (dismissed):

src/sparseml/transformers/nm_training_arguments.py
```diff
@@ -0,0 +1,96 @@
...
+        # set total batch sizes as default
+        if not self.total_train_batch_size and not self.per_device_train_batch_size:
+            self.total_train_batch_size = 8
+        if not self.total_eval_batch_size and not self.per_device_eval_batch_size:
+            self.total_eval_batch_size = 8
+
+        # If total batch size is defined at this point,
+        # per device batch size needs to be computed
+        if self.total_train_batch_size:
+            self.per_device_train_batch_size = self.total_train_batch_size // max(
```
natuan, line 90: Would it be better if we use this total_train_batch_size to dictate both the per_device_train_batch_size and the gradient_accumulation_steps (so the "steps" here mean backward steps)---this way we make sure that this total batch size, if specified, is consistent even between runs with different num gpus? We could start with an accumulation step of 1, calculating the corresponding per device batch size; if that fails then we double the accumulation step (and so on).

anmarques: Hi Tuan. I fail to see how gradient_accumulation_steps can be set "consistently" from the total batch size alone. In my understanding, the accumulation batch size will be total_train_batch_size * gradient_accumulation_steps. On the other hand, per_device_train_batch_size = total_batch_size / nGPUs, regardless of the accumulation steps.

natuan: Hi Alex, my thinking was that we could interpret "total_train_batch_size" as your "accumulation batch size" above, which I think is consistent with the base trainer code here: https://github.com/neuralmagic/transformers/blob/master/src/transformers/trainer.py#L1085
in which the args.train_batch_size I think is computed from per_device_batch_size and num gpus as here: https://github.com/neuralmagic/transformers/blob/master/src/transformers/training_args.py#L699

So taking all together, we would have total_train_batch_size is a func of per_device_batch_size, gradient_accumulation_steps, and num gpus. We know num gpus, and then I'm thinking we can adjust the other two params for the batch to fit gpu memory.

anmarques: I think I understand that now. However, how would we check if it the batch fits into GPU memory? I would rather allow the user to specify either total_train_batch_size or per_device_train_batch_size (together with gradient_accumulation_steps) and compute the other from the expression in L1085 in trainer.py above. Does that address your concern?

anmarques: Please review the latest version. This one should have the total batch size take the accumulation steps into consideration

src/sparseml/transformers/nm_training_arguments.py
```diff
@@ -0,0 +1,102 @@
...
+                (
+                    "Cannot specify both total_eval_batch_size and"
+                    " per_device_eval_batch_size."
+                )
+            )
+
+        # If neither total nor per device batch sizes are defined
+        # set total batch sizes as default
+        if not self.total_train_batch_size and not self.per_device_train_batch_size:
+            self.total_train_batch_size = 8
```
dbogunowicz, line 83: I understand that `8` is the original, default value for both `total_train_batch_size` and `total_eval_batch_size`? I am concerned that we are somewhat "hiding" those default values from the user. But if user expectation is to obtain default value of `8` in HF TrainingArguments, I am probably cool with this.

src/sparseml/transformers/nm_training_arguments.py
```diff
@@ -0,0 +1,102 @@
...
+"""
+
+from dataclasses import dataclass, field
+from typing import Optional
+
+from transformers import TrainingArguments
+
+
+@dataclass
+class NMTrainingArguments(TrainingArguments):
```
dbogunowicz, line 27: Thank you for very clear walkthrough of the diff!

src/sparseml/transformers/nm_training_arguments.py
```diff
@@ -0,0 +1,102 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
```
dbogunowicz, line 1: Very general comment, is there a reason for forking the SparseML repository, creating branch locally in your fork, and then finally merging the PR into original repository's master?
I guess that for people who are inside NM (not external contributors), it would be advisable to work directly within our repo, since then other team members can e.g. quickly merge your branch or test your branch without cloning your fork. Or maybe there are some hidden benefits of your method which I am just failing to see? :)

anmarques: I was just following the guidelines for contributions, no strong reason. But noted, will work from our repo for following changes.

src/sparseml/transformers/nm_training_arguments.py
anmarques: I just used the default values from HF.

spacemanidol:
@alexm-nm given how many commits have happened since you set up this PR I would suggest either starting from scratch on new branch or resolving all the conflicts

anmarques:
This PR is ready but needs to be merged by hand. Will create a new PR soon and delete this one when done.

anmarques:
Too much time has passed. Will re-implement if needed

PR #564, opened 2022-02-21
Allow for export samples together with onnx

natuan:
This change enables exporting sample inputs and outputs for transformers (together with onnx). It could be run separately or as part of training/evaluation.

### DIFF
--- a/src/sparseml/transformers/export.py
+++ b/src/sparseml/transformers/export.py
@@ -60,9 +60,13 @@ import inspect
 import logging
 import math
 import os
-from typing import Any, Optional
+from typing import Any, Dict, Optional
 
+import numpy
+import onnxruntime
+from torch import Tensor
 from torch.nn import Module
+from torch.utils.data.dataloader import DataLoader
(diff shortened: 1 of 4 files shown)


markurtz (changes requested):
@natuan we want to remove export to onnx from the training flows and training scripts and keep this only in the onnx export script. This enables us to keep a separation of responsibilities and concerns that make testing and code changes in the future much easier. 

Additionally, is there any way we can add in the samples export to the export_onnx function and add it to the CLI arguments? Everything we push up now should have the ability to be run from the CLI along with the Python APIs

spacemanidol (approved):

natuan:
> @natuan we want to remove export to onnx from the training flows and training scripts and keep this only in the onnx export script. This enables us to keep a separation of responsibilities and concerns that make testing and code changes in the future much easier.
> 
> Additionally, is there any way we can add in the samples export to the export_onnx function and add it to the CLI arguments? Everything we push up now should have the ability to be run from the CLI along with the Python APIs

I'll close this PR and try to move sample generation to export_onnx

PR #573, opened 2022-02-25
Fix mfac gradcheck

KSGulin:
When testing M-FAC on very small networks with global sparsity, some batches could produce all-zero gradients. In turn, the modifier will think not all gradients were collected. This fix patches that issue by explicitly counting gradients collected.


bfineran (approved):
great catch

dhuangnm:
@KSGulin if this fix is needed for the release, please cherry-pick it over to the release/0.11 branch since we have cut the release, thanks!

markurtz (approved):

PR #575, opened 2022-02-28
Fix: remove transformers duplicate console logs

KSGulin:
Prevent Transformers logger from also printing SparseML logs, resulting in duplicate console logging


bfineran (approved):

dbogunowicz (approved):
Looking good, giving it green light.

PR #579, opened 2022-03-03
Proper handling of MFAC dict of numgrads

KSGulin:


rahul-tuli (approved):

dbogunowicz:
Could shortly describe in your own words what is the issue and how this fixes it?

KSGulin:
@dbogunowicz sure. num_grads can be defined in terms of an int or a dict that maps sparsity to num_grads. In the latter case, the check I'm changing fails because it tries to compare an int with a dict. The function I'm replacing it with takes a num_grads argument of int or dict and returns the correct num_grads as an int

bfineran (approved):

dbogunowicz (approved):

PR #580, opened 2022-03-03
Add metadata handling to the manager (recipe stages)

dbogunowicz:
This PR facilitates the flow of metadata through managers and recipes.
- The manager can now digest new metadata, as well as retrieve metadata from checkpoint files and do a comparison between the two.
- Enable composition of recipes in four configurations: staged w. normal, normal w. staged, normal w. normal and staged w. staged
- The manager now has the ability to dump the metadata into the recipe.

Sample usage:

Given `checkpoint_recipe`:
```
version: 1.1.0

stage_0_name:
  __metadata__:
    metadata: None
    level: 0

  stage_0_modifiers:
      - !EpochRangeModifier
          end_epoch: 3.0
          start_epoch: 0.0
  
      - !GMPruningModifier
          end_epoch: 3.0
          final_sparsity: 0.8
          global_sparsity: False
          init_sparsity: 0.2
          inter_func: cubic
          leave_enabled: True
          mask_type: unstructured
          params: ['re:.*weight']
          start_epoch: 0.0
          update_frequency: 1.0
  

stage_1:
  __metadata__:
    metadata: None
    level: 1

  stage_1_modifiers:
      - !ConstantPruningModifier
          end_epoch: 8
          params: ['re:.*weight']
          start_epoch: 5
          update_frequency: -1
  
      - !EpochRangeModifier
          end_epoch: 9.0
          start_epoch: 3.0
```

`recipe`:
```
version: 1.1.0

modifiers:
    - !EpochRangeModifier
        end_epoch: 3.0
        start_epoch: 0.0

    - !GMPruningModifier
        end_epoch: 3.0
        final_sparsity: 0.8
        global_sparsity: False
        init_sparsity: 0.2
        inter_func: cubic
        leave_enabled: True
        mask_type: unstructured
        params: ['re:.*weight']
        start_epoch: 0.0
        update_frequency: 1.0
 ```

`metadata`: `{metadata: None, level: 3}`

We pass both `metadata` and `recipe` to `ScheduledModifierManager` to create `recipe_manager`. Note: those two pieces of information would be stored in separate attributes; this would not create yaml recipe with metadata just yet.

Finally, we compose two managers/recipes to create a new manager, which combines recipes and metadata. Lastly, when saving the recipe, all the information is fused together and saved as `recipe.yaml`. 

```
recipe_manager = ScheduledModifierManager.from_yaml( file_path=recipe, metadata=metadata)
recipe_manager = ScheduledModifierManager.compose_staged(
              base_recipe=checkpoint_recipe,
              additional_recipe=recipe_manager,
          )

recipe_manager.save("recipe.yaml")
with open(recipe_path, "r") as file:
      final_recipe = file.read()
```
Final recipe becomes:

```
version: 1.1.0

stage_0:
  __metadata__:
    metadata: None
    level: 0

  stage_0_modifiers:
      - !EpochRangeModifier
          end_epoch: 3.0
          start_epoch: 0.0
  
      - !GMPruningModifier
          end_epoch: 3.0
          final_sparsity: 0.8
          global_sparsity: False
          init_sparsity: 0.2
          inter_func: cubic
          leave_enabled: True
          mask_type: unstructured
          params: ['re:.*weight']
          start_epoch: 0.0
          update_frequency: 1.0
  

stage_1:
  __metadata__:
    metadata: None
    level: 1

  stage_1_modifiers:
      - !ConstantPruningModifier
          end_epoch: 8
          params: ['re:.*weight']
          start_epoch: 5
          update_frequency: -1
  
      - !EpochRangeModifier
          end_epoch: 9.0
          start_epoch: 3.0
  

stage_3:
  __metadata__:
    metadata: None
    level: 3

  stage_3_modifiers:
      - !EpochRangeModifier
          end_epoch: 12.0
          start_epoch: 9.0
  
      - !GMPruningModifier
          end_epoch: 12.0
          final_sparsity: 0.8
          global_sparsity: False
          init_sparsity: 0.2
          inter_func: cubic
          leave_enabled: True
          mask_type: unstructured
          params: ['re:.*weight']
          start_epoch: 9.0
          update_frequency: 1.0
 ```


src/sparseml/optim/manager.py
```diff
@@ -283,15 +288,25 @@ def max_epochs(self) -> int:
 
         return max(vals) if len(vals) > 0 else -1
 
-    def save(self, file_path: str):
+    def save(self, file_path: str, include_metadata: bool):
```
bfineran, line 291: can default this to True

bfineran: especially important for not breaking any existing integrations

bfineran, line 291: need to also update `compose_staged ` as discussed offline

src/sparseml/pytorch/optim/manager.py
```diff
@@ -252,6 +252,7 @@ def from_yaml(
         file_path: Union[str, Recipe],
         add_modifiers: Optional[List[Modifier]] = None,
         recipe_variables: Optional[Union[Dict[str, Any], str]] = None,
+        metadata: Optional[Dict[str, Any]] = None,
```
bfineran, line 255: let's propagate this to `src/keras` and `src/tensorflow_v1` as well

bfineran (commented):
leaving a few initial comments, also as discussed, will need to add support for reading in previous metadata during `from_yaml` for reserialization as well as support for having metadata stored by stage

bfineran (changes requested):
changes look good at a high level. can you rebase onto latest main?

src/sparseml/optim/helpers.py
```diff
@@ -438,3 +441,48 @@ def _maybe_parse_number(val: str) -> Union[str, float, int]:
...
+    new, incoming metadata ('metadata').
+    If there exists a key in new metadata which also exists in previous metadata,
+    it will overwrite the old metadata. However, it is prohibitive to pass unseen keys.
+    :param metadata: New metadata
+    :param yaml_str: String representation of the recipe YAML file,
+        (maybe contains previous metadata)
+    :return: Validated metadata
+    """
+
+    previous_metadatas = dict(
```
dbogunowicz, line 458: Instead use the functionality I've coded before.

src/sparseml/optim/manager.py
```diff
@@ -283,15 +349,47 @@ def max_epochs(self) -> int:
...
+        :param checkpoint_manager: ...
+        :param include_metadata: boolean indicator whether metadata shall be
+            appended to the yaml file before saving.
         """
         file_path = clean_path(file_path)
         create_parent_dirs(file_path)
 
-        with open(file_path, "w") as yaml_file:
-            yaml_file.write(str(self))
+        if checkpoint_manager:
```
dbogunowicz, line 368: they compose themselves, using the fact that we have the class method.

src/sparseml/optim/helpers.py
```diff
@@ -438,3 +444,86 @@ def _maybe_parse_number(val: str) -> Union[str, float, int]:
...
+        return prev_metadata
+
+    elif not metadata and prev_metadata:
+        return prev_metadata
+
+    else:
+        # if we pass fresh metadata to a staged recipe with no previous metadata,
+        # we overwrite all the stages with the fresh metadata.
+        # Wondering whether this is correct behaviour, maybe raise ValueError?
+        if is_recipe_staged:
```
dbogunowicz, line 522: I would love to talk about the best behaviour here.

tests/sparseml/pytorch/optim/test_manager.py
```diff
@@ -41,6 +42,413 @@
...
+      start_epoch: 14
+      update_frequency: 2
+  - !EpochRangeModifier
+      end_epoch: 18.0
+      start_epoch: 12.0
+"""
+METADATA = """{{'metadata': None, 'level': {level}}}"""
+
+# TODO: Investigate a peculiar bug, in `optim/modifier.py Line 319
+# yaml.safe_load() improperly parses end_epoch for ACDC
```
dbogunowicz, line 361: Those tests fail due to solely one, weird behavior which only pertains to the 'end_epoch' attribute of ACDCModifier. I have found that the incorrect parsing in `optim/modifier` at L319 is the culprit. @bfineran let's take a look at it together,  remain clueless for now :)

src/sparseml/optim/manager.py
```diff
@@ -93,6 +103,7 @@ def compose_staged(
         cls,
         base_recipe: Union[str, "BaseManager"],
         additional_recipe: Union[str, "BaseManager"],
+        additional_metadata=Dict[str, Any],
```
dbogunowicz, line 106: Being pretty stubborn about pushing metadata merging dictionary here, let's discuss it.

src/sparseml/optim/manager.py
```diff
@@ -283,15 +375,39 @@ def max_epochs(self) -> int:
...
+
+        if include_metadata:
+            if check_if_staged_recipe(container):
+                for stage_name, metadata in self._metadata.items():
+                    container[stage_name]["metadata"] = metadata
+
+            else:
+                container["metadata"] = self._metadata["__metadata__"]
+
+            """This is a hack to make sure,
```
dbogunowicz, line 397: Can this be done more elegantly?

src/sparseml/optim/manager.py
```diff
@@ -118,19 +131,87 @@ def compose_staged(
...
+        ):
+            additional_stages = deepcopy(additional_recipe.modifiers)
+
+            # Using the first name of additional_recipe stages to
+            # generate the name of stage for base_recipe.
+            # If first_stage_name = [...]_[digit] ->
+            #   base_stage_name = [...]_[digit - 1]
+            # Else: base_stage_name = [first_stage_name]_0
+            #       (this will also happen if digit is negative)
+            # TODO: Maybe refactor it to account for negative numbers
```
dbogunowicz, line 196: I could fix it if critical.

src/sparseml/optim/helpers.py
```diff
@@ -247,7 +250,10 @@ def evaluate_recipe_yaml_str_equations(recipe_yaml_str: str) -> str:
                 val, variables, non_val_variables
             )
 
-    return rewrite_recipe_yaml_string_with_classes(container)
+    # I find it really helpful to use the container returned by this function.
+    # I'd love to either make this return only the container
+    # or the (func(container), container)
+    return rewrite_recipe_yaml_string_with_classes(container), container
```
dbogunowicz, line 256: Better `return container` and pushing out `rewrite_recipe_yaml_string_with_classes`?

bfineran: why would we prefer that here? this is a top level function that takes in a valid SparseML recipe, so it should also return one. the container here will not have the proper `Modifier` classes loaded

src/sparseml/optim/helpers.py
```diff
@@ -247,7 +250,10 @@ def evaluate_recipe_yaml_str_equations(recipe_yaml_str: str) -> str:
                 val, variables, non_val_variables
             )
 
-    return rewrite_recipe_yaml_string_with_classes(container)
+    # I find it really helpful to use the container returned by this function.
+    # I'd love to either make this return only the container
+    # or the (func(container), container)
```
bfineran, line 255: nit: remove

src/sparseml/optim/helpers.py
bfineran: also if the only purpose of returning container here is so we can get a loaded dict of the yaml string (and don't actually care about the variables, just use `load_recipe_yaml_str_no_classes`

src/sparseml/optim/helpers.py
```diff
@@ -438,3 +444,86 @@ def _maybe_parse_number(val: str) -> Union[str, float, int]:
             return float(val)
         except Exception:
             return val
+
+
+def validate_metadata(metadata: dict, yaml_str: str) -> dict:
```
bfineran, line 449: let's redo the logic here with the following rule:

1. if no metadata if found in the `yaml_str` then all is good
2. if metadata if found in the `yaml_str` and is equal to the given `metadata` dict, then all is good
3. if the given `metadata` and `yaml_str` metadata conflict we raise a warning and possible exception

in other words, we should never merge, because the metadata for a previous run has no bearing on the current run and if the metadata differs then we want the user to know that they might not be able to reproduce that same exact result (of course, that may be the goal :) )

src/sparseml/optim/helpers.py
```diff
@@ -438,3 +444,86 @@ def _maybe_parse_number(val: str) -> Union[str, float, int]:
...
+    it will overwrite the old metadata. However, it is prohibitive to pass novel keys.
+
+    :param metadata: New metadata
+    :param yaml_str: String representation of the recipe YAML file,
+        (may contains previous metadata)
+    :return: Validated metadata
+    """
+
+    default_metadata_key = "__metadata__"
+    _, container = evaluate_recipe_yaml_str_equations(yaml_str)
```
bfineran, line 473: to get the container, using `load_recipe_yaml_str_no_classes` should be sufficient, no need to fuss with eval

bfineran (commented):
one more thing, in all of the `from_yaml` functions, if no metadata is provided, then we should load in any metadata from the yaml string. (this is especially important for checkpoint recipes). right now it looks like this is done by the merge, but as requested, we should never merge the given and loaded metadata dicts

src/sparseml/optim/manager.py
bfineran, line 106: as discussed in call, let's get rid of additional metadata. this should be completely controlled by the given recipe/manager objects.

every stage on either side of the merge should keep its original metadata in the final composed recipe

src/sparseml/optim/manager.py
```diff
@@ -118,19 +131,87 @@ def compose_staged(
         if not isinstance(additional_recipe, BaseManager):
             additional_recipe = cls.from_yaml(additional_recipe)
 
-        if isinstance(base_recipe.modifiers, OrderedDict):
-            raise ValueError(
-                "non-staged recipes not yet supported for Manager.compose_staged "
-                "found base_recipe with non_staged modifiers"
+        # Both base_recipe and additional_recipe are non-staged_recipes
```
bfineran, line 134: as discussed, let's default the `base_recipe` stage name to `pre_{first_additional_stage_name}` and `additional_recipe` stage name to `stage_{num_base_stages + 1}`

bfineran (changes requested):

src/sparseml/optim/helpers.py
```diff
@@ -493,3 +494,83 @@ def _maybe_parse_number(val: str) -> Union[str, float, int]:
...
+
+    default_metadata_key = "__metadata__"
+    container = load_recipe_yaml_str_no_classes(yaml_str)
+    is_recipe_staged = check_if_staged_recipe(container)
+
+    if is_recipe_staged:
+        stage_names = [
+            stage_name
+            for stage_name, stage_dict in container.items()
+            if isinstance(stage_dict, dict)
```
bfineran, line 530: this technically isn't sufficient for if an item is a stage or not, we also need to check if stage_dict has a key with "modifiers" in its name

src/sparseml/optim/helpers.py
```diff
@@ -493,3 +494,83 @@ def _maybe_parse_number(val: str) -> Union[str, float, int]:
...
+        ]
+        prev_metadata = {
+            stage_name: container[stage_name]["metadata"]
+            for stage_name in stage_names
+            if "metadata" in container[stage_name]
+        }
+
+    else:
+        if "metadata" in container.keys():
```
bfineran, line 539: the logic here for consecutive read/writes doesn't make sense, we are adding `"__metadata__"` as the new metadata but searching specifically for an existing key of `"metadata"`.  I propose we enforce that a key is only official metadata if it's name is `"__metadata__"`.  And we keep the value of `"__metadata__"` stored as `RECIPE_METADATA_KEY = "__metadata"` in https://github.com/neuralmagic/sparseml/blob/main/src/sparseml/utils/helpers.py (see area where ALL_TOKEN and others are defined).

This will 1) simplify our logic here and 2) enforce that we consistently can find metadata in the same place

src/sparseml/optim/helpers.py
```diff
@@ -493,3 +494,83 @@ def _maybe_parse_number(val: str) -> Union[str, float, int]:
...
+            prev_metadata = {default_metadata_key: container["metadata"]}
+        else:
+            prev_metadata = {}
+
+    # checking if new metadata does not conflict with
+    # the previous, possibly overwriting previous_data
+    if metadata and prev_metadata:
+        for k, v in metadata.items():
```
bfineran, line 547: nit: ambiguous variable names

src/sparseml/optim/helpers.py
```diff
@@ -493,3 +494,83 @@ def _maybe_parse_number(val: str) -> Union[str, float, int]:
...
+
+    elif not metadata and prev_metadata:
+        return prev_metadata
+
+    else:
+        if is_recipe_staged:
+            return {
+                stage_name: metadata
+                for stage_name, stage_dict in container.items()
+                if isinstance(stage_dict, dict)
```
bfineran, line 573: need the same stronger check for stages here

src/sparseml/optim/helpers.py
```diff
@@ -493,3 +494,83 @@ def _maybe_parse_number(val: str) -> Union[str, float, int]:
...
+            prev_metadata = {default_metadata_key: container["metadata"]}
+        else:
+            prev_metadata = {}
+
+    # checking if new metadata does not conflict with
+    # the previous, possibly overwriting previous_data
+    if metadata and prev_metadata:
+        for k, v in metadata.items():
+            key_found = False
+            for stage_name, metadata_dict in prev_metadata.items():
```
bfineran, line 549: previous metadata is not necessarily staged. also due to number of nested statements, I'd reccomend splitting this section out

src/sparseml/optim/helpers.py
```diff
@@ -493,3 +494,83 @@ def _maybe_parse_number(val: str) -> Union[str, float, int]:
...
+                if k in metadata_dict.keys():
+                    if is_recipe_staged:
+                        container[stage_name]["metadata"][k] = v
+                    else:
+                        container["metadata"][k] = v
```
bfineran, line 554: not following the logic here, why are we overriding the container with old metadata and then returning the previous metadata?

src/sparseml/optim/helpers.py
```diff
@@ -493,3 +494,83 @@ def _maybe_parse_number(val: str) -> Union[str, float, int]:
...
+
+                    key_found = True
+            if not key_found:
+                raise ValueError(
+                    f"The previously unseen metadata key {k} "
+                    f"has not been recognized in previous "
+                    f"metadata."
```
bfineran, line 561: nit: these lines don't need to be f-strings

src/sparseml/optim/manager.py
```diff
@@ -118,19 +123,76 @@ def compose_staged(
...
+        if isinstance(base_recipe.modifiers, List) and isinstance(
+            additional_recipe.modifiers, List
+        ):
+            # Need to generate stage names for two standard recipes
+            base_stage_name, additional_stage_name = [f"stage_{i}" for i in range(2)]
```
bfineran, line 131: let's hardcode here, a little easier to read :)

src/sparseml/optim/manager.py
```diff
@@ -118,19 +123,76 @@ def compose_staged(
...
+            base_recipe_metadata[base_stage_name] = base_recipe_metadata.pop(
+                "__metadata__"
             )
-        if isinstance(additional_recipe.modifiers, OrderedDict):
-            raise ValueError(
-                "non-staged recipes not yet supported for Manager.compose_staged "
-                "found additional_recipe with non_staged modifiers"
+            additional_recipe_metadata[
+                additional_stage_name
+            ] = additional_recipe_metadata.pop("__metadata__")
```
bfineran, line 146: why do we need to pop the "__metadata__" key? why does an unstaged recipe have this key?

src/sparseml/optim/manager.py
```diff
@@ -118,19 +123,76 @@ def compose_staged(
...
+
+            base_stages = deepcopy(base_recipe.modifiers)
+
+            additional_stage_name = f"stage_{len(base_stages) + 1}"
+            if additional_stage_name in base_stages.keys():
+                raise ValueError(
+                    f"Generated new stage name: {additional_stage_name}, "
+                    f"but there already exists"
+                    f"a stage with that name in the checkpoint file. "
+                    f"Please edit the stage name in the checkpoint file."
```
bfineran, line 161: nit: these lines don't need to be f-strings

src/sparseml/optim/manager.py
```diff
@@ -118,19 +123,76 @@ def compose_staged(
...
+
+            additional_stages = {
+                additional_stage_name: deepcopy(additional_recipe.modifiers)
+            }
+
+            base_recipe_metadata = deepcopy(base_recipe._metadata)
+            additional_recipe_metadata = deepcopy(additional_recipe._metadata)
+            additional_recipe_metadata[
+                additional_stage_name
+            ] = additional_recipe_metadata.pop("__metadata__")
```
bfineran, line 172: can any of this logic be shared in the above/below blocks?

src/sparseml/optim/manager.py
```diff
@@ -333,33 +401,71 @@ def iter_modifiers(self) -> Generator[None, None, BaseModifier]:
             for mod in modifiers_list:
                 yield mod
 
-    def to_string_lines(self) -> List[str]:
+    def to_string_lines(self, include_metadata: bool = False) -> List[str]:
```
bfineran, line 404: why default to false here and true above?

src/sparseml/optim/manager.py
```diff
@@ -333,33 +401,71 @@ def iter_modifiers(self) -> Generator[None, None, BaseModifier]:
...
         """
+        :param include_metadata: boolean indicator whether metadata shall be
+            appended to the yaml file before saving..
```
bfineran, line 407: nit: extra period

src/sparseml/optim/manager.py
```diff
@@ -333,33 +401,71 @@ def iter_modifiers(self) -> Generator[None, None, BaseModifier]:
             for mod in modifiers_list:
                 yield mod
 
-    def to_string_lines(self) -> List[str]:
+    def to_string_lines(self, include_metadata: bool = False) -> List[str]:
         """
+        :param include_metadata: boolean indicator whether metadata shall be
```
bfineran, line 406: add default value to docstring

src/sparseml/optim/manager.py
```diff
@@ -283,15 +349,17 @@ def max_epochs(self) -> int:
 
         return max(vals) if len(vals) > 0 else -1
 
-    def save(self, file_path: str):
+    def save(self, file_path: str, include_metadata: bool = True):
         """
         :param file_path: the file path to save the yaml config representation to
+        :param include_metadata: boolean indicator whether metadata shall be
```
bfineran, line 355: add default value to docstring

src/sparseml/optim/manager.py
```diff
@@ -333,33 +401,71 @@ def iter_modifiers(self) -> Generator[None, None, BaseModifier]:
...
+        Parse `self._metadata` into list of strings.
+        :param stage: Name of the current recipe stage.
+            If stage = None, we are dealing with standard, unstaged recipe.
+        :return: a list of lines for a string / yaml representation of the
+            metadata for the given stage in the manager
+        """
+        yaml_str_lines = []
+
+        if stage:
+            yaml_str_lines.append("  metadata:")
```
bfineran, line 436: let's make sure to be propagate changes of the metadata key here

bfineran (changes requested):
see comments, biggest concern is the potential for inconsistent metadata keys

src/sparseml/optim/manager.py
dbogunowicz: So to clarify:

The current state would save the recipe as :

```stage_4:
  metadata:
    var: None
    level: 1

  stage_4_modifiers:
      - !ConstantPruningModifier
          end_epoch: 17
          ...
```

what you have in mind is that metadata key always includes the stage name e.g.
```
stage_4:
  stage_4_metadata:
    var: None
    level: 1

  stage_4_modifiers:
      - !ConstantPruningModifier
          end_epoch: 17
          ...
```

Right?

src/sparseml/optim/manager.py
dbogunowicz: The reason is --> if `def to_string_lines(self, include_metadata: bool = True)` this test would fail:
`tests/sparseml/pytorch/optim/test_manager.py::test_manager_staged_recipe_serialization`,
since it contains the following line of code:
`manager_yaml_str = str(manager)`.

What happens is: `str` calls dunder `__str__()` which in turn calls `to_string_lines` so that the recipe yaml gets assembled. Since this `manager` does not have any metadata AND we call `to_string_lines()` directly from `__str__()` and not from `save()`, this is how we prevent the test from failing.

So we either:
1. keep `include_metadata` as False to keep to old functionality
2. slightly change `tests/sparseml/pytorch/optim/test_manager.py::test_manager_staged_recipe_serialization` test.

I guess that the user is not expected to use `__str__()` but would rather assemble yaml through `save` function. This is why I have decided to make `include_metadata` default to False. 

Does it make sense? I am open for suggestions.

src/sparseml/optim/helpers.py
dbogunowicz: Good catch, corrected.

src/sparseml/optim/helpers.py
dbogunowicz: This piece of code has been refactored to be more readable.

src/sparseml/optim/helpers.py
dbogunowicz: This piece of code has been refactored to be more readable.

src/sparseml/optim/helpers.py
dbogunowicz: This piece of code has been refactored to be more readable.

src/sparseml/optim/manager.py
dbogunowicz: So the way how I think about `manager._metadata` is:
if `manager`'s recipe is staged:
manager._metadata = {stage_name1: <metadata1>, stage_name2: <metadata2>, ...}
if `manager`'s recipe is not staged:
manager._metadata = {__metadata__: <metadata>}

this construction ```dict[new_key] = dict.pop(old_key)``` is one liner for replacing `old_key` with `new_key`.

So when working with e.g staged checkpoint recipe and new standard recipe:
1. `checkpoint_manager._metadata =  {stage_name1: <metadata1>, stage_name2: <metadata2>, ...}`
2.  compute `stage_name3`
2. `new_manager._metadata =  {stage_name3: <metadata3>}` 
3.  Combine -> combined_manager._metadata =  {stage_name1: <metadata1>, stage_name2: <metadata2> stage_name3: <metadata3>}`

Does this make sense?

src/sparseml/optim/helpers.py
```diff
@@ -512,3 +517,117 @@ def _maybe_parse_number(val: str) -> Union[str, float, int]:
...
+                    raise ValueError(
+                        "Attempting to find and overwrite overwrite checkpoint metadata"
+                        f"with metadata key {var_name}, but {var_name}"
+                        "is not present in checkpoint metadata."
+                    )
+
+        return (
+            checkpoint_metadata
+            if is_container_staged
+            else {RECIPE_METADATA_KEY: checkpoint_metadata}
```
bfineran, line 614: let's simplify the logic here:

1. checkpoint metadata + new metadata
       if they are equal return either
       otherwise raise warning with any differences logged
2. return the one that exists (expanding as needed for stages)


```pyton

def _check_warn_dict_difference(stage_name: optional, old_metadata, new_metadata):
    pass


if checkpoint_metadata and metadata:
    if staged:
          for stage_name in stage_names:
                 _check_warn_dict_difference(stage_name, container[stage_name][metadata_key], metadata)
   else:
         _check_warn_dict_difference(None, container[metadata_key], metadata)
  if metadata:
      if staged:
           return {stage_name: metadata for stage_name in stage_names}
      else:
           return metadata
elif checkpoint_metadata:
   return checkpoint_metadata
else:
 pass

```

bfineran (changes requested):

src/sparseml/optim/helpers.py
```diff
@@ -512,3 +518,118 @@ def _maybe_parse_number(val: str) -> Union[str, float, int]:
...
+    stage_names = _get_recipe_stage_names(container)
+    for stage_name in stage_names:
+        if RECIPE_METADATA_KEY in container[stage_name].keys():
+            metadata[stage_name] = container[stage_name][RECIPE_METADATA_KEY]
+    if metadata and (len(metadata) != len(stage_names)):
+        raise ValueError(
+            "It seems that some stages in your checkpoint recipe"
+            "contain metadata and some don't."
+            "This is not allowed: either all the "
+            f"stages contain {RECIPE_METADATA_KEY} key or none does."
```
bfineran, line 541:
```suggestion
            "Some stages in recipe do not. Either all or no stages must "
            f"must contain the {RECIPE_METADATA_KEY} key"
```

editing for brevity and spacing

src/sparseml/optim/helpers.py
```diff
@@ -512,3 +518,118 @@ def _maybe_parse_number(val: str) -> Union[str, float, int]:
...
+    return stage_names
+
+
+def _check_warn_dict_difference(original_dict, new_dict):
+    if original_dict != new_dict:
+        warnings.warn(
+            f"Attempting to overwrite old metadata: {original_dict} "
+            f"with new metadata: {new_dict}. "
+            "It is prohibitive to alter the metadata of the recipe. "
+            "The new metadata will be omitted and discarded."
```
bfineran, line 567:
```suggestion
            "This may lead to different results than the original run of the recipe. "
            "The previous metadata will be omitted and discarded. Ignore if a "
            "change in metadata is expected"
```
The old metadata... correct?

src/sparseml/optim/helpers.py
```diff
@@ -512,3 +518,118 @@ def _maybe_parse_number(val: str) -> Union[str, float, int]:
...
+        if is_container_staged
+        else _extract_metadata_from_recipe(container)
+    )
+
+    if checkpoint_metadata:
+        if metadata:
+            if is_container_staged:
+                for stage_name in _get_recipe_stage_names(container):
+                    # check if metadata is staged...
+                    if _get_recipe_stage_names(container) == list(metadata.keys()):
```
bfineran, line 600: can we move this check outside the for loop to see if the metadata is staged? Also would like to remove the ordering check (ie list equality vs set equality here) as it may not be enforced with plain dicts

src/sparseml/optim/helpers.py
```diff
@@ -512,3 +518,118 @@ def _maybe_parse_number(val: str) -> Union[str, float, int]:
...
+                        _check_warn_dict_difference(
+                            container[stage_name][RECIPE_METADATA_KEY],
+                            metadata[stage_name],
+                        )
+                    # ...or not
```
bfineran, line 605: nit: unnecessary

src/sparseml/optim/helpers.py
```diff
@@ -512,3 +518,118 @@ def _maybe_parse_number(val: str) -> Union[str, float, int]:
...
+        _extract_metadata_from_staged_recipe(container)
+        if is_container_staged
+        else _extract_metadata_from_recipe(container)
+    )
+
+    if checkpoint_metadata:
+        if metadata:
+            if is_container_staged:
+                for stage_name in _get_recipe_stage_names(container):
+                    # check if metadata is staged...
```
bfineran, line 599: nit: remove ...

src/sparseml/optim/manager.py
```diff
@@ -118,19 +123,70 @@ def compose_staged(
         if not isinstance(additional_recipe, BaseManager):
             additional_recipe = cls.from_yaml(additional_recipe)
 
-        if isinstance(base_recipe.modifiers, OrderedDict):
-            raise ValueError(
-                "non-staged recipes not yet supported for Manager.compose_staged "
-                "found base_recipe with non_staged modifiers"
+        base_recipe_metadata = deepcopy(base_recipe._metadata)
+        additional_recipe_metadata = deepcopy(additional_recipe._metadata)
```
bfineran, line 127: let's add an `@property` for metadata so it can be accessed by `manager.metadata` instead of as a private field. may come up elsewhere

src/sparseml/optim/manager.py
```diff
@@ -333,33 +395,77 @@ def iter_modifiers(self) -> Generator[None, None, BaseModifier]:
...
+        if stage:
+            yaml_str_lines.append(f"  {RECIPE_METADATA_KEY}:")
+            if not isinstance(self._metadata[stage], dict):
+                yaml_str_lines[-1] += f" {self._metadata[stage]}"
+            else:
+                for key, value in self._metadata[stage].items():
+                    yaml_str_lines.append(f"    {key}: {value}")
+        else:
+            yaml_str_lines.append(f"{RECIPE_METADATA_KEY}:")
+            if not isinstance(self._metadata[stage], dict):
```
bfineran, line 438: isn't `stage` False here? how should this work?

src/sparseml/optim/manager.py
```diff
@@ -333,33 +395,77 @@ def iter_modifiers(self) -> Generator[None, None, BaseModifier]:
...
+        """
+        Parse `self._metadata` into list of strings.
+        :param stage: Name of the current recipe stage.
+            If stage = None, we are dealing with standard, unstaged recipe.
+        :return: a list of lines for a string / yaml representation of the
+            metadata for the given stage in the manager
+        """
+        yaml_str_lines = []
 
+        if stage:
```
bfineran, line 429: if indentation is the main difference between these cases, see if we can combine the logic and just have variables for indentation and such

bfineran (changes requested):
logic changes looking good.  left some nits. additioanlly for the tests, why do we need `eval()` around the strings? And instead of using format, let's change that to a function instead (like fake_metadata() or something)

bfineran:
also @dbogunowicz looks like the pytorch tests are failing, please take a look

src/sparseml/optim/manager.py
dbogunowicz: This does not work obviously and should have been addressed before. Fixed now.

src/sparseml/optim/manager.py
dbogunowicz: Indentation is not the only difference, let's keep it as it is.

src/sparseml/optim/helpers.py
dbogunowicz: Good point. I would however place the `is_metadata_staged` check once we know that `checkpoint_metadata` and `metadata` and `is_container_staged`. Those are the logical required conditions before the `is_metadata_staged` check.

src/sparseml/optim/helpers.py
dbogunowicz: Wait. This is something which we are not in sync still.

You suggest that if we have `previous_metadata` and pass `metadata` and they are not equal, we should raise a warning, yet still ALLOW to completely overwrite the `previous_metadata` with the new `metadata` right? This means wiping `previous_metadata` completely and replacing it with `metadata`. i.e. we trust the user that they know what they are doing if they decide to pass `metadata` along with a recipe that already contains metadata. 
E.g.
`previous metadata = {x:2, y:3, z:4}` and `metadata = {u:5}` --> result will be `{u:5}` right?

dbogunowicz:
@bfineran regarding the way how I generate metadata for tests I would like to defend my implementation. I think it its shorter and simpler and more pythonic in a way, then writing a helper function `fake_metadata()`. Let's take a look:

`METADATA = """{{"foo:  2 ,"{key}":{value}}}"""` allows us to quickly switch placeholders for parameters using `.format()`
`m1 = METADATA .format(key="bar", value=3) -> '{"foo:  2 ,"bar": 3}`

Now `eval()` simply turns `m1` string into dict.

src/sparseml/optim/helpers.py
dbogunowicz: At the end of the day, it is really about how much we trust the user. I will change the functionality of my code to behave as described above.

bfineran: yes, this is what we determined. Once we start a new training run with a recipe, the previous metadata from that recipe is now irrelevant.  In your example, if x,y, and z no longer exist or have had their values changed in the new run, why should they be tracked by the new run just because they were set previously?  More concretely, say a checkpoint run was done at batch size 16 and we launch a new run that for some reason does not track batch size in metadata - it would be wrong to assume that the new run is using batch size 16 and thus should not be considered in the new metadata.

Please make sure that the logic is consistent with this - the previous metadata should always be overwritten if new metadata is given.

bfineran:
> @bfineran regarding the way how I generate metadata for tests I would like to defend my implementation. I think it its shorter and simpler and more pythonic in a way, then writing a helper function `fake_metadata()`. Let's take a look:
> 
> `METADATA = """{{"foo: 2 ,"{key}":{value}}}"""` allows us to quickly switch placeholders for parameters using `.format()` `m1 = METADATA .format(key="bar", value=3) -> '{"foo: 2 ,"bar": 3}`
> 
> Now `eval()` simply turns `m1` string into dict.

Thanks for the explanation, definitely see why the decision was made. Going to push back further on this however for two reasons:

1. while the use of `eval()` may be contained here, it is generally an unsafe function and not considered good python practice. We have discouraged the use of eval in our repo (and that's why we implemented our own "restricted" eval for recipe variables).
2. I'm not sold that repeatedly formatting a string and then casting to a dictionary (assuming the variable given for the `format` function produces a valid dict string and doesn't cause an eval error) is any more pythonic than creating a simple function that returns a dict literal based on its inputs.

src/sparseml/optim/helpers.py
dbogunowicz: Got it. This is the current behavior now, I confirm. Will double check tho (or quickly write some more tests if necessary)

bfineran (approved):
looks good, thanks for the hard work on this.

small ask that before landing we add in the description for this PR an example of a simple recipe, loaded into ScheduledModifierManager with metadata in `from_yaml` then what is written to (str) of that manager

KSGulin (approved):
LGTM!

PR #582, opened 2022-03-03
Load `arch_key` from checkpoint

rahul-tuli:
The goal of this PR is to allow reading in `arch_key` from a checkpoint if it is not passed in as an argument, (if passed in then passed value takes precedence)
Before this PR the `arch_key` was a required argument while invoking our training flows, for example:

```bash
    --recipe-path quantization-recipe.yaml \ 
    --sparse-transfer-learn \
    --arch-key resnet50 \
    --dataset imagenette \
    --checkpoint_path ./resnet-checkpoint.pth \
    --train-batch-size 32 \
    --test-batch-size 256 \
    --dataset-path dataset
```

After this PR if the checkpoint contains `arch_key` then the argument can be omitted, it is automatically inferred from the checkpoint, if the checkpoint also does not contain `arch_key` then appropriate assertion is raised

example checkpoint: 
```python
{
 
 'model_state_dict': model.state_dict(),
 'optimizer_state_dict': optimizer.state_dict(),
 'arch_key': 'resnet50',
  ...

}
```

```bash
    --recipe-path quantization-recipe.yaml \ 
    --sparse-transfer-learn \
    --dataset imagenette \
    --checkpoint_path ./resnet-checkpoint.pth \
    --train-batch-size 32 \
    --test-batch-size 256 \
    --dataset-path dataset
```


src/sparseml/pytorch/image_classification/export.py
```diff
@@ -260,6 +261,19 @@ class ExportArgs:
     )
 
     def __post_init__(self):
+        if self.arch_key is None:
+            assert self.checkpoint_path, (
```
bfineran, line 265: let's do raise a ValueError here instead

src/sparseml/pytorch/image_classification/export.py
```diff
@@ -260,6 +261,19 @@ class ExportArgs:
...
+                "Must provide a checkpoint path if " "no arch_key is provided"
+            )
+
+            checkpoint = torch.load(self.checkpoint_path)
```
bfineran, line 269: let's avoid doing the load and assert here since we only want to load from torch once (it can take a while to read in the data for larger models).

bfineran (commented):
logic looks good overall, but we should definitely push the load to later on in the logic. this should be made a lot easier with the Trainer refactor

rahul-tuli:
> logic looks good overall, but we should definitely push the load to later on in the logic. this should be made a lot easier with the Trainer refactor

I would like to leave this as is until the Trainer refactor lands, and take this up in a separate PR. Doing a single load would require refactoring code in [src/sparseml/pytorch/models/registry](https://github.com/neuralmagic/sparseml/blob/442dec37f02480b7bb1dd6b712fddc6f8ad3a583/src/sparseml/pytorch/models/registry.py#L346)

src/sparseml/pytorch/image_classification/export.py
rahul-tuli: refer to [this comment](https://github.com/neuralmagic/sparseml/pull/582#issuecomment-1059006841)

src/sparseml/pytorch/image_classification/export.py
rahul-tuli: Done!

bfineran:
> > logic looks good overall, but we should definitely push the load to later on in the logic. this should be made a lot easier with the Trainer refactor
> 
> I would like to leave this as is until the Trainer refactor lands, and take this up in a separate PR. Doing a single load would require refactoring code in [src/sparseml/pytorch/models/registry](https://github.com/neuralmagic/sparseml/blob/442dec37f02480b7bb1dd6b712fddc6f8ad3a583/src/sparseml/pytorch/models/registry.py#L346)

sounds good, let's wait on the trainer to be finalized then before adding this intermediate support since it will change how models are saved/loaded

bfineran (approved):

dbogunowicz (approved):

PR #587, opened 2022-03-04
Pytorch Modifiers Move

rahul-tuli:
The goal of this PR is to move remaining pytorch modifiers from `src/sparseml/pytorch/optim` to the new `src/sparseml/pytorch/sparsification` package for making it one stop shop for anything related to modifiers,

Note this is a breaking change,

Example,

Before this PR
```python
from sparseml.pytorch.optim import DistillationModifier
```

After this PR
```python
from sparseml.pytorch.sparsification import DistillationModifier
```


src/sparseml/pytorch/optim/manager.py
```diff
@@ -26,7 +26,7 @@
 from torch.optim.optimizer import Optimizer
 
 from sparseml.optim import BaseManager, load_recipe_yaml_str, parse_recipe_variables
-from sparseml.pytorch.optim.modifier import Modifier, ScheduledModifier
+from sparseml.pytorch.sparsification.modifier import Modifier, ScheduledModifier
```
bfineran, line 29: can you check that this doesn't add any circular dependencies between optim and sparsification packages?

dbogunowicz:
Looks good to me. If there are no import problems and all the tests are green after rebasing on top of the master then it seems solid. Great work Rahul, it feels good to have modifiers in one place.

dbogunowicz (approved):

bfineran (approved):
approving, but let's wait until release 0.11 is cut to land

src/sparseml/pytorch/base.py
```diff
@@ -49,7 +49,7 @@
 
 
 _TORCH_MIN_VERSION = "1.0.0"
-_TORCH_MAX_VERSION = "1.9.100"  # set bug to 100 to support all future 1.9.X  versions
+_TORCH_MAX_VERSION = "1.10.100"  # set bug to 100 to support all future 1.10.X  versions
```
dbogunowicz, line 52: We are sticking to pytorch`1.9.X`

dbogunowicz (commented):
@rahul-tuli this PR is a bit rusty, could you update the git history, then I would take another look/ do some testing.

rahul-tuli:
Closing this PR in favor of newly rebased PR #689

PR #588, opened 2022-03-04
Add support for initializing empty BaseModifier

KSGulin:
This is needed to support the removal of `log_types` in an upcoming PR. `log_types` is currently the only init arg for `BaseModifier`. With its removal, YAML serialization for `BaseModifier` fails


bfineran (approved):
good digging here

dbogunowicz (approved):
LGTM

PR #590, opened 2022-03-07
Make staged recipes be properly overwritten by `recipe_variables` argument.

dbogunowicz:
When a manager calls a `from_yaml()` function to construct itself...
image
...it can take `recipe_variables`  argument. It allows overwriting chosen variables in the recipe. While this function was working correctly with standard recipes, now it also parses staged recipes. The overwriting happens in a greedy way i.e. the code looks at every possible variable in the raw recipe (at every level). If it finds a key that corresponds to the entry it`recipe_variables`: do overwrite.

### DIFF
--- a/src/sparseml/optim/helpers.py
+++ b/src/sparseml/optim/helpers.py
@@ -198,7 +198,9 @@ def parse_recipe_variables(
 def update_recipe_variables(recipe_yaml_str: str, variables: Dict[str, Any]) -> str:
     """
     :param recipe_yaml_str: YAML string of a SparseML recipe
-    :param variables: variables dictionary to update recipe top level variables with
+    :param variables: variables dictionary to update recipe top level variables with.
+                      If recipe contains stages, it will parse the whole recipe
+                      and substitute ANY variable with the corresponding name.
     :return: given recipe with variables updated
     """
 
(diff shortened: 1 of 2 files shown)


src/sparseml/optim/helpers.py
```diff
@@ -198,7 +198,9 @@ def parse_recipe_variables(
 def update_recipe_variables(recipe_yaml_str: str, variables: Dict[str, Any]) -> str:
     """
     :param recipe_yaml_str: YAML string of a SparseML recipe
-    :param variables: variables dictionary to update recipe top level variables with
+    :param variables: variables dictionary to update recipe top level variables with.
+                      If recipe contains stages, it will parse the whole recipe
```
rahul-tuli, line 202: Comments should be tab spaced on new lines

dbogunowicz: I should learn this once for all haha, thanks

KSGulin (commented):
Left one comment, otherwise looks good to me!

src/sparseml/optim/helpers.py
```diff
@@ -207,15 +209,52 @@ def update_recipe_variables(recipe_yaml_str: str, variables: Dict[str, Any]) ->
         # yaml string does not create a dict, return original string
         return recipe_yaml_str
 
-    for key in variables:
-        if key not in container:
+    if check_if_staged_recipe(container):
+        key_found = False
```
bfineran, line 213: should this be at the scope of the loop below so we can raise an error on any missing key?

src/sparseml/optim/helpers.py
```diff
@@ -207,15 +209,59 @@ def update_recipe_variables(recipe_yaml_str: str, variables: Dict[str, Any]) ->
...
+                        if var_key in modifier.keys():
+                            container[key][i][var_key] = variables[var_key]
+                            key_found = True
+
+                else:
+                    if var_key == key:
+                        container[var_key] = variables[var_key]
+                        key_found = True
+
+    if not key_found:
```
bfineran, line 258: this needs to be moved inside the loop, if any key is missing we want to raise (also notice that `var_key` is a loop variable, so we would only print the last variable regardless if that one was found or not

src/sparseml/optim/helpers.py
```diff
@@ -207,15 +209,59 @@ def update_recipe_variables(recipe_yaml_str: str, variables: Dict[str, Any]) ->
...
         return recipe_yaml_str
 
-    for key in variables:
-        if key not in container:
-            raise ValueError(
-                f"updating recipe variable {key} but {key} is not currently "
-                "set in existing recipe. Set the variable in the recipe in order "
-                "to overwrite it."
-            )
+    key_found = False
```
bfineran, line 212: should be moved inside the loop, we want to make sure every key is found

bfineran (changes requested):

src/sparseml/optim/helpers.py
dbogunowicz: Excellent catch, thank you Ben!

bfineran (approved):

KSGulin (approved):
Good stuff

PR #591, opened 2022-03-07
Remove: `export` to `onnx` by default

rahul-tuli:
The goal of this PR is to remove `onnx export` by default for `sparseml.image_classification.train` entry-point, 

use `sparseml.image_classification.export` for onnx exports

### DIFF
--- a/src/sparseml/pytorch/image_classification/train.py
+++ b/src/sparseml/pytorch/image_classification/train.py
@@ -615,7 +615,6 @@ def train(
                     helpers.save_model_training(
                         model,
                         optim,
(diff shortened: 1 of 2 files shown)


bfineran (changes requested):
let's make sure to update the kwargs of this helper function to remove no longer needed items

rahul-tuli:
> let's make sure to update the kwargs of this helper function to remove no longer needed items

Addressed! Good catch!

bfineran (approved):

xkazuka (approved):

PR #592, opened 2022-03-07
fixing reduction in teacher distillation function

spacemanidol:


spacemanidol:
Bug fix for proper scaling of teacher loss

markurtz (approved):

bfineran (dismissed):

src/sparseml/pytorch/optim/modifier_distillation.py
```diff
@@ -355,10 +353,10 @@ def _calc_distill_loss(self, student_val: Tensor, teacher_val: Tensor) -> Tensor
             TF.kl_div(
                 input=TF.log_softmax(student_val / self._temperature, dim=-1),
                 target=TF.softmax(teacher_val / self._temperature, dim=-1),
-                reduction="batchmean",
+                reduction="sum",
             )
             * (self._temperature ** 2)
-        )
+        ) / (student_val.shape[0] * student_val.shape[1])
```
natuan, line 359: This is correct for MLM only right? For others like QA the student_val here is [batch_size, embedding_size] and we want to average over batch_size only. Should be divided by "student_val.numel() / student_val.shape[-1]" (and keep the reduction as "sum")

natuan (changes requested):

src/sparseml/pytorch/optim/modifier_distillation.py
spacemanidol: Im pretty sure we want to do this form for all types of tasks. The sum will do the KL divergence over all sets of logits and then the division will normalize the dimensions. While this is not what we were doing for QA this is probably what we should have been doing.

natuan: We don't want to average over the last dimension because it's the one you already sum over in the KL definition (in order for it to have its proper meaning). As an example, consider a data point (question, context) in SQuAD with max_seq_length of 384; so its output by the model is a vector of 384 elements, each interpretable as the probability that the start positions would be at that element index. All these elements together make the predicted distribution, and the KL(x1) is a sum over 384 terms. The loss of this batch of 1 data point is KL(x) / 1.

spacemanidol:
dupe with eldars fix

PR #593, opened 2022-03-07
torch setuptools bugfix

rahul-tuli:
Few torch versions have introduced a subtle bug, 

When `setuptools>59.6` is used with `torch 1.9 or 1.10`, an `AttributeError` is raised on import of `torch.tensorboard.SummaryWriter`, this causes `sparseml.image_classification.train` to fail. The goal of this PR is to  print occurance and ignore tensorboard logger when this error is observed


xkazuka (dismissed):

bfineran (approved):

bfineran (dismissed):
need to update for quality checks

bfineran (approved):

markurtz (approved):

PR #596, opened 2022-03-08
[WIP] Trainer wrapper for image classification model

rahul-tuli:
The Goal of this PR is to refactor logic and make training image classification models easier, 
Right now this PR is a work in progress, and any ideas on a more extensible design are welcome

```python
trainer = ImageClassificationTrainer(
            checkpoint_path = 'models/resnet50.pth',
            recipe_path='recipes/my-recipe.yaml',
            dataset = 'imagenette',
            dataset_path = 'dataset',
            train_batch_size= 32,
            test_batch_size = 64,

)
trainer.run_training()
```

### DIFF
--- a/src/sparseml/pytorch/models/registry.py
+++ b/src/sparseml/pytorch/models/registry.py
@@ -207,8 +207,11 @@ class ModelRegistry(object):
         if not isinstance(key, List):
             key = [key]
 
-        def decorator(const_func):
-            wrapped_constructor = ModelRegistry._registered_wrapper(key[0], const_func)
+        def decorator(constructor_func):
+            wrapped_constructor = ModelRegistry._registered_wrapper(
+                key[0],
+                constructor_func,
+            )
 
             ModelRegistry.register_wrapped_model_constructor(
                 wrapped_constructor,
@@ -293,10 +296,10 @@ class ModelRegistry(object):
     @staticmethod
     def _registered_wrapper(
         key: str,
-        const_func: Callable,
+        constructor_func: Callable,
     ):
-        @merge_args(const_func)
-        @wrapper_decorator(const_func)
+        @merge_args(constructor_func)
+        @wrapper_decorator(constructor_func)
         def wrapper(
             pretrained_path: str = None,
             pretrained: Union[bool, str] = False,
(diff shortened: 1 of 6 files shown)


src/sparseml/pytorch/models/registry.py
```diff
@@ -207,8 +207,11 @@ def register(
         if not isinstance(key, List):
             key = [key]
 
-        def decorator(const_func):
-            wrapped_constructor = ModelRegistry._registered_wrapper(key[0], const_func)
+        def decorator(constructor_func):
```
bfineran, line 210: confirming that the changes here are just for readability?

src/sparseml/pytorch/utils/trainer.py
```diff
@@ -0,0 +1,562 @@
...
+
+_LOGGER = logging.getLogger(__name__)
+
+
+class ImageClassificationTrainer:
+    "Utility class for training image classification models"
+
+    def __init__(
+            self,
+            checkpoint_path: str,
```
bfineran, line 48: we definitely need to cut down on the number of arguments we have here. for example, the dataloaders should be able to be constructed outside of the trainer and then passed in. Additionally, let's also keep model saving outside of the trainer for now. and can you remind me on the blocks we saw for model construction needing to be in the trainer class?

src/sparseml/pytorch/utils/trainer.py
```diff
@@ -0,0 +1,562 @@
...
+        self.device = device
+        self.debug_steps = debug_steps
+        self.save_best_after = save_best_after
+        self.save_epochs = save_epochs
+
+        self.dataset_kwargs = dataset_kwargs or {}
+        self.optim_kwargs = optim_kwargs or {}
+
+        self.world_size = int(os.environ.get("WORLD_SIZE", 1))
+        self.env_rank = int(os.environ.get("RANK", -1))
```
bfineran, line 103: DDP should be handled outside of the trainer, if possible, is there any reason it has to be in here? most of the rank specific calls should happen outside the trainer, right?

bfineran (changes requested):
left some initial comments, net net, can we move stuff like dataloader construction, model saving, and DDP logic to outside of the Trainer? As it is, there are too many arguments specific to our IC train script

src/sparseml/pytorch/models/registry.py
rahul-tuli: yes!

PR #599, opened 2022-03-09
mask creator refactor

bfineran:
* rename "block" mask creator to "block4"
* reintroduce formerly removed `[x,y]` block mask creator/notation
* helper function for creating basic mask creators
* checked through existing zoo and example recipes to confirm none are affected


markurtz (approved):

mgoin (approved):
nice!

PR #600, opened 2022-03-09
updating batch size for optimizer

spacemanidol:


src/sparseml/transformers/sparsification/trainer.py
```diff
@@ -202,9 +202,16 @@ def create_optimizer(self):
         if not self.manager:
             return
 
+        num_devices = (
+            torch.distributed.get_world_size()
+            if torch.distributed.is_initialized()
+            else self.args._n_gpu
+        )
+        if num_devices < 1:
```
natuan, line 210: @spacemanidol do you know if we need these two lines? If is_initialized is true then can get world size returns -1?

natuan, line 210: @spacemanidol do you know if we need these two lines? If is_initialized is true then can get world size returns -1?

jeanniefinks:
Product being deprecated by early June 2, 2025. Closing PR and will not resolve.

PR #607, opened 2022-03-11
Hotfix: ACDCPruningModifier; allow arguments `start_epoch`, `end_epoch`, `update_frequency` to be floats.

dbogunowicz:
In SparseML recipes, by convention, modifier arguments `start_epoch`, `end_epoch` and `update_frequency` are assumed to be floats. 

In `ACDCPruningModifier`, we are using those three arguments to figure out the mapping from epoch number to "ac/dc phase" (either compression or decompression). This mapping happens in L260. `src/sparseml/pytorch/sparsification/pruning/modifier_pruning_acdc.py`:

```
 compression_phases = [
            math.floor(epoch / update_frequency) % 2 == 0.0
            for epoch in range(start_epoch, end_epoch)
]
```
The `range` function requires `start_epoch` and `end_epoch` to be integers. IMO, this requirement is consistent with the implicit assumptions in the original paper/code of ACDC. This implicit assumption, to my mind, also extends to the `update_frequency` variable.

At the current state, the user has to specify those three arguments explicitly as integers in a recipe. This PR asserts that the user has the freedom to either define them as int or float, as long as this float can be directly evaluated as an integer.

E.g `start_epoch` can be `6` or `6.0` but cannot be `6.1`.

### DIFF
--- a/src/sparseml/pytorch/sparsification/pruning/modifier_pruning_acdc.py
+++ b/src/sparseml/pytorch/sparsification/pruning/modifier_pruning_acdc.py
@@ -293,6 +293,7 @@ class ACDCPruningModifier(BasePruningModifier):
             raise ValueError(
                 f"The ACDCPruningModifier assumes that attributes "
                 f"`start_epoch`, `end epoch` and `update frequency` "
-                f"are integers. However, type({x}) is float."
+                f"are integers or floats, which evaluate to integers. 
+                f"However: type(x)==float and x.is_integer() == False."
             )
         return int(x)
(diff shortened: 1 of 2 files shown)


anmarques (dismissed):

src/sparseml/pytorch/sparsification/pruning/modifier_pruning_acdc.py
```diff
@@ -273,3 +278,21 @@ def _finish_on_compression(
...
+        If True, return and integer, else, raise ValueError.
+
+        :param x: an integer or a float.
+        :return: an integer
+        """
+        if isinstance(x, float) and not x.is_integer():
+            raise ValueError(
+                f"The ACDCPruningModifier assumes that attributes "
+                f"`start_epoch`, `end epoch` and `update frequency` "
+                f"are integers. However, type({x}) is float."
```
KSGulin, line 296: Might help to reword this error message a bit. As it stands, it suggest to the user that the issue is the `type()` of x and not the value

anmarques (approved):

KSGulin (approved):

PR #612, opened 2022-03-15
Fix: ACDC, stop mutating the `epoch_end` attribute

dbogunowicz:
Fixed some inconvenient design choices on my end.

`_finish_on_compression` was the function which asserted that the training finishes on the last compression phase. As a byproduct, it was changing the `end_epoch` attribute and leading to a super ugly UI:

1. User specifies ACDCPruningModifier's `end_epoch: X` in the recipe before training.
2. `_finish_on_compression( ..., end_epoch = X)` would return `end_epoch = Y`
3. After the training the manager would save the resulting recipe with a different `end_epoch: Y`....
4. ...resulting in a big wtf moment for the user (did for me :))

Now, instead of using `_finish_on_compression` method, I am just making sure that the last `update_frequency` epochs of the training are always compression epochs. This is functionally equivalent to the previous code, but does not silently mutate the value of `end_epoch`.


src/sparseml/pytorch/sparsification/pruning/modifier_pruning_acdc.py
```diff
@@ -169,8 +167,10 @@ def get_applied_sparsity_for_epoch(
             should be set to different sparsities, should return a list of those values
             in the order the parameters appear in the mask manager for this object
         """
+
         self._num_phase = math.floor((epoch - self.start_epoch) / self.update_frequency)
-        if self._num_phase % 2 == 0:
+
+        if self._num_phase % 2 == 0 and self.end_epoch - self.update_frequency > epoch:
```
bfineran, line 173: nice

bfineran (approved):

rahul-tuli (approved):

PR #615, opened 2022-03-16
Refactor `update_recipe_variables`

dbogunowicz:
The original function was super messy so I took Ben's advice and refactored it a bit.

### DIFF
--- a/src/sparseml/optim/helpers.py
+++ b/src/sparseml/optim/helpers.py
@@ -195,6 +195,68 @@ def parse_recipe_variables(
     return recipe_variables
 
 
+def _update_staged_container_value(var_key, var_value, container):
+    """
+    Extends the functionality of _update_container_value to handle staged containers.
+
+    :param var_key: The key we are attempting to find in the container.
+    :param var_value: The value which will overwrite the previous value
+        every time var_key is found in the container's keys.
+    :param container: A container generated from a YAML string of SparseML recipe
+    :return: (optionally mutated) container, as well as key_value
+        (True if var_key found in container's attributes, otherwise False)
+    """
+    key_found = False
+
+    for container_key, container_value in container.items():
+        if not isinstance(container_value, dict):
+            # checking contents of global variables
+            if var_key == container_key:
+                container[var_key] = var_value
+                key_found = True
+
+        else:
+            # checking contents of a stage
+            stage_container, stage_key_found = _update_container_value(
+                var_key, var_value, container_value
+            )
+            container[container_key] = stage_container
+            if stage_key_found:
+                key_found = True
+
+    return container, key_found
+
+
+def _update_container_value(var_key: str, var_value, container):
+    """
+    Checks whether there is at least one attribute (key) in the container, with the same
+    name as var_key. If this is the case, key_found = True, otherwise False.
+    Additionally, during checking, when var_key key is present in any of the container's
+    attributes, this attribute's value gets overwritten with var_value.
+
+    :param var_key: The key we are attempting to find in the container.
+    :param var_value: The value which will overwrite the previous value
+        every time var_key is found in the container's keys.
+    :param container: A container generated from a YAML string of SparseML recipe
+    :return: (optionally mutated) container, as well as key_value
+        (True if var_key found in container's attributes, otherwise False)
+    """
+    key_found = False
+
+    for key, value in container.items():
+        if not isinstance(value, list):
+            if var_key == key:
+                container[var_key] = var_value
+                key_found = True
+        else:
+            for i, modifier in enumerate(value):
+                if var_key in modifier.keys():
+                    container[key][i][var_key] = var_value
+                    key_found = True
+
+    return container, key_found
+
+
 def update_recipe_variables(recipe_yaml_str: str, variables: Dict[str, Any]) -> str:
     """
     :param recipe_yaml_str: YAML string of a SparseML recipe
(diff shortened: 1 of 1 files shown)


src/sparseml/optim/helpers.py
```diff
@@ -195,6 +195,68 @@ def parse_recipe_variables(
     return recipe_variables
 
 
+def _update_staged_container_value(var_key, var_value, container):
```
bfineran, line 198: let's rename `_update_staged_container_value` to `_update_staged_recipe_variable` as container is a little ambiguous outside of the scope of the functions

src/sparseml/optim/helpers.py
```diff
@@ -195,6 +195,68 @@ def parse_recipe_variables(
...
+                var_key, var_value, container_value
+            )
+            container[container_key] = stage_container
+            if stage_key_found:
+                key_found = True
+
+    return container, key_found
+
+
+def _update_container_value(var_key: str, var_value, container):
```
bfineran, line 230: let's rename `_update_container_value ` to `_update_recipe_variable` as container is a little ambiguous outside of the scope of the functions

src/sparseml/optim/helpers.py
bfineran, line 198: nit: `var_key` -> `var_name`

src/sparseml/optim/helpers.py
```diff
@@ -195,6 +195,68 @@ def parse_recipe_variables(
...
+    """
+    key_found = False
+
+    for key, value in container.items():
+        if not isinstance(value, list):
+            if var_key == key:
+                container[var_key] = var_value
+                key_found = True
+        else:
+            for i, modifier in enumerate(value):
```
bfineran, line 252: nit: ambiguous variable name `i`, prefer `idx`

src/sparseml/optim/helpers.py
```diff
@@ -195,6 +195,68 @@ def parse_recipe_variables(
...
+                key_found = True
+
+        else:
+            # checking contents of a stage
+            stage_container, stage_key_found = _update_container_value(
+                var_key, var_value, container_value
+            )
+            container[container_key] = stage_container
+            if stage_key_found:
+                key_found = True
```
bfineran, line 225: if `stage_key_found` is guaranteed to be a bool, could do `key_found = key_found or stage_key_found` or `key_found |= stage_key_found`. not safe if you think `stage_key_found` could change type somehow

src/sparseml/optim/helpers.py
```diff
@@ -195,6 +195,68 @@ def parse_recipe_variables(
...
+    """
+    Extends the functionality of _update_container_value to handle staged containers.
+
+    :param var_key: The key we are attempting to find in the container.
+    :param var_value: The value which will overwrite the previous value
+        every time var_key is found in the container's keys.
+    :param container: A container generated from a YAML string of SparseML recipe
+    :return: (optionally mutated) container, as well as key_value
+        (True if var_key found in container's attributes, otherwise False)
+    """
```
bfineran, line 208: nit: pydocs only should be added for public methods, for private, can give inline comments about logic (ie any key at top level or within a stage of the recipe that matches `var_name` will be updated to `var_value`

src/sparseml/optim/helpers.py
```diff
@@ -195,6 +195,68 @@ def parse_recipe_variables(
...
+    Additionally, during checking, when var_key key is present in any of the container's
+    attributes, this attribute's value gets overwritten with var_value.
+
+    :param var_key: The key we are attempting to find in the container.
+    :param var_value: The value which will overwrite the previous value
+        every time var_key is found in the container's keys.
+    :param container: A container generated from a YAML string of SparseML recipe
+    :return: (optionally mutated) container, as well as key_value
+        (True if var_key found in container's attributes, otherwise False)
+    """
```
bfineran, line 243: nit: pydocs only should be added for public methods, for private, can give inline comments about logic

bfineran (changes requested):
refactor looks great, left some nits

src/sparseml/optim/helpers.py
dbogunowicz: The problem which I am trying to solve is the following. Imagine a container, where there is:
- one global variable.
- three stages. 
The key we are looking for is in the **second stage**.

if we apply `key_found = key_found or stage_key_found` the following thing might happen:
0. initially `key_found = False`
1. after checking one global variable `key_found = False` (we did not find the key)
2. after checking the first stage `stage_key_found = False` (variable not in the stage)
3. after checking the second stage `stage_key_found = True` (yay, variable found)
4. after checking the third stage `stage_key_found = False` (because no variable in the third stage).

We end up with `False or False`, even though the variable was present in the second stage!

src/sparseml/optim/helpers.py
dbogunowicz: I really enjoy those situations where I get to learn more about the best rules for creating docstrings/pydocs. 

From my naive perspective, it makes sense to document some private methods if it potentially adds value. Afaik, as long as there is an underscore preceding the class/function name, this block of code will not be recognised by pydoc, so no harm here (https://stackoverflow.com/questions/12040381/using-pydoc-for-non-public-private-classes). Additionally, for some private classes/funcs it seems beneficial to add some more verbose documentation, especially for new joining ml-engs in the future.

What do you think?

bfineran: agreed that verbose documentation even for private methods is important. Keeping them out of the triple quote pydoc format will prevent them from being found and added to our official docs when we build docs eventually.

dbogunowicz: Understood.

bfineran (approved):

rahul-tuli (approved):

PR #628, opened 2022-03-21
Fix log level + add additional logging tests

KSGulin:
- Fixed error caused by `level` not being passed down correctly. 
- Added additional logging tests for `level` and `LoggerManager`
- Made minor interface changes for the loggers for better usability

### DIFF
--- a/src/sparseml/pytorch/utils/logger.py
+++ b/src/sparseml/pytorch/utils/logger.py
@@ -184,6 +184,7 @@ class LambdaLogger(BaseLogger):
                 Optional[Dict[str, float]],
                 Optional[int],
                 Optional[float],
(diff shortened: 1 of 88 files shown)


bfineran (approved):

dbogunowicz (approved):

src/sparseml/pytorch/utils/logger.py
```diff
@@ -204,6 +205,7 @@ def lambda_func(
             Optional[Dict[str, float]],
             Optional[int],
             Optional[float],
+            Optional[int],
```
anmarques, line 208: I expected to see some description about what "level" means in the comment below.

PR #634, opened 2022-03-22
Nit: fixing logging printing in transformers qa script

dbogunowicz:
Before: `WARNING  Process rank: -1, device: cuda:0, n_gpu: 4distributed training: False, 16-bits training: True`
Now `WARNING  Process rank: -1, device: cuda:0, n_gpu: 4, distributed training: False, 16-bits training: True`


KSGulin (commented):
Nice find. It looks like this is a nit across all 4 transformers tasks. Let's add the fix for:
`masked_language_modeling.py`
`text_classification.py`
`token_classification.py`

KSGulin (dismissed):

bfineran (approved):

kazukageorge (approved):

PR #635, opened 2022-03-22
ACDC: Passing `epoch = float('inf)` when applying recipe to a model.

dbogunowicz:
When applying a checkpoint recipe containing ACDCPruningModifier to a model, the modifier's `get_applied_sparsity_for_epoch` method receives `epoch=float('inf')`. This breaks the modifier (`cannot convert float infinity to integer`).
This diff is my best and quickest way how to mitigate this problem, but I am open for suggestions.


bfineran (dismissed):
good catch, this looks reasonable, main thing we want to ensure is that when epoch=inf, we apply the compression sparsity

dbogunowicz:
note to myself: I am pretty confident my diff will not influence the inference; it exclusively affects the training. But I shall double-check before merging.

bfineran (approved):

rahul-tuli (approved):
lgtm!

PR #639, opened 2022-03-23
fixing 2nd dataset load issue

spacemanidol:


src/sparseml/transformers/masked_language_modeling.py
```diff
@@ -28,7 +28,7 @@
 
 # You can also adapt this script on your own masked language modeling task.
 # Pointers for this are left as comments
-
+import pdb
```
bfineran, line 31: nit: remove pdb import

natuan (commented):
@spacemanidol looks like you need to run isort

src/sparseml/transformers/masked_language_modeling.py
spacemanidol: fixed

bfineran (approved):

markurtz (approved):

PR #643, opened 2022-03-25
prevent block mask creator from mutating modifier mask type

bfineran:


KSGulin (approved):

bfineran:
all credit to @KSGulin

PR #650, opened 2022-03-28
[MFAC] cache previous GPUtil return

### LINKED ISSUE #645: M-FAC-like pruner crashes due to the GPUtil not seeing any GPUs
I am experimenting with `M-FAC` like `BaseGradualPruningModifier` and what I've Implemented is the class inheriting
from either `MFACPruningModifier` and `ACDCPruningModifier` which I've called `ACDCMFACPruningModifier`.

```
@PyTorchModifierYAML()
class ACDCMFACPruningModifier(ACDCBasePruningModifier, MFACPruningModifier):
   ....
```

```
I init the Modifier via the SparseML recipe:

  - !ACDCMFACPruningModifier
    compression_sparsity: 0.75
    start_epoch: 150
    end_epoch: 500
    update_frequency: 10
    params: ['re:(.*blocks.*.attn.qkv.weight)|(.*blocks.*.attn.proj.weight)|(.*blocks.*.mlp.fc.*weight)'] 
    mask_type: unstructured
    global_sparsity : True
    num_grads: 256
    fisher_block_size: 64
```

And for the long run, as in this recipe at the start of the compression stage training crashes on this step:

```
2022-03-27 21:37:34 sparseml.pytorch.sparsification.pruning.modifier_pruning_mfac INFO     Using Small Block Fast Fisher Inverse Implementation 
Using Small Block Fast Fisher Inverse Implementation 
...
File "/nfs/scistore14/alistgrp/dkuznede/miniconda3/envs/deep_learning/lib/python3.9/site-packages/sparseml/pytorch/sparsification/pruning/modifier_pruning_mfac.py", line 1506, in _get_free_gpu_memory 
    return [gpus_all[idx].memoryFree for idx in gpu_idx] 
  File "/nfs/scistore14/alistgrp/dkuznede/miniconda3/envs/deep_learning/lib/python3.9/site-packages/sparseml/pytorch/sparsification/pruning/modifier_pruning_mfac.py", line 1506, in <listcomp> 
    return [gpus_all[idx].memoryFree for idx in gpu_idx] 
IndexError: list index out of range 
```

I've delved into the issue and found out, that at this step:
`gpus_all = GPUtil.getGPUs()`
returns empty list. 

I've rerun the experiment with a short run with a few epochs.

    start_epoch: 2
    end_epoch: 8
    update_frequency: 1

There were no any issues and the GPUtil returned the list of all GPUs available. Training finished normally. 

The SparseML recipe is instantiated in the standard way:

    manager = ScheduledModifierManager.from_yaml(args.sparseml_recipe)
    # wrap optimizer   
    optimizer = manager.modify(model, optimizer, steps_per_epoch=len(loader_train))

Seems like it's some a hardware-like issue underhood. What could you suggest to look into?

 I am working with the nightly release of SparseML.

Version of packages installed
---
sparseml-nightly              0.12.0.20220326 
cudatoolkit               11.3.1               h2bc3f7f_2
GPUtil                        1.4.0
torch                         1.10.2

System
---
GPU - `RTX3090`
OS - `Debian GNU/Linux 10 (buster)`

KSGulin:
Potential fix for #645. The fix caches returns from `_get_free_gpu_memory()` and uses them when the function fails to execute

### DIFF
--- a/src/sparseml/pytorch/sparsification/pruning/modifier_pruning_mfac.py
+++ b/src/sparseml/pytorch/sparsification/pruning/modifier_pruning_mfac.py
@@ -1463,20 +1463,24 @@ def cache_gpu_mem_return(func):
     Cache previous return of GPUtil to be re-used in case future GPUtil call fails to
     detect available devices.
     """
-    prev_return = {"rt": []}
+    prev_return = {}
     safety_scale = 0.8
 
-    @wraps
-    def cached_gpu_mem_func(*args, **kwargs):
+    def cached_gpu_mem_func(device_idx=[], clear_cache=True):
+        key = str(device_idx)
         try:
-            prev_return["rt"] = func(*args, **kwargs)
-            return prev_return["rt"]
+            prev_return[key] = func(device_idx, clear_cache)
+            return prev_return[key]
         except Exception:
             _LOGGER.warning(
                 f"Failed to get GPU available memory. Using previous memory read, "
                 f"scaled down to {safety_scale*100:.2f}% for a safety margin"
             )
-            return [mem * safety_scale for mem in prev_return["rt"]]
+            return (
+                [mem * safety_scale for mem in prev_return[key]]
+                if key in prev_return
+                else []
+            )
 
     return cached_gpu_mem_func
 
@@ -1498,7 +1502,6 @@ def _get_free_gpu_memory(
     reading, but comes at a (small) cost to pytorch tensor allocation speed. In the case
     of very high frequency calls, it may be better to turn clear_cache off.
     """
-
     if not device_idx:
         device_idx = list(range(torch.cuda.device_count()))
     if not device_idx:
--- a/src/sparseml/pytorch/sparsification/pruning/modifier_pruning_mfac.py
+++ b/src/sparseml/pytorch/sparsification/pruning/modifier_pruning_mfac.py
@@ -1466,6 +1466,7 @@ def cache_gpu_mem_return(func):
     prev_return = {}
     safety_scale = 0.8
 
+    @wraps(func)
     def cached_gpu_mem_func(device_idx=[], clear_cache=True):
         key = str(device_idx)
         try:
@@ -1502,6 +1503,7 @@ def _get_free_gpu_memory(
     reading, but comes at a (small) cost to pytorch tensor allocation speed. In the case
     of very high frequency calls, it may be better to turn clear_cache off.
     """
+
     if not device_idx:
         device_idx = list(range(torch.cuda.device_count()))
     if not device_idx:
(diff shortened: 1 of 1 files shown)


src/sparseml/pytorch/sparsification/pruning/modifier_pruning_mfac.py
```diff
@@ -1458,6 +1458,29 @@ def _get_num_grads_for_sparsity(
...
 
+def cache_return(func):
+    """
+    Cache previous return of GPUtil to be re-used in case future GPUtil call fails to
+    detect available devices.
+    """
+    prev_return = {"rt": []}
+    safety_scale = 0.8
+
+    def wrapper(*args):
```
rahul-tuli, line 1469: can we change the name to something more intuitive rather than just `wrapper`?

src/sparseml/pytorch/sparsification/pruning/modifier_pruning_mfac.py
```diff
@@ -1458,6 +1458,29 @@ def _get_num_grads_for_sparsity(
...
 
 
+def cache_return(func):
+    """
+    Cache previous return of GPUtil to be re-used in case future GPUtil call fails to
+    detect available devices.
+    """
+    prev_return = {"rt": []}
+    safety_scale = 0.8
+
```
rahul-tuli, line 1468: Can we wrap inside a `@functools.wraps` to preserve info about the decorated function?

rahul-tuli (changes requested):
Lgtm once the wrapper is wrapped in `functools.wraps`

src/sparseml/pytorch/sparsification/pruning/modifier_pruning_mfac.py
```diff
@@ -1458,6 +1458,30 @@ def _get_num_grads_for_sparsity(
     return num_grads[sparsity_thresholds[idx]]
 
 
+def cache_return(func):
```
bfineran, line 1461: nit: rename since specific to gpu mem func

bfineran (changes requested):
shouldn't we account for the device idx in the cache?

rahul-tuli (dismissed):

KSGulin:
> shouldn't we account for the device idx in the cache?

@bfineran `_get_free_gpu_memory` returns available memory in same order that PyTorch sees the devices in, so there is no need to track device idx

KSGulin:
@bfineran @rahul-tuli Updated to track device idx after all. In case the function is re-purposed for different flows

bfineran (changes requested):
one more requested change - if for some reason there's a failure and the cache is empty, we will get a `KeyError`, we should definitely raise a `RuntimeException` or something similar if they key is not found in cache

bfineran:
also looks like quality is red @KSGulin

KSGulin:
@bfineran if cache is empty it returns an empty list, which the rest of the MFAC modifier takes as a signal to use CPU. The quality error is not replicated locally. Will investigate further

bfineran:
@KSGulin ah yup, if a user specifies that GPU should be used though, we should at least warn if falling back to CPU

KSGulin:
@bfineran good point, warning added.

bfineran (approved):

markurtz (approved):

PR #655, opened 2022-03-29
[Staged Recipes Integration] `ScheduledModifierManager` adds "framework metadata" to new metadata.

dbogunowicz:
This diff grants the `ScheduledModifierManager` the ability to insert "framework metadata" to incoming metadata.
The logic is as follows:
If new metadata is `None`, no framework metadata is added to it.
If new metadata is not `None` and there is no previous framework metadata: we add metadata.
If new metadata is not `None` and there is previous framework metadata: we overwrite previous framework metadata while raising a warning that we have done so.

Open questions:
- experiencing a bug regarding circular imports when fetching sparse ml version `from src.sparseml import version_base`. Did not spend too much time investigating, we can fix it before the merge or during our tri-weekly meeting.
- I am still not sure whether metadata=None should be equivalent to ignoring framework metadata insertion. Let's discuss that.

### DIFF
--- a/src/sparseml/optim/helpers.py
+++ b/src/sparseml/optim/helpers.py
@@ -16,11 +16,14 @@
 Helper functions for base Modifier and Manger utilities
 """
 import json
+import platform
 import re
 import warnings
 from contextlib import suppress
+from copy import deepcopy
 from typing import Any, Dict, Optional, Tuple, Union
 
+import torch
 import yaml
 
 from sparseml.utils import (
@@ -30,6 +33,13 @@ from sparseml.utils import (
 )
 from sparsezoo import Zoo
 from sparsezoo.objects import Recipe
+
+
+# from src.sparseml import version_base
+
+# having circular dependency problem with the
+# line above, hack for now
+version_base = "0.12.0"
 
 
 __all__ = [
@@ -569,6 +579,41 @@ def _check_warn_dict_difference(original_dict, new_dict):
     return new_dict
 
 
+def _add_framework_metadata(metadata, is_metadata_staged):
+    framework_metadata = {
+        "python_version": platform.python_version(),
+        "torch_version": torch.__version__,
+        "sparseml_version": version_base,
+    }
+    if is_metadata_staged:
+        for stage_name, stage_value in metadata.items():
+            shared_keys = set(stage_value.keys()).intersection(
+                set(framework_metadata.keys())
+            )
+            if shared_keys:
+                warnings.warn(
+                    "Overwriting metadata stage "
+                    f"(stage name: {stage_name}) key(s) {shared_keys} "
+                    "with new value(s) "
+                    f"{ {k:v for k,v in framework_metadata.items() if k in shared_keys} }"  # noqa E501
+                )
+            _stage_value = deepcopy(stage_value)
+            _stage_value.update(framework_metadata)
+            metadata[stage_name] = _stage_value
+    else:
+        shared_keys = set(metadata.keys()).intersection(set(framework_metadata))
+        if shared_keys:
+            warnings.warn(
+                "Overwriting metadata "
+                f"key(s) {shared_keys} "
+                f"with new value(s) "
+                f"{ {k: v for k, v in framework_metadata.items() if k in shared_keys} }"  # noqa E501
+            )
+        metadata.update(framework_metadata)
+
+    return metadata
+
+
 def validate_metadata(metadata: dict, yaml_str: str) -> dict:
     """
     Compare the metadata (previous_metadata) carried over from the recipe
@@ -577,6 +622,11 @@ def validate_metadata(metadata: dict, yaml_str: str) -> dict:
     If attempting to overwrite previous metadata with the new metadata,
     the script throws a warning and overwrites the previous metadata.
     Otherwise, it propagates the new metadata in the correct form.
+
+    Note: the function `_add_framework_metadata`, only gets executed if
+        `metadata` evaluates to True. This means that if e.g. we pass
+        `metadata = None`, framework metadata will not be added to the
+        output of this function.
 
     :param metadata: New metadata
     :param yaml_str: String representation of the recipe YAML file,
@@ -611,9 +661,16 @@ def validate_metadata(metadata: dict, yaml_str: str) -> dict:
                         checkpoint_metadata[stage_name] = _check_warn_dict_difference(
                             container[stage_name][RECIPE_METADATA_KEY], metadata
                         )
+                checkpoint_metadata = _add_framework_metadata(
+                    metadata=checkpoint_metadata, is_metadata_staged=True
+                )
             else:
                 checkpoint_metadata = _check_warn_dict_difference(
                     container[RECIPE_METADATA_KEY], metadata
+                )
+
+                checkpoint_metadata = _add_framework_metadata(
+                    metadata=checkpoint_metadata, is_metadata_staged=False
                 )
 
         return (
@@ -624,6 +681,9 @@ def validate_metadata(metadata: dict, yaml_str: str) -> dict:
 
     else:
         if metadata:
+            metadata = _add_framework_metadata(
+                metadata=metadata, is_metadata_staged=False
+            )
             return (
                 {
                     stage_name: metadata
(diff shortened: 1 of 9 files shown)


bfineran (changes requested):
looks good overall - main change I'd like to propose is adding the framework metadata at the manager level (before or after `validate_metadata`) just to make this a little less "hidden" since it will affect all metadata writes

dbogunowicz:
@bfineran 

Very good point about pushing the `add_framework_metadata` functionality to the manager level. I have decided to insert the function just after `validate_metadata`. This is mainly because the output of `validate_metadata` has a unified structure, regardless of whether the processed yaml recipe was staged or not. It always ends up being Dict[str, Dict], which helps reduce the boilerplate code in `add_framework_metadata`.

I have also decided to alter `add_framework_metadata` s.t. the returned metadata always contains the framework metadata. Previously, if the metadata for a stage was `None`, I skipped adding the framework metadata, but I find it silly after reconsideration. I am quite confident that from now on we will rarely encounter "empty" metadata, and by default always adding framework metadata is the correct thing to do.

src/sparseml/optim/helpers.py
```diff
@@ -16,13 +16,17 @@
...
 """
 import json
+import platform
 import re
 import warnings
 from contextlib import suppress
+from copy import deepcopy
 from typing import Any, Dict, Optional, Tuple, Union
 
+import torch
```
bfineran, line 26: cannot have a torch specific import in the generic helpers file (ie `keras` and `tensorflow` managers won't work with this), will suggest a small change to `add_framework_metadata` to support adding in extra framework specific metadata

src/sparseml/optim/helpers.py
```diff
@@ -569,6 +574,44 @@ def _check_warn_dict_difference(original_dict, new_dict):
     return new_dict
 
 
+def add_framework_metadata(metadata: Dict[str, Dict]) -> Dict[str, Dict]:
```
bfineran, line 577:
```suggestion
def add_framework_metadata(metadata: Dict[str, Dict], **extra_metadata) -> Dict[str, Dict]:
```
let's do something like this where each framework will pass in their specific version and then update the metadata dict with this metadata as well

i.e. for pytorch it will pass `add_framework_metadata(metadata, torch=torch.__version__)`

src/sparseml/optim/helpers.py
```diff
@@ -569,6 +574,44 @@ def _check_warn_dict_difference(original_dict, new_dict):
...
+                    f"stage, stage name: {stage_name}"
+                    if stage_name != RECIPE_METADATA_KEY
+                    else ""
+                )
+                warning_msg = (
+                    f"Overwriting metadata {warning_if_stage} key(s) "
+                    f"{shared_keys} with new value(s) "
+                    f"{ {k:v for k,v in framework_metadata.items() if k in shared_keys} }"  # noqa E501
+                )
+                warnings.warn(warning_msg)
```
bfineran, line 607: nit: use logger.warning instead of warnings

src/sparseml/optim/helpers.py
```diff
@@ -569,6 +574,44 @@ def _check_warn_dict_difference(original_dict, new_dict):
...
+            _stage_value = deepcopy(stage_value)
```
bfineran, line 608: nit: prefer not to name locally scoped variables with a `_` prefix

bfineran (changes requested):
looking good. question:

so for recipes with existing metadata and no new metadata passed in, we don't want to override the metadata (ie for checkpoint recipes) how do we make sure that we don't inject the framework metadata here?
Later on would also be nice to add a check that the checkpoint metadata matches the framework metadata, but we can add that in later

one more thought is we could nest the framework metadata into the metadata dicts so that they will be easier to track and contain for later (can also nest the integration specific metadata)

src/sparseml/optim/helpers.py
dbogunowicz: good catch, i will apply it to `_check_warn_dict_difference` as well.

dbogunowicz:
@bfineran great catches and good point on the logic of injection of framework metadata. I have a proposal of the solution for this issue ready and will propose it to you during our meeting today.

Once this is resolved, this PR will be ready for the next review. The nits, as well as the request to have framework metadata nested have been addressed it the recent commit.

dbogunowicz:
@bfineran ready for the next round.

src/sparseml/pytorch/optim/manager.py
```diff
@@ -281,14 +284,25 @@ def from_yaml(
...
+        validated_metadata = validate_metadata(metadata, yaml_str)
 
-        manager = ScheduledModifierManager(modifiers=modifiers, metadata=metadata)
+        if metadata is not None:
+            logging.info(
+                "The new `metadata` has been passed to the manager. "
+                f"This will change the metadata in the recipe {file_path}"
+                f"including updating the metadata key {FRAMEWORK_METADATA_KEY} "
+                "to reflect your current setup."
+            )
```
bfineran, line 298: instead of this log, let's just do an info log after the metadata is complete that prints out the entire metadata.  If new metadata is passed from the from yaml constructor, then it likely won't have the framework metadata key so does not make sense to log that we are updating it

bfineran: let's also do this for the other framework managers.

src/sparseml/pytorch/optim/manager.py
```diff
@@ -281,14 +284,25 @@ def from_yaml(
...
-
         if add_modifiers:
             modifiers.extend(add_modifiers)
 
-        metadata = validate_metadata(metadata, yaml_str)
+        validated_metadata = validate_metadata(metadata, yaml_str)
 
-        manager = ScheduledModifierManager(modifiers=modifiers, metadata=metadata)
+        if metadata is not None:
+            logging.info(
```
bfineran, line 293: nit: prefer `_LOGGER.info`

bfineran (commented):
logic looks good now.  IMO we don't need to warn so much about overwriting framework metadata because in general this will never happen since framework metadata is only added if `metadata != None` which happens for training recipes

dbogunowicz:
@bfineran example print of the log from the new method of  BaseManager` from`optim.manager.py` -> `self._info_log_metadata()` (so that you don't have to checkout my branch and test it yourself):
```
The new `metadata` has been passed to the manager. Metadata supplied to the recipe:	
{
 "stage_0": {
  "metadata": "None",
  "level": 0,
  "framework_metadata": {
   "python_version": "3.8.10",
   "sparseml_version": "0.12.0.20220407",
   "torch_version": "1.9.1+cu111"
  }
 },
 "stage_1": {
  "metadata": "None",
  "level": 1,
  "framework_metadata": {
   "python_version": "3.8.10",
   "sparseml_version": "0.12.0.20220407",
   "torch_version": "1.9.1+cu111"
  }
 },
 "stage_3": {
  "metadata": null,
  "level": 3,
  "framework_metadata": {
   "python_version": "3.8.10",
   "sparseml_version": "0.12.0.20220407",
   "torch_version": "1.9.1+cu111"
  }
 }
}
```

src/sparseml/optim/manager.py
```diff
@@ -520,6 +527,31 @@ def qat_active(self, epoch: float) -> bool:
             else False
         )
 
+    def _info_log_metadata(self):
+        logging.info(
```
bfineran, line 531: nit: create a logger object `_LOGGER` like other files in repo and switch to `_LOGGER.info`

src/sparseml/optim/manager.py
```diff
@@ -520,6 +527,31 @@ def qat_active(self, epoch: float) -> bool:
...
+            "The new `metadata` has been passed to the manager. "
+            "Metadata supplied to the recipe:\t"
+        )
+        logging.info(json.dumps(self._metadata, indent=1))
```
bfineran, line 535: to make the logging cleaner let's change to:
```python
metadata_str = json.dumps(self._metadata, indent=1)
_LOGGER.info(f"Created recipe manager with metadata {metadata_str}")
```

bfineran (commented):
quick suggestion to clean up log

bfineran (dismissed):

KSGulin (dismissed):

KSGulin (approved):

bfineran (approved):

PR #659, opened 2022-03-30
Add save fcn to loggers and implement for wandb

eldarkurtic:


eldarkurtic:
It might be a good idea to raise `NotImplementedError()` instead of `return False` in the base class. If you agree, I can push an update for it.

bfineran (commented):
thanks @eldarkurtic, let's get this propagated to the `LoggerManager` as well so it is easier to surface to integrations / within our API

bfineran (approved):

KSGulin (approved):

PR #661, opened 2022-03-31
[Add tests] To hotfix regarding the saving of `None` metadata

dbogunowicz:
Tests related to https://github.com/neuralmagic/sparseml/pull/660.

Asserting that:

1. While Rahul's fix corrects the behavior of function `metadata_to_string_lines` in `optim/manager.py` for an unstaged recipe, I am making sure that the logic is foolproof for staged recipes.
2. Once we take a recipe without metadata (staged or non-staged) and save it using `manager.save()`, the resulting YAML file can be once again correctly digestible when reused. I was mostly concerned about what happens if we consume a recipe with `__metadata__:   <empty> `. Fortunately, it is being parsed correctly.

- [x] use str(manager) instead of saving recipes across all metadata tests

### DIFF
--- a/tests/sparseml/pytorch/optim/test_manager.py
+++ b/tests/sparseml/pytorch/optim/test_manager.py
@@ -468,6 +468,68 @@ def test_lifecycle_manager_staged(
     assert final_recipe == expected_recipe
 
 
+TWO_STAGES_RECIPE = """version: 1.1.0
+
+stage_0:
+
+  stage_0_modifiers:
+      - !EpochRangeModifier
+          end_epoch: 3.0
+          start_epoch: 0.0
+
+      - !GMPruningModifier
+          end_epoch: 3.0
+          final_sparsity: 0.8
+          global_sparsity: False
+          init_sparsity: 0.2
+          inter_func: cubic
+          leave_enabled: True
+          mask_type: unstructured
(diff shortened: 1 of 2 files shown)


tests/sparseml/pytorch/optim/test_manager.py
```diff
@@ -468,6 +468,68 @@ def test_lifecycle_manager_staged(
...
+@pytest.mark.parametrize(
+    "recipe",
+    [
+        STANDARD_RECIPE_1,  # passing standard recipe without metadata
+        TWO_STAGES_RECIPE,  # passing staged recipe without metadata
+    ],
+)
+def test_lifecycle_manager_staged_no_metadata(recipe):
+    temp_dir = tempfile.mkdtemp()
+    recipe_path = os.path.join(temp_dir, "recipy.yaml")
```
rahul-tuli, line 518: nit: Maybe this is a typo
```suggestion
    recipe_path = os.path.join(temp_dir, "recipe.yaml")
```

dbogunowicz: Yeah, I keep making this same typo over and over 🥇

bfineran (approved):

KSGulin (approved):

PR #663, opened 2022-03-31
Quantization refactor

anmarques:
Changes in quantization:

1. Created flag to remove Q/DQ from output activations for conv layers. Set to True by default.
2. Changed default to True for flag to remove Q/DQ from output activations for linear layers.
3. Created flag to change number of bits for weight activation (similar to the already existing flag for activation bits).
4. Created flag to avoid quantization of BatchNorm layers when fusing is not used. Set to True by default.
5. Created flag to enable quantization compatible with tensorrt. Set to False by default.
    - Sets activation quantization to symmetric and UINT8.
    - Sets default fusing function to 'no_fuse'.
6. Created a wrapper module (BNWrapper) that allows one to freeze BatchNorm statistics when BN layers are not fused. PyTorch default only works when BN layers are fused.
7. Crated a wrapper module (_AddReLU) for the FloatFunctional class that performs the Add + ReLU operations in ResNet. This wrapper enables the module to be wrapped by a QATWrapper when quantized, and is set to quantize only the first input.


src/sparseml/pytorch/sparsification/quantization/helpers.py
```diff
@@ -164,9 +330,15 @@ def __init__(
         output_qconfigs: Union[
             "torch.quantization.QConfig", str, List["torch.quantization.QConfig"]
         ] = "asymmetric",
+        symmetric_activations: Optional[bool] = None,
```
bfineran, line 333: nit: make sure to update class docstring

src/sparseml/pytorch/sparsification/quantization/helpers.py
```diff
@@ -433,68 +708,92 @@ def remove_activation_qat_by_layer_name(module: Module, layer_class_names: List[
...
+    if weight_dtype is None:
+        _weight_dtype = torch.qint8
+    else:
+        _weight_dtype = weight_dtype
 
-    weight_observer_kwargs.update(weight_qconfig_kwargs or {})
-    weight_observer = torch_quantization.FakeQuantize.with_args(
-        **weight_observer_kwargs,
+    weight_observer = get_observer(
+        _symmetric_weights, _weight_dtype, weight_bits, False, weight_qconfig_kwargs
```
bfineran, line 764: why hardcode reduce range to false?

src/sparseml/pytorch/sparsification/quantization/helpers.py
```diff
@@ -433,68 +708,92 @@ def remove_activation_qat_by_layer_name(module: Module, layer_class_names: List[
...
-        qscheme=weight_qscheme,
-        reduce_range=reduce_range,
+    activation_observer = get_observer(
+        symmetric_activations,
+        activation_dtype,
+        activation_bits,
+        reduce_range,
+        activation_qconfig_kwargs,
     )
+    if symmetric_weights is None:
```
bfineran, line 753: why not just default to True?

src/sparseml/pytorch/sparsification/quantization/helpers.py
```diff
@@ -433,68 +708,92 @@ def remove_activation_qat_by_layer_name(module: Module, layer_class_names: List[
 
 
 def get_qat_qconfig(
-    symmetric_activations: bool = False,
-    symmetric_weights: bool = True,
+    symmetric_activations: Optional[bool] = None,
```
bfineran, line 711: why are we defaulting variables to None here instead of setting a default?

src/sparseml/pytorch/sparsification/quantization/modifier_quantization.py
```diff
@@ -476,15 +533,15 @@ def _check_quantization_update(
...
+                quant_module.apply(freeze_bn_stats)
             self._bn_stats_frozen = True
 
     def _enable_module_qat(self, module: Module):
         # fuse module Conv-BNs
-        if (
-            self._model_fuse_fn_name is not None
-            and self._model_fuse_fn_name != "no_fuse"
-        ):  # module class fn
+        if self.model_fuse_fn_name == "conv_bn_relus":
```
bfineran, line 541: looks like this changes the default behavior so "conv_bn_relus" needs to be passed in? to maintain recipe backwards compatibility, we need to maintain the previous behavior of None using the default function

src/sparseml/pytorch/sparsification/quantization/modifier_quantization.py
```diff
@@ -494,38 +551,105 @@ def _enable_module_qat(self, module: Module):
...
+            activation_bits=self.activation_bits,
+            weight_bits=self.weight_bits,
         )
+
+        # prepare each module / submodule for quantization
         for name, quant_module in self._modules_to_quantize:
             # wrap any modules with wrap_qat set to True as QATWrapper(s)
             configure_module_qat_wrappers(
                 quant_module,
+                symmetric_activations=_symmetric_activations,
```
bfineran, line 618: looking at all these repeated arguments, it may be time to pull some out into a dict or dataclass to make passing easier

src/sparseml/pytorch/sparsification/quantization/modifier_quantization.py
```diff
@@ -494,38 +551,105 @@ def _enable_module_qat(self, module: Module):
...
+        # this fix wraps BN modules adding with a new module class that supports
+        # methods related to freezing/unfreezing BN statistics.
+        configure_module_bn_wrappers(module)
+
+        # set qconfig.
+        # if tensorrt flag is used, set activation and weights to symmetric
+        # quantization.
+        # otherwise, use the default values set in get_qat_qconfig
+        if self.tensorrt:
+            _symmetric_activations = True
```
bfineran, line 591: nit: prefer to only use leading `_` in names for class/module names that are meant to be private

src/sparseml/pytorch/sparsification/quantization/modifier_quantization.py
```diff
@@ -494,38 +551,105 @@ def _enable_module_qat(self, module: Module):
                     )
                 )
             module_fuse_fn(**self._model_fuse_fn_kwargs)
-        elif self._model_fuse_fn_name is None:  # default auto fn
-            self._model_fuse_fn_kwargs["inplace"] = True
-            fuse_module_conv_bn_relus(module, **self._model_fuse_fn_kwargs)
 
-        activation_qconfig_kwargs = self._get_updated_activation_qconfig_kwargs()
+        # build list of layer types that should not quantize output activations
```
bfineran, line 555: any reason this block is up here before the function call? Also given how many lines the list takes, let's try and get this moved to the helpers.py file

bfineran (dismissed):
looks good overall, left a few comments that once resolved should be in a reasonable state to land these features. one concern is worry over the growing number of arguments being passed around to create qconfigs

Will put up a separate PR to facilitate ONNX export of the new ConvInteger ops to link this PR with #644

bfineran:
looks like this also breaks some tests as well since we change the default behavior for quantizing major ops. will take a look

bfineran:
one other thing to note is that I am very wary / leaning against defaulting to using the new quantization scheme for Convs as this will break existing workflows. I do think that we should eventually default, let's sync on a strategy for this

src/sparseml/pytorch/sparsification/quantization/helpers.py
anmarques: Prior to refactor, reduce range would affect both weights and activations. My understanding is that reduce range are in place to guard against out-of-range values and thus should only affect activations (since weights are known and cannot become out-of-range at runtime). At very least we need to decouple reduce range from activations and weights. I hardcoded the weights to False to avoid adding another argument.

src/sparseml/pytorch/sparsification/quantization/helpers.py
anmarques: I agree that setting it to True makes sense here.

src/sparseml/pytorch/sparsification/quantization/helpers.py
anmarques: See reply above.

src/sparseml/pytorch/sparsification/quantization/modifier_quantization.py
anmarques: The default behavior still is "conv_bn_relus". self.model_fuse_fn_name is a wrapper for self._model_fuse_fn_name. I handle the None case inside the wrapper now (lines 237-254). This is needed because the tensorrt flag changes the default to 'no_fuse'.

src/sparseml/pytorch/sparsification/quantization/modifier_quantization.py
anmarques: See reply above about new dataclass.

src/sparseml/pytorch/sparsification/quantization/modifier_quantization.py
anmarques: Resolved.

src/sparseml/pytorch/sparsification/quantization/modifier_quantization.py
anmarques: Done.

src/sparseml/pytorch/sparsification/quantization/helpers.py
```diff
@@ -69,6 +92,177 @@
...
+        else:
+            return True
+
+    @symmetric_weights.setter
+    def symmetric_weights(self, value: bool):
+        if self._symmetric_weights is None:
+            self._symmetric_weights = value
+
+
+class BNWrapper(Module):
```
bfineran, line 159: if we don't plan on using this outside of this file, let's rename to `_BNWrapper` and move to bottom of file. Also in the docstring, should specify that this is for when the BN is not fused

src/sparseml/pytorch/sparsification/quantization/modifier_quantization.py
```diff
@@ -494,50 +560,79 @@ def _enable_module_qat(self, module: Module):
...
+        # quantization.
+        # otherwise, use the default values set in QConfigProperties
+        qproperties = QConfigProperties(
+            activation_bits=self.activation_bits,
+            weight_bits=self.weight_bits,
+            activation_qconfig_kwargs=self.activation_qconfig_kwargs,
             weight_qconfig_kwargs=self.weight_qconfig_kwargs,
+            reduce_range=self.reduce_range,
         )
+        if self.tensorrt:
```
bfineran, line 592: let's add a log here stating that we are overriding quantization to target tensorrt (symmetric int8 quantization)

bfineran (commented):
big fan of these updates @anmarques. few small comments. going to put up a PR for testing and merge on your approval @anmarques;.

src/sparseml/pytorch/sparsification/quantization/helpers.py
anmarques: Done

src/sparseml/pytorch/sparsification/quantization/helpers.py
```diff
@@ -432,69 +608,57 @@ def remove_activation_qat_by_layer_name(module: Module, layer_class_names: List[
...
 
 
-def get_qat_qconfig(
-    symmetric_activations: bool = False,
-    symmetric_weights: bool = True,
-    reduce_range: bool = False,
-    activation_qconfig_kwargs: Optional[Dict[str, Any]] = None,
-    weight_qconfig_kwargs: Optional[Dict[str, Any]] = None,
-) -> "torch.quantization.QConfig":
+def get_qat_qconfig(qproperties: QConfigProperties) -> "torch.quantization.QConfig":
```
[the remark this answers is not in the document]
anmarques: 😂

src/sparseml/pytorch/sparsification/quantization/modifier_quantization.py
anmarques: Done.

bfineran:
merged in export flow (#644)

bfineran:
merged in fixes for testing and additional tests from #685

markurtz (dismissed):

spacemanidol (dismissed):

markurtz (dismissed):

rahul-tuli (dismissed):

mgoin (approved):

markurtz (approved):

spacemanidol (commented):
look good

Issue #674, opened 2022-04-05
Wrong scaling of KL-divergence calculation for multi-dim inputs


eldarkurtic:
**Describe the bug**
When inputs are K-dimensional (K > 2), as in the masked-language-modelling case with `(batch_size, max_sequence_size, vocabulary_size)`, the `reduction="batchnorm"` scales only with the size of the first dimension [see pytorch docs](https://pytorch.org/docs/stable/generated/torch.nn.KLDivLoss.html#:~:text=elif%20reduction%20%3D%3D%20%22batchmean%22%3A%20%20%23%20mathematically%20correct%0A%20%20%20%20loss%20%3D%20loss_pointwise.sum()%20/%20input.size(0)).

**Expected behavior**
The kl-div should be scaled with "total num. of elements" / "num. of elements in the support of the distribution"(i.e. the last dimension `dim=-1` used for `softmax` calculation).

**Errors**
Extremely huge KD loss values in MLM experiments, which make it hard to converge.

PR #675, opened 2022-04-05
Fix KL-div calculation when inputs are multidim

eldarkurtic:
Fixes https://github.com/neuralmagic/sparseml/issues/674


markurtz (approved):
Thanks Eldar

bfineran (approved):

Issue #676, opened 2022-04-06
Some Mappings missed when checking for Dict type only


eldarkurtic:
**Describe the bug**
In the case of masked-language-modeling experiment, inputs to the model (outputs from the tokenizer) are instances of the `BatchEncoding` class ([HF source code](https://github.com/huggingface/transformers/blob/198c335d219a5eb4d3f124fdd1ce1a9cd9f78a9b/src/transformers/tokenization_utils_base.py#L167)) which is a subclass of `UserDict`.  
Unfortunately, a `UserDict` is not an instance of the `Dict` class and checks for `isinstance(tensors, Dict)` will be completely missed.

```python
from collections import UserDict
from typing import Dict

assert not isinstance(UserDict(), Dict)
``` 

**To Reproduce**
1) Run a masked-language-modelling experiment with KD and pytorch 1.11 version. All checks for `isinstance(tensors, Dict)` will be missed and result with many exceptions.

2) Run a masked-language-modelling experiment with KD and any pytorch version prior to 29 Nov. 2021, with disabled pin_memory feature. 

**Additional Context**
This bug didn't appear in current experiments with pytorch versions before 29 Nov. 2021 because all runs by default use `pin_memory = True`. And the `pin_memory = True` had a small issue where input `data` of any `Mapping` type would always be converted to `Dict`, instead of preserving the original `type(data)`. With this issue fixed in newer versions of pytorch, `pin_memory = True` will indeed return the original type, which in our MLM case is `BatchEncoding(UserDict)`, and this will miss all `isinstance(tensors, Dict)` checks.

Relevant links:
- [pin_memory update commit](https://github.com/pytorch/pytorch/commit/f776f30780d12ab0385e4010b07d88327a8abfd5)
- [new version of pin_memory](https://github.com/pytorch/pytorch/blob/93e650b8f7bd582746f12a0a1f0b94d76cce5e32/torch/utils/data/_utils/pin_memory.py#L53)

eldarkurtic:
Reopening because the same issue happens at a different place now.

eldarkurtic:
Reopening because the same issue happens at a different place now.

PR #677, opened 2022-04-06
Fix missed checks for Mapping types

eldarkurtic:
Fixes https://github.com/neuralmagic/sparseml/issues/676

To discuss: I've noticed that there are some other utility functions in `src/sparseml/pytorch/utils/helpers.py` that could be affected by this issue of missing the `isinstance(tensor, Dict)` check when a `tensor` is dict-like but fails the `isinstance` check (like for example `UserDict`). I've fixed only functions that affected the masked-language-modeling experiments, and where I am sure that the code inside the `isinstance(tensor, Mapping)` is calling methods that all `Mapping`s should have (for example `.items()`, `.values()`, `**tensors`). 
Let me know if it makes sense to check other utility functions too.


markurtz (approved):

bfineran (approved):

markurtz:
@eldarkurtic look like we have a conflict now with some code that landed to enable DistilBERT. Could you take a look through for resolving?

eldarkurtic:
@markurtz should be resolved and squashed now

bfineran (approved):

KSGulin (approved):

PR #678, opened 2022-04-06
fix mask from threshold logic to mask exactly target number

bfineran:
in the general case, parameter sparsity masks are currently created by masking all values beneath a threshold where the threshold is the k'th largest value in the parameter and k = `round(target_sparsity * parameter.size)`. In general, this gives us a mask at the target sparsity. However, if multiple elements have the same value of the threshold, this could cause the mask to not threshold on the truly desired element.

Further, this could cause reproducibility issues in creating masks with the exact same sparsity from the same recipe but with different weights.  This especially comes up when reloading weights from structured pruned models and we rely on being able to first prune a model to the exact recipe sparsity before loading the thinned weights.

This PR proposes a fix to reconcile the number of pruned values when thresholding fails.

This PR also adds a check that mask creators prune to exactly the target sparsity (note that the StructuredPruningModifier tests will fail on this check prior to the fixes in this PR)

I'm not sure the over_masked case is possible given our setup (should always prune under all values that equal the threshold) am open to taking it out due to this. I also spent more time than I'd like to admit working on a version of the fix using tensor ops, but it usually ended up with extra steps (considering in most models, there will be minimal overlap of thresholds), or still having some off by one issues.


src/sparseml/pytorch/sparsification/pruning/mask_creator.py
```diff
@@ -130,9 +130,24 @@ def create_sparsity_masks(
                 masks.append(tensor.new_ones(tensor.shape))
                 continue
 
+            num_elem = tensor.numel()
+            target_num_mask = round(num_elem * sparsity_target)
             min_val = tensor.min().item()
+
             if threshold.item() > min_val:
-                masks.append((tensor > threshold).type(tensor.type()))
+                threshold_mask = (tensor > threshold).type(tensor.type())
```
KSGulin, line 138: Looks like a 1 off error here. Threshold is calculated so that all params `< threshold` should be pruned. By Keeping params `> threshold`, we're pruning 1 extra each time. A simple `>=` should fix it, but I'm not sure if there's a risk of this not evaluating properly with floats.
```suggestion
                threshold_mask = (tensor >= threshold).type(tensor.type())
```

KSGulin (commented):
There is a 1-off error. Otherwise looks like a great fix!

src/sparseml/pytorch/sparsification/pruning/mask_creator.py
KSGulin, line 138: A one-off error occurs here, but it seems it stems from `_threshold_from_sparsity()` and not from the changes here. If threshold is meant to be the value s.t. all tensors > threshold are kept, then line 189 should be changed to:
`round(sparsity * tensor.numel()) - 1`

bfineran:
good catch @KSGulin resolved!

spacemanidol (approved):

rahul-tuli (approved):

PR #680, opened 2022-04-06
Long mask_type backwards compatibility

KSGulin:
Yolov5 quantized models on sparsezoo have length 4 mask_types which are not currently supported (i.e. [1, 4, 1, 1]. This PR adds backwards compatibility for those masks.

### DIFF
--- a/src/sparseml/sparsification/modifier_pruning.py
+++ b/src/sparseml/sparsification/modifier_pruning.py
@@ -168,7 +168,7 @@ class GMPruningModifier(BaseModifier, BaseScheduled, BaseUpdate):
         self._leave_enabled = convert_to_bool(leave_enabled)
         self._inter_func = inter_func
         self._mask_type = mask_type
-        if len(self._mask_type) > 2 and all(i==1 for i in mask_type[2:]):
+        if len(self._mask_type) > 2 and all(i == 1 for i in mask_type[2:]):
             self._mask_type = self._mask_type[:2]
         self._leave_enabled = convert_to_bool(leave_enabled)
 
--- a/src/sparseml/sparsification/modifier_pruning.py
+++ b/src/sparseml/sparsification/modifier_pruning.py
@@ -168,7 +168,7 @@ class GMPruningModifier(BaseModifier, BaseScheduled, BaseUpdate):
         self._leave_enabled = convert_to_bool(leave_enabled)
         self._inter_func = inter_func
         self._mask_type = mask_type
-        if len(self._mask_type) > 2 and all(i == 1 for i in mask_type[2:]):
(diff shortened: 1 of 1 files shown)


bfineran (commented):
@KSGulin LGTM, please check quality issue.

src/sparseml/sparsification/modifier_pruning.py
```diff
@@ -168,6 +168,8 @@ def __init__(
         self._leave_enabled = convert_to_bool(leave_enabled)
         self._inter_func = inter_func
         self._mask_type = mask_type
+        if len(self._mask_type) > 2 and all(i==1 for i in mask_type[2:]):
```
bfineran, line 171:
```suggestion
        if len(self._mask_type) > 2 and all(i == size for size in mask_type[2:]):
```

bfineran (dismissed):

src/sparseml/sparsification/modifier_pruning.py
```diff
@@ -168,6 +168,8 @@ def __init__(
         self._leave_enabled = convert_to_bool(leave_enabled)
         self._inter_func = inter_func
         self._mask_type = mask_type
+        if len(self._mask_type) > 2 and all(i == 1 for i in mask_type[2:]):
+            self._mask_type = self._mask_type[:2]
```
dbogunowicz, line 172: The idea is that  `mask_types` with `len(mask_types)=4` e.g. `[1, 4, 1, 1]` should be stripped down from the last two ones so that they end up being `[1,4]` right?

I am wondering about this condition `len(self._mask_type) > 2`. Is there a possibility that `len(self._mask_type) == 3`?
If no, we can change the first condition to ` if len(self._mask_type) == 4`
If yes, would we be still removing the last two values e.g. if `len(self._mask_type) = [3, 2, 1]`--> we get `[3,2]`.

KSGulin: @dbogunowicz good question. It's currently enforced that `len(mask_type) <= 2`. Anything otherwise would throw an error. The only longer masks I've encountered are `[1, 4, 1, 1]` which would get stripped to `[1, 4,]` as you mentioned. However, I'm not confident I've seen the only variety of this issue and this fix would strip down any mask of length greater than 2 with trailing 1's back down to size 2. A good way to think of removing 1's is as analogous to the `np.squeeze()` function. Simply removing redundant dimensions

dbogunowicz:
@KSGulin got ya. Please resolve the our conversation :)

dbogunowicz (dismissed):

bfineran (approved):

dbogunowicz (approved):
okayoo

PR #682, opened 2022-04-07
[Fix] `composed_staged` failing with modifiers with dynamically inferred `end_epoch`

dbogunowicz:
Before:
when incrementing the `end_epoch` value in L213 `optim/manager.py`, the modifiers with attribute `end_epoch=-1` either verbosely or silently failed. 
Now:
this PR allows:
- for modifiers in the last stage of the recipe:  to preserve their implicit (dynamic?) `end_epoch==-1` attribution
- for modifiers not in the last stage of the recipe: to swap `end_epoch=-1` for the largest `end_epoch`, so that for every non-last stage of the recipe each modifier has a clear, fully laid out set of attributes.


tests/sparseml/pytorch/optim/test_manager.py
```diff
@@ -535,3 +535,100 @@ def test_manager_staged_recipe_serialization(staged_recipe):
...
+
+
+@pytest.mark.parametrize(
+    "base_recipe, additional_recipe, expected_composed_recipe",
+    [
+        # Testing modifiers with dynamic `end_epoch` attribute
+        (SIMPLE_RECIPE, SIMPLE_RECIPE, COMPOSED_RECIPE),
+    ],
+)
+def test_composed_staged(base_recipe, additional_recipe, expected_composed_recipe):
```
bfineran, line 630: is there not an existing test that already does this?

src/sparseml/optim/manager.py
```diff
@@ -207,9 +207,28 @@ def compose_staged(
...
+            # make sure that for the modifiers in base_stages
+            # with the initial attribute `end_epoch` = -1,
+            # this attribute value is replaced with `base_end_epoch`
+            for base_modifiers in base_stages.values():
+                for base_modifier in base_modifiers:
+                    if (
+                        hasattr(base_modifier, "end_epoch")
+                        and base_modifier.end_epoch == -1
+                    ):
+                        base_modifier._init_end = base_end_epoch
```
bfineran, line 220: where is _init_end used?

tests/sparseml/pytorch/optim/test_manager.py
dbogunowicz: True, there is one, but it does not use those "problematic modifiers". I will merge those tests together.

src/sparseml/optim/manager.py
dbogunowicz: Without _init_end modification, we get ValueError at L707 `optim/modifier.py`:
```
        if self._end_comparator is None and self._end_epoch != self._init_end:
            raise ValueError(
                "end_epoch of {} must be equal the init value of {} for {}".format(
                    self._end_epoch, self._init_end, self.__class__.__name__
                )
            )
```

tests/sparseml/pytorch/optim/test_manager.py
bfineran: yeah, since we're using pytest parameterize we can just add in the test strings and not need to duplicate code

src/sparseml/optim/manager.py
bfineran: looks good, one thing we need to eventually consider is some modifiers we may want to have automatically end at the end of a stage if indefinite (maybe constant pruning), but some should stick around forever such as quantization. I think we're ok for now, but as this gets more used and we hit corner cases and new modifiers this may come up

dbogunowicz: Hmm, this is a super interesting corner case. I like the idea of handling them progressively. At the same time, we may now start documenting them and logging them into the backlog, what do you think?

bfineran (approved):

src/sparseml/optim/manager.py
KSGulin: @bfineran what's the context for the conditional on optim/modifier.py::707? Because it seems it's designed to prevent modifications of the end epoch and I'm wondering if it's an anti pattern to work around the conditional rather than to modify/delete it

dbogunowicz: @KSGulin very good point

bfineran: idea here was if something doesn't have an end comparator then the end epoch likely should be -1 or something else that shouldn't be set which is why this check is in place.

spacemanidol (approved):

PR #684, opened 2022-04-07
Yolov5 docs update

KSGulin:
Documentation update in support of Yolov5 update to V6.1 (neuralmagic/yolov5#26)


bfineran (dismissed):
LGTM @KSGulin will need to update any recipes/model cards in zoomodels as well

anmarques (dismissed):

dbogunowicz (dismissed):
Approve for now, I will also test those docs soon by following the instruction step-by-step.

KSGulin:
@anmarques Good catch Alex. Updated corresponding PRs to use the legacy hyperparameter files

integrations/ultralytics-yolov5/setup_integration.sh
```diff
@@ -6,5 +6,6 @@
 
 git clone https://github.com/neuralmagic/yolov5.git
 cd yolov5
-git checkout release/0.11
+git checkout release/0.12
 pip install -r requirements.txt
+pip install sparseml[torch, torchvision]
```
dbogunowicz, line 11: nit, actually very general observation.
This syntax for `pip install` does not work for mac machines where 
`pip install sparseml'[torch, torchvision]'` is the way to go.
I propose, maybe not now, but in the future, to make have if-else statement (if one command fails ("ubuntu command"), we try to run the alternative("mac command")). 
This would also apply to `setup.py` in SparseML.

dbogunowicz (dismissed):
as discussed in private conversation with @KSGulin , I did run some rudimentary testing on the yolov5 and it works as expected (the base functionalities).

bfineran (dismissed):

integrations/ultralytics-yolov5/setup_integration.sh
```diff
@@ -6,5 +6,5 @@
 
 git clone https://github.com/neuralmagic/yolov5.git
 cd yolov5
-git checkout release/0.11
+git checkout update-6.1
```
bfineran, line 9:
```suggestion
git checkout release/0.12
```

bfineran (approved):

spacemanidol (approved):

PR #685, opened 2022-04-07
Quant refactor testing

bfineran:
testing for #663


anmarques (approved):
Check on quantization limits will fail if TensorRT=True is used. This condition is not checked in the current tests.

PR #686, opened 2022-04-08
[Fix] Reverting back to `torch==1.9.0`

dbogunowicz:
In the last release, before we knew about the incoming problems with onnx export, I updated the latest supported torch version to `1.10.2` (0.11 release). However, we are not yet ready to support it and the sensible thing to do is to rollback, and bring back `1.10.X` in release 0.13.

Kudos to @KSGulin for catching this. To reproduce the issue, one should grab a switch branch to #632  install fresh dependencies from `setup.py`, and run the included transformers integration tests. 

They will fail, directly because the transformers `export` method fails if the torch version is higher than `1.9.X`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,12 +56,12 @@ _deepsparse_deps = [
     f"{'deepsparse' if is_release else 'deepsparse-nightly'}~={version_nm_deps}"
 ]
 _pytorch_deps = [
-    "torch>=1.1.0,<=1.9.0",
+    "torch>=1.1.0,<=1.9.1",
     "tensorboard>=1.0",
     "tensorboardX>=1.0",
     "gputils",
 ]
-_pytorch_vision_deps = _pytorch_deps + ["torchvision>=0.3.0,<=0.10.0"]
+_pytorch_vision_deps = _pytorch_deps + ["torchvision>=0.3.0,<=0.10.1"]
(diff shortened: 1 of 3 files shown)


KSGulin (dismissed):

bfineran (commented):
Let's bump the minor version of torch in our setup.ph from 1.9.0 to the latest 1.9 release  minor release (ie 1.9.1)

setup.py
```diff
@@ -56,12 +56,12 @@
     f"{'deepsparse' if is_release else 'deepsparse-nightly'}~={version_nm_deps}"
 ]
 _pytorch_deps = [
-    "torch>=1.1.0,<=1.10.2",
+    "torch>=1.1.0,<=1.9.0",
```
dbogunowicz, line 59:
```suggestion
    "torch>=1.1.0,<=1.9.1",
```

setup.py
```diff
@@ -56,12 +56,12 @@
...
     "tensorboard>=1.0",
     "tensorboardX>=1.0",
     "gputils",
 ]
-_pytorch_vision_deps = _pytorch_deps + ["torchvision>=0.3.0,<=0.11.3"]
+_pytorch_vision_deps = _pytorch_deps + ["torchvision>=0.3.0,<=0.10.0"]
```
dbogunowicz, line 64:
```suggestion
_pytorch_vision_deps = _pytorch_deps + ["torchvision>=0.3.0,<=0.10.1"]
```

KSGulin (approved):

bfineran (approved):

PR #690, opened 2022-04-08
add pass to thin from checkpoint while loading torch models

bfineran:
#678 adds support to ensure consistency between thinned models when reloading weights with `applied_structure`, however for extra consistency and backwards compatibility with models before this fix, this PR adds a pass to thin a model to match the layer shapes in a checkpoint to avoid size mismatch errors on torch load.

Testing is included for toy conv and FC models


markurtz (dismissed):

rahul-tuli (dismissed):

src/sparseml/pytorch/utils/helpers.py
```diff
@@ -957,3 +962,99 @@ def torch_distributed_zero_first(local_rank: int):
     yield
     if local_rank == 0:
         torch.distributed.barrier()
+
+
+def thin_model_from_checkpoint(model: Module, state_dict: Dict[str, Any]):
+    """
+    Updates any Linear/Conv/BN layers in the given model to match their
+    respective shapes in the given state dict. Purpose of for compatibility
```
rahul-tuli, line 970: nit: 
```suggestion
    respective shapes in the given state dict. Purpose of compatibility
```

spacemanidol (approved):

rahul-tuli (approved):

PR #702, opened 2022-04-12
Add: Functionality to resume-training from a previously stopped run.

rahul-tuli:
This PR will add support for resuming training from a previously stopped run; the epoch is read from the checkpoint if present. However, the recipe must be provided explicitly.

Add `--resume` argument, and provide a checkpoint with epoch #; 

For example:
```bash
--recipe-path zoo:cv/classification/resnet_v1-50/pytorch/sparseml/imagenette/pruned-conservative?recipe_type=original \
--checkpoint-path ~/models/custom-checkpoint/fake-epoch-2.pth \
--pretrained True --dataset imagenette --dataset-path dataset \
--train-batch-size 128 --test-batch-size 256 --loader-num-workers 8 \
--model-tag resnet50-imagenette-pruned-conservative --resume
```

The checkpoint must contain epoch # when training was stopped, else it is assumed to be epoch 0
```python
{
     "state_dict" : ...,
     "epoch" : 2,
      .
      .
}
```

### DIFF
--- a/src/sparseml/pytorch/models/registry.py
+++ b/src/sparseml/pytorch/models/registry.py
@@ -76,8 +76,9 @@ class ModelRegistry(object):
         pretrained_dataset: str = None,
         load_strict: bool = True,
         ignore_error_tensors: List[str] = None,
+        return_extras: bool = False,
         **kwargs,
-    ) -> Union[Module, Tuple[Module, Optional[str]]]:
+    ) -> Union[Module, Tuple[Module, Optional[str], Optional[int]]]:
         """
         Create a new model for the given key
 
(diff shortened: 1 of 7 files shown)


src/sparseml/pytorch/models/registry.py
```diff
@@ -91,19 +92,28 @@ def create(
...
-            if "arch_key" in _checkpoint:
-                key_copy = _checkpoint["arch_key"]
+            if "arch_key" in checkpoint:
+                key_copy = checkpoint["arch_key"]
             else:
                 raise ValueError("No `arch_key` found in checkpoint")
 
+        extras.append(key or key_copy)
+
+        if "epoch" in checkpoint:
```
dbogunowicz, line 113: Maybe adding an error or warning if the epoch is not in checkpoint? By default, we should be saving epoch inside the checkpoint right?

dbogunowicz: Same applies to "arch_key" btw

dbogunowicz:
@rahul-tuli @bfineran correct me if I am wrong, but shouldn't we be making sure that when resuming, not only do we extract the correct epoch from the checkpoint, but also the correct state of the optimizer (e.g. learning rate, scheduler state), as well as applying checkpoint weights?

PR #703, opened 2022-04-12
Switch off fp16 on QAT start

rahul-tuli:
This PR updates `ImageClassificationTrainer` to switch off `fp16` on QAT start 

Noting that `torch > 1.9`. does not crash if mixed precision is not switched off before QAT start thus the additional check

### DIFF
--- a/src/sparseml/pytorch/image_classification/utils/helpers.py
+++ b/src/sparseml/pytorch/image_classification/utils/helpers.py
@@ -370,13 +370,14 @@ def save_model_training(
     :param arch_key: if provided, the `arch_key` will be saved in the
         checkpoint
     """
-    has_top1 = "top1acc" in val_res.results
-    metric_name = "top-1 accuracy" if has_top1 else "val_loss"
-    metric = val_res.result_mean("top1acc" if has_top1 else DEFAULT_LOSS_KEY).item()
-    print(
-        f"Saving model for epoch {epoch} and {metric_name} "
-        f"{metric} to {save_dir} for {save_name}"
-    )
+    if val_res is not None:
+        has_top1 = "top1acc" in val_res.results
+        metric_name = "top-1 accuracy" if has_top1 else "val_loss"
+        metric = val_res.result_mean("top1acc" if has_top1 else DEFAULT_LOSS_KEY).item()
+        print(
+            f"Saving model for epoch {epoch} and {metric_name} "
+            f"{metric} to {save_dir} for {save_name}"
+        )
     exporter = ModuleExporter(model, save_dir)
     exporter.export_pytorch(
         optim,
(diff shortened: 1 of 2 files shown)


src/sparseml/pytorch/image_classification/utils/trainer.py
```diff
@@ -124,7 +124,9 @@ def __init__(
 
         self.optim_name = optim_name
         self.epoch = 0
-
+        self.device_context = ModuleDeviceContext(
```
bfineran, line 127: nit: rename to `self._device_context`

src/sparseml/pytorch/image_classification/utils/helpers.py
```diff
@@ -370,13 +370,14 @@ def save_model_training(
...
         checkpoint
     """
-    has_top1 = "top1acc" in val_res.results
-    metric_name = "top-1 accuracy" if has_top1 else "val_loss"
-    metric = val_res.result_mean("top1acc" if has_top1 else DEFAULT_LOSS_KEY).item()
-    print(
-        f"Saving model for epoch {epoch} and {metric_name} "
-        f"{metric} to {save_dir} for {save_name}"
-    )
+    if val_res is not None:
```
bfineran, line 373: when does this come up?

rahul-tuli: The val_res are typed as `Optional[ModuleRunResults]` hence the check enforces certain operations to be performed only when `val_res` is `non-None`. 

Noting: That this is a very rare situation. For me the situation arose with a somewhat faulty recipe when a quantization modifier was present without any actual training, i.e
```yaml
quantization_start_epoch: 0

quantization_modifiers:
  - !QuantizationModifier
    start_epoch: eval(quantization_start_epoch)
```

Omitting this check for the specified recipe will throw an `AttributeError: 'NoneType' object has no attribute 'results'` Which I feel is the wrong error message for this situation. With this check in place the program exits elegantly with no training performed. Any thoughts/suggestions are welcome

bfineran (approved):

src/sparseml/pytorch/image_classification/utils/trainer.py
```diff
@@ -170,6 +172,10 @@ def run_one_epoch(
         train_mode = mode == "train"
         validation_mode = not train_mode
 
+        if torch.__version__ < "1.9" and self.manager.qat_active(epoch=self.epoch):
```
KSGulin, line 175: > Noting that torch > 1.9. does not crash if mixed precision is not switched off before QAT start
Does torch == 1.9.x crash? Because it would pass above this check

KSGulin (commented):
LGTM! Just one small comment/question

src/sparseml/pytorch/image_classification/utils/trainer.py
rahul-tuli: It wouldn't pass the above check I confirmed:
```python
In [1]: "1.9.1" < "1.9"
Out[1]: False
```

Anything >1.9 dos not need mixed precision to be turned off so it's good

KSGulin (approved):

PR #709, opened 2022-04-15
Scripts for bio tasks

spacemanidol:


src/sparseml/transformers/text_classification.py
```diff
@@ -592,14 +608,17 @@ def preprocess_function(examples):
...
-    # Log a few random samples from the training set:
-    if training_args.do_train:
-        for index in random.sample(range(len(train_dataset)), 3):
-            _LOGGER.info(f"Sample {index} of training set: {train_dataset[index]}.")
-
+    
+    
     # Get the metric function
-    if data_args.task_name is not None:
+    if data_args.task_name is not None and data_args.task_name not in ["biosses","pubmedqa","chemprot","DDI","GAD","hoc","bioasq"]:
```
bfineran, line 614: nit: any reason not to move the other `elif`s up and keep `elif data_args.task_name is not None` as the last/default elif so we don't need to maintain this list?

src/sparseml/transformers/text_classification.py
```diff
@@ -630,6 +649,11 @@ def compute_metrics(p: EvalPrediction):
     else:
         data_collator = None
 
+    # Log a few random samples from the training set:
+    if training_args.do_train:
+        for index in random.sample(range(len(train_dataset)), 3):
```
bfineran, line 654: what's the reason for this?

bfineran (commented):
looks good to me, assumed these tasks have been tested/verified on your end @spacemanidol ?

src/sparseml/transformers/text_classification.py
dbogunowicz: sanity check to make sure that we are feeding the correct data structure into the model? if my assumption is correct, it would be probably more useful for the user to see something like 
` _LOGGER.info(f"Sample {index} of training set: {train_dataset[index].size()}.")`
rather than full tensor print.

PR #710, opened 2022-04-17
Match inputs w/ forward (corner case)

natuan:
Need to filter inputs based on forward also e.g. when manager is not available.


src/sparseml/transformers/sparsification/trainer.py
```diff
@@ -340,6 +340,9 @@ def compute_loss(
             or not self.manager.enabled
             or not self.manager.distillation_modifiers
         ):
+            inputs = {
```
dbogunowicz, line 343: Looks good to me. @natuan I am just curious, when this corner case may occur in practice?

bfineran (approved):

KSGulin (approved):

PR #718, opened 2022-04-18
Delete testing log files and check for file creation

KSGulin:
Currently ~130 files are being created when running the PyTorch test suite. wandb and tensorboard logs account for all of these files. This PR adds an environment variable to indicate testing mode, which will route logs to a temporary test logging directory and clean it up afterwards. It also adds a test to check for whether files were created within the run directory (and its subdirectories) during test run.

Potential issues - The test checks whether files were created in the directory (and subdirectories) during the timespan of testing, not whether the tests created the files. This shouldn't pose an issue for GHA, but devs will need to keep that in mind while testing locally.


tests/sparseml/conftest.py
```diff
@@ -0,0 +1,28 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
+import pytest
+
+
+@pytest.fixture(scope="session", autouse=True)
```
dbogunowicz, line 20: Sweet!

bfineran (commented):
lgtm, this won't account for files opened/created outside of `.`, but good check to have

dbogunowicz (commented):
I am super happy with this fix, great detective job! 

One thing which keeps me from giving it a green light is that I was not able to see it working properly when launching my `make test`.

Could you paste  your desired command line output (e.g. screenshot from the terminal)? Maybe I am missing something.

KSGulin:
@dbogunowicz If there are no additional files created, you won't see anything in the output terminal. This is because the assertion is in a fixture and is not its own test. But, when the assertion does fail, pytest generates an `ERROR` line and the test still fails

dbogunowicz:
> @dbogunowicz If there are no additional files created, you won't see anything in the output terminal. This is because the assertion is in a fixture and is not its own test. But, when the assertion does fail, pytest generates an `ERROR` line and the run is considered a failure

Got ya. But, still, I am not satisfied with the current state. The assertion checks if the test session creates additional files. Only if it doesn't (or in fact removes files, both scenarios are rather improbable) it violates the assertion and raises AssertionError. 

I'm failing to see how this logic is practical. What makes sense to me is e.g. we set a limit of how many files can be created per session (we have a rough idea now), and if this gets exceeded, we throw an error. Does this make sense?

bfineran (commented):
@KSGulin From what I understood, I thought there was a problem with the maximum number of files open at peak during testing. if we do not clean up to the end is this problem addressed?

KSGulin:
@dbogunowicz I believe the idea is that we'd like to keep to the standard of not creating any new files during testing. If that's not in fact the standard that we want to uphold, then I'd agree. 

Also, @bfineran makes a great point. The results show that the file usage was significantly decreased during testing runtime, but files doesn't get deleted until the very end of the session. I'm not sure why that's the case, but I'll update this patch to delete files after each function

dbogunowicz:
> @dbogunowicz I believe the idea is that we'd like to keep to the standard of not creating any new files during testing. If that's not in fact the standard that we want to uphold, then I'd agree.
> 
> Also, @bfineran makes a great point. The results show that the file usage was significantly decreased during testing runtime, but files doesn't get deleted until the very end of the session. I'm not sure why that's the case, but I'll update this patch to delete files after each function

Ok, this sounds very good to me.

bfineran (approved):
pushing to land this now that test files opened error seems mitigated

dbogunowicz (approved):

PR #720, opened 2022-04-19
[WIP] SparseML Dockerfile and build processes

dbogunowicz:
Related to: https://app.asana.com/0/1201735099598270/1201964439431277/f


KSGulin (commented):
Looks like it's almost there, just needs a couple fixes.

The dockerfile builds and runs fine, but support for pytorch cuda operations isn't included. There's two roots of the cause:
1. Pytorch hasn't compiled a 1.9.1 version that supports cuda 11.3. The closest combo that's available via pip installable is torch 1.9.1 + cuda 11.1
2. For older version of pytorch, the cuda version needs to be specified on install. e.g. `pip install torch==1.9.1+cu111 torchvision==0.10.1+cu111 --extra-index-url https://download.pytorch.org/whl/cu111`

docker/README.md
```diff
@@ -1 +1,8 @@
-[TODO: fill in docs for docker examples and how to run]
\ No newline at end of file
+## SparseML docker image
+This directory contains the Dockerfile to create minimal SparseML docker image.
+This image is based off the official NVIDIA development Ubuntu 18.04.5 LTS image (with CUDA support).
+
+In order to build and launch this image, run from the root directory:
+- for compute platform CUDA 10.2: `docker build --build-arg CUDA_VERSION=10.2 -t sparseml_docker . && docker run -it --gpu all sparseml_docker`
```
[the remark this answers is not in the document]
dbogunowicz: Good catch.

dbogunowicz:
> Looks like it's almost there, just needs a couple fixes.
> 
> The dockerfile builds and runs fine, but support for pytorch cuda operations isn't included. There's two roots of the cause:
> 
> 1. Pytorch hasn't compiled a 1.9.1 version that supports cuda 11.3. The closest combo that's available via pip installable is torch 1.9.1 + cuda 11.1
> 2. For older version of pytorch, the cuda version needs to be specified on install. e.g. `pip install torch==1.9.1+cu111 torchvision==0.10.1+cu111 --extra-index-url https://download.pytorch.org/whl/cu111`

Interesting @KSGulin . Now I see why your dockerfile in containers repo was quite complex.
So I did look into it and it seems like this is the right way to solve the problem: 

1. Scenario when the dockerfile installs Cuda 11.3:
You're suggesting to actually swap Cuda 11.3 for Cuda 11.1, because this is the latest cuda version that supports `torch==1.9.1`. Do you have a source on that?
Once Cuda 11.1 is installed:
`pip install torch==1.9.1+cu110 torchvision==0.10.0+cu110 -f https://download.pytorch.org/whl/torch_stable.html`
2. Scenario when using Cuda 10.2:
The internet suggests that it should work out of the box simply `pip install torch==1.9.1 torchvision==0.10.0`

Thanks for making me aware of those issues, this is really interesting. If I still did not get this right, let's have a chat today.

docker/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+
+FROM base as cuda-10.2
+ARG TORCH_VERSION
+ARG TORCHVISION_VERSION
+RUN pip3 install --upgrade pip torch==$TORCH_VERSION torchvision==$TORCHVISION_VERSION
+
+FROM base as cuda-11.3.1
+ARG TORCH_VERSION
+ARG TORCHVISION_VERSION
+RUN pip3 install --upgrade pip torch==$TORCH_VERSION torchvision==$TORCHVISION_VERSION --extra-index-url https://download.pytorch.org/whl/cu113
```
bfineran, line 34: do we need to append +cu111 to the default torch version here to get the right one downloaded?

KSGulin:
@dbogunowicz this [page](https://pytorch.org/get-started/previous-versions/) has all the install commands for legacy versions of pytorch. For 1.9.1 installed using wheel on linux, looks like explicit cuda version callout is needed in both

dbogunowicz:
> @dbogunowicz this [page](https://pytorch.org/get-started/previous-versions/) has all the install commands for legacy versions of pytorch. For 1.9.1 installed using wheel on linux, looks like explicit cuda version callout is needed in both

This is great, thank you. Updated the PR.

bfineran (commented):
LGTM overall, please add notes on testing completed in PR description

docker/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+FROM base as cuda-11.1
+ARG TORCH_VERSION
+ARG TORCHVISION_VERSION
+RUN pip3 install --upgrade pip torch==${TORCH_VERSION}+cu111 torchvision==${TORCHVISION_VERSION}+cu111 -f https://download.pytorch.org/whl/torch_stable.html
+
+FROM cuda-$CUDA_VERSION as target
+
+# Install SparseML
+RUN pip3 install sparseml
```
rahul-tuli, line 38: will adding `--no-cache-dir` be good practice here

rahul-tuli, line 38: Maybe something like the following
```suggestion
RUN pip3 install --no-cache-dir --upgrade sparseml
```

dbogunowicz: Interesting.
```
I think there is a good reason to use --no-cache-dir when you are building Docker images. The cache is usually useless in a Docker image, and you can definitely shrink the image size by disabling the cache.
```

dbogunowicz:
Tested using command
```
sparseml.image_classification.train --recipe-path zoo:cv/classification/resnet_v1-50/pytorch/sparseml/imagenet/pruned85_quant-none-vnni  --arch-key resnet50  --pretrained pruned-moderate     --dataset imagenette --dataset-path dataset     --train-batch-size 4 --test-batch-size 8
```
image

docker/README.md
```diff
@@ -1 +1,7 @@
-[TODO: fill in docs for docker examples and how to run]
\ No newline at end of file
+## SparseML docker image
+This directory contains the Dockerfile to create minimal SparseML docker image.
+This image is based off the official NVIDIA development Ubuntu 18.04.5 LTS image (with CUDA support).
+
+In order to build and launch this image, run from the root directory:
+- for compute platform CUDA 10.2: `docker build --build-arg CUDA_VERSION=10.2 -t sparseml_docker . && docker run -it --gpus all sparseml_docker`
+- for compute platform CUDA 11.1: `docker build --build-arg CUDA_VERSION=11.1 -t sparseml_docker . && docker run -it --gpus all sparseml_docker`
```
bfineran, line 7: let's break down this example usage section into `build` which is separate for either cuda version and then `run` which will be the same and we can provide example commands to run / how to get into an interactive shell

bfineran (commented):
LGTM, please add description of systems this container was tested on and for which sparseml use cases, also left a suggestion for the README

rahul-tuli (dismissed):

docker/README.md
dbogunowicz: Ok, the import problems are fixed now.

But the nvidia images are still faulty, so I am keeping the hack that I mentioned inplace.

KSGulin (approved):
Great work on this!

bfineran (approved):

PR #722, opened 2022-04-20
N:M pruning mask creator

KSGulin:
Implements the N:M sparsity pruning mask creator, as is done [here.](https://github.com/NM-sparsity/NM-sparsity)

Some maneuvering was needed to make the mask creator work with the existing sparsity structure. i.e. sparsity is a required field and the provided sparsity must align with the effective sparsity within `1e-2`. As a result, recipes using N:M pruning must specify both the N:M ratio and a sparsity that matches the ratio.

An example use of the N:M mask:

```
- !GMPruningModifier
    params: __ALL_PRUNABLE__
    start_epoch: 0
    end_epoch: 0.5
    init_sparsity: 0.0
    final_sparsity: 0.75
    update_frequency: 1.0
    mask_type: "1:4"
```
Note that an N:M mask is specified by a string of the form "N:M"


src/sparseml/pytorch/sparsification/pruning/mask_creator.py
```diff
@@ -583,17 +584,116 @@ def _get_blocked_tens_shape_and_validate(
...
+                w_b = torch.ones(tensor_temp.shape, device=tensor_temp.device)
+                masks.append(
+                    w_b.scatter_(dim=1, index=index, value=0).reshape(
+                        original_tensor.permute(0, 2, 3, 1).shape
+                    )
+                )
+            elif len(tensor.shape) == 2:
+                # N:M sparsity for linear layers
+                tensor_temp = tensor.detach().abs().reshape(num_groups, self._M)
+                index = torch.argsort(tensor_temp, dim=1)[:, : int(self._M - self._N)]
```
dbogunowicz, line 667: If you see fit, perhaps it would make sense to make the code a bit compact here (e.g. L667 and L657 are duplicates). If you think that it would hurt readability, you may disregard this comment.

dbogunowicz (commented):
lgtm, good job, very readable implementation.

src/sparseml/pytorch/sparsification/pruning/mask_creator.py
KSGulin: I think it's a good callout, but I'll opt for keeping it as is. Since `index` depends on `tensor_temp` and those are different between the two versions.

spacemanidol (dismissed):

tests/sparseml/pytorch/sparsification/pruning/test_mask_creator.py
```diff
@@ -161,3 +162,35 @@ def test_four_block_mask_creator_matches_block(tensors, sparsity_val):
     for mask_1, mask_2 in zip(masks_1, masks_2):
         assert mask_1.shape == mask_2.shape
         assert torch.all(mask_1 == mask_2)
+
```
bfineran, line 165: do we have any way here of testing that the created masks comply to the N:M spec?

src/sparseml/pytorch/sparsification/pruning/mask_creator.py
```diff
@@ -583,17 +584,116 @@ def _get_blocked_tens_shape_and_validate(
...
+        unstructured mask creator, 'block4' for 1x4 block pruning, 'N:M' where N and M
+        are integers for N:M pruning, or a list of two integers for custom block
+        pruning (does not support padding)
     :return: mask creator object created from the mask type
     """
     if mask_type == "unstructured":
         return UnstructuredPruningMaskCreator()
     elif mask_type == "block4":
         return FourBlockMaskCreator()
+    elif ":" in mask_type:
```
bfineran, line 689: let's also add an alias of `tensorrt_2_4` or maybe just `tensorrt`

bfineran (changes requested):
quick suggestion and question then LGTM. Great job

KSGulin:
Thanks for the suggestions @bfineran! Both have been implemented

bfineran (approved):
looks great @KSGulin

dbogunowicz (approved):

Issue #704, opened 2022-04-13
onnxruntime-gpu - is there a reson this is not an option?


meeki007:
**Is your feature request related to a problem? Please describe.**
built the onnxruntime-gpu to fulfill requirement
```
 Could not find a version that satisfies the requirement onnxruntime>=1.0.0 (from sparseml[torch,torchvision]) (from versions: )
No matching distribution found for onnxruntime>=1.0.0 (from sparseml[torch,torchvision])
```
**Describe the solution you'd like**
can you make an XOR for onnxruntime >=1.0.0 or onnxruntime-gpu>=1.0.0 in [sparseml](https://github.com/neuralmagic/sparseml)/setup.py

**Additional context**
should work. git cloning https://github.com/neuralmagic/sparseml.git and going to try and see if it still works after changing line in setup.py to onnxruntime-gpu>=1.0.0
It should as 
```
$ python3 -c 'import onnxruntime as rt; print(rt.__version__)'
```
gives me
```
1.12.0
```
using onnxruntime-gpu I built.

markurtz:
Hi @meeki007, thanks for bringing this issue up. We'll dive in and get a solution up soon. In the meantime, let us know if you have any issues with editing the setup.py.

meeki007:
no issues worked fine using: onnxruntime-gpu>=1.0.0 in setup.py
I'm a python nublit but getting better. I lack the skillz to create a PR for this. Give me a few more months in python and I'll be able to help with this simple if than checks stuff.

Oh I did notice that when I installed sparsml:
```
sudo -H python3 -m pip install -e ./[torch,torchvision]
```
later when I tried `SparseML Ultralytics YOLOv5 Integration`
I found that it filed checks on the github.com/neuralmagic/yolov5/tree/release/0.12)/requirements.txt
```
sparseml[torch,torchvision]>=0.12
```
as it is looking for a package installed via python and not by dev build
as I could run sparseml in python and it version checked fine I just removed the requirement along with the requirement for 
```
opencv-python
```
as I have a full cuda enabled version of openCV built, make, and installed
as of right now sparseml is working. going to let it run tonight sparse-ing away

Great work on this. Sorry I'm a JS/BASH/node guy. I look forward to contributing as i can with python as I learn syntax of it.

dbogunowicz:
Hey @meeki007. Thank you for your active support!
I am working on the issue, the solution is coming soon.

dbogunowicz:
[...]

PR #723, opened 2022-04-21
[Fix] Onnxruntime-gpu install not compatible

dbogunowicz:
Related to issue: https://github.com/neuralmagic/sparseml/issues/704


dbogunowicz:
Converting back to draft, found that this silently breaks onnx and transformer tests. Working on a solution.

bfineran (approved):
LGTM, any reason this was moved to a draft?

dbogunowicz:
> LGTM, any reason this was moved to a draft?

@bfineran yeah, I did not get a chance to run some manual sanity check if this really works.
Just tested by creating a fresh venv, installing only `onnxruntime-gpu` and then running `python setup.py install`.

image

Looking good.

bfineran (changes requested):
blocking this for now since as discussed offline this will not address the issue for a prebuilt wheel

dbogunowicz:
Closing, replaced by #789.

PR #725, opened 2022-04-21
re-add fix symmetric zero points for unit8 quantization (#604)

bfineran:
changes from #604 were lost at some point, likely in a merge resolution


markurtz (approved):

anmarques (approved):
Produced a model with the fix and it worked as intended!

PR #727, opened 2022-04-21
Transformers tests update with legacy recipe support

KSGulin:
This PR updates transformers tests in the following ways:

- Copy `original.md` recipe to `recipe.yaml` so that it's automatically ingestible by the the trainer
- Add quantization test case, which requires the recipe support above


src/sparseml/transformers/export.py
```diff
@@ -157,6 +157,7 @@ def export_transformer_to_onnx(
         model_path, model_max_length=sequence_length
     )
     model = load_task_model(task, model_path, config)
+
```
dbogunowicz, line 160:
```suggestion
```

tests/sparseml/transformers/test_transformers.py
```diff
@@ -80,14 +86,33 @@ def _compare_onnx_models(model_1, model_2):
         assert nodes1_count[node] == nodes2_count[node]
 
 
+def _attempt_copy_original_recipe(model_path):
```
dbogunowicz, line 89: I understand the purpose of this function, but also would like to stress the "temporariness" of this solution -> should be removed once we move to `recipe.yaml` by default everywhere (if needed) in our framework files.
```suggestion
# TODO: Remove this function once yaml recipes supplied
def _attempt_copy_original_recipe(model_path):
```

tests/sparseml/transformers/test_transformers.py
```diff
@@ -108,14 +133,15 @@ def setup(self, model_stub, recipe_present, task):
     def test_load_weights_apply_recipe(self, setup):
         model, recipe_present, task = setup
         model_path = model.framework_files[0].dir_path
+        _attempt_copy_original_recipe(model_path)
```
dbogunowicz, line 136: I'd rather execute this function on the setup. Should be more foolproof once we keep adding more tests / decide to run tests selectively.

tests/sparseml/transformers/test_transformers.py
```diff
@@ -131,6 +157,7 @@ def test_export_to_onnx(self, setup):
         model, recipe_present, task = setup
         path_onnx = model.onnx_file.downloaded_path()
         model_path = model.framework_files[0].dir_path
+        _attempt_copy_original_recipe(model_path)
```
dbogunowicz, line 160: See comment above about having this function on the setup

tests/sparseml/transformers/test_transformers.py
```diff
@@ -152,13 +179,15 @@ def test_outputs_ort(self, setup):
         path_onnx = model.onnx_file.downloaded_path()
         model_path = model.framework_files[0].dir_path
         inputs_path = model.data_inputs.path
+        _attempt_copy_original_recipe(model_path)
```
dbogunowicz, line 182: See comment above about having this function on the setup

tests/sparseml/transformers/test_transformers.py
```diff
@@ -152,13 +179,15 @@ def test_outputs_ort(self, setup):
...
 
         input_data = load_numpy_list(inputs_path)[0]
 
         path_retrieved_onnx = export_transformer_to_onnx(
             task=task,
             model_path=model_path,
             onnx_file_name=self.onnx_retrieved_name,
+            sequence_length=next(iter(input_data.values())).shape[0],
```
dbogunowicz, line 190: Oh shoot, it looks like you were not able to find the `sequence_length` in the attributes of the model.
Tbh, I really do not like the fact that one needs to go such an extra mile to retrieve it.
@bfineran how do you feel about adding `sequence_length` as an attribute of a transformer model (downloaded from stub)?

dbogunowicz (commented):
added few comments, happy to see those tests expand :)

btw, really like how smooth this review was, due to having  context from our  yesterday's session.

tests/sparseml/transformers/test_transformers.py
KSGulin: From what I can tell, you need to load in the transformer model to get that attribute. The `model` shown here is a SpraseZoo model object, rather than a transformers model object. We probably don't want to add a `sequence_length` property to that class, but what might be a reasonable solution is to implement a function to determine sequence length within the transformers `export.py`, once the actual model is loaded.

KSGulin:
Great points @dbogunowicz. Comments addressed and ready for re-review.

tests/sparseml/transformers/test_transformers.py
bfineran: This is fine for now, so we can compare outputs potentially. Sequence length can be changed on compilation so is not necessarily tied to a particular model.

bfineran (approved):
LGTM, logic for copying over zoo recipe to framework files will be useful until zoomodels are updated to latest standards

dbogunowicz (approved):
let it snow

PR #729, opened 2022-04-22
Implement mask-transfer between models at init

eldarkurtic:
* masks are transferred from the given model at initialization
* to keep masks on, one should add ConstantPruningModifier in the yaml
recipe
* add apply_weight_mask function to transfer masks between models
* add command-line argument --apply_mask_from to all Transformers examples
* useful for: mask-only transfer experiments (for example lottery-ticket style)


natuan (approved):
Thanks for the feature. I think it'll be helpful.

src/sparseml/transformers/utils/model.py
```diff
@@ -137,6 +147,10 @@ def question_answering_from_pretrained(
             model_name_or_path,
             **kwargs,
         )
+        if mask_name_or_path:
+            mask = AutoModelForQuestionAnswering.from_pretrained(mask_name_or_path)
```
bfineran, line 151: let's add an info log here, at minimum should say we are performing the mask transfer and give the model path, should also consider listing applied sparsities

bfineran (commented):
looks like a cool feature @eldarkurtic. lgtm pending comment about logging

eldarkurtic:
@bfineran I have pulled out transfer of masks to a staticmethod of SparseAutoModel class, to avoid repeating a bunch of code for each task. Now, one could enable this feature by adding:
```python
        if mask_name_or_path:
            SparseAutoModel._transfer_masks(
                mask_name_or_path,
                model_name_or_path,
                model,
                delayed,
            )
```` 
Logging info is also included in it.

bfineran (dismissed):

markurtz (commented):
@eldarkurtic I like the feature and implementation, I'm a bit worried about the potential for small requests like this to turn into a lot of additional arguments in the future for these scripts. Given that we're applying the mask and changing the model in place and that logic should be the same across everything that uses it, do you see any issue with migrating this to a CLI utility script? That way we limit number of changes across the integration along with testing

eldarkurtic:
@markurtz makes sense, given that this feature would do the same thing for all models. What do you mean by CLI utility script? 
The only potential issue I can see with generalizing this feature is model loading, given that they can have unique loading procedures and should probably be implemented on a case-by-case basis.

jeanniefinks:
Product being deprecated by early June 2, 2025. Closing PR and will not resolve.

PR #730, opened 2022-04-25
Added versions of AC/DC with different pruning criteria - Magnitude/G…

Godofnothing:
The original M-FAC paper https://arxiv.org/abs/2107.03356 and the current implementation in sparseml has only AC/DC implementation with the layer-wise MagnitudePruning modifier. 

However, the formulation of AC/DC allows to use any pruning scorer - GlobalMagnitude/M-FAC/Movement on the compression step. I've created implementation that allows to use different pruning scorers:

- `ACDCMagnitudePruningModifier`
- `ACDCGlobalMagnitudePruningModifier`
- `ACDCMovementPruningModifier`
- `ACDCMFACPruningModifier`

There is a flaw, that in my implementation in order to work with constructors in subclasses,  I had to add `**kwargs` in the constructors of the base modifiers : `MagnitudePruningModifier, GlobalMagnitudePruningModifier, MovementPruningModifier, MFACPruningModifier`.

### DIFF
--- a/src/sparseml/pytorch/sparsification/pruning/modifier_pruning_acdc.py
+++ b/src/sparseml/pytorch/sparsification/pruning/modifier_pruning_acdc.py
@@ -13,32 +13,40 @@
 # limitations under the License.
 
 import math
-from typing import List, Union
+
+from typing import List, Dict, Union, Optional
 
 from torch.nn import Module, Parameter
 from torch.optim.optimizer import Optimizer
 
-from sparseml.pytorch.sparsification.modifier import ModifierProp, PyTorchModifierYAML
+from sparseml.pytorch.optim.modifier import ModifierProp, PyTorchModifierYAML
 from sparseml.pytorch.sparsification.pruning.mask_creator import (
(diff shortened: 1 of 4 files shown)


KSGulin (commented):
@Godofnothing this is a great catch and suggestion

The only high level comment I have is that it may cleaner to add a serializable `pruning_scorer` keyword to the ACDC modifier and use dynamic inheritance on `ACDCPruningModifier` init to assign the second parent class. That way we can avoid the extra class definitions and maintain one modifier title for ACDC.  Thoughts @bfineran @dbogunowicz?

Thanks for the contribution!

dbogunowicz:
Thank you @Godofnothing. We really appreciate the time and effort you took with our framework.
I will look into the PR later in the day. I just want to leave here a mental note, that we should also add tests to this PRs before we decide to merge.

Yes, @KSGulin , I agree with you on the serializable keyword and inheritance suggestion.

Godofnothing (commented):
Agree with the suggestions

PR #737, opened 2022-04-28
[Fix] Follow-up fix for #731 (Fix 'self' and 'disable' not working for transformers distillation)

dbogunowicz:
731 was not entirely fixing the problem.

### DIFF
--- a/src/sparseml/transformers/sparsification/question_answering.py
+++ b/src/sparseml/transformers/sparsification/question_answering.py
@@ -210,7 +210,11 @@ class QuestionAnsweringTrainer(TrainerInterface, _QuestionAnsweringTrainer):
     def _remove_unused_columns(
         self, dataset: "datasets.Dataset", description: Optional[str] = None
     ):
-        if self._signature_columns is None and self.teacher is not None and self.teacher not in ("disable", "self"):
+        if (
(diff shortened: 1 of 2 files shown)


bfineran (approved):

bfineran (changes requested):
check `quality` and can we see if this is in the other tasks as well?

src/sparseml/transformers/sparsification/question_answering.py
```diff
@@ -210,7 +210,7 @@ def __init__(
     def _remove_unused_columns(
         self, dataset: "datasets.Dataset", description: Optional[str] = None
     ):
-        if self._signature_columns is None and self.teacher is not None:
+        if self._signature_columns is None and self.teacher is not None and self.teacher not in ("disable", "self"):
```
natuan, line 213: Can you put in the same fix to a similar function inside sparsification/trainer.py?

natuan (changes requested):

dbogunowicz:
@natuan to my best knowledge, those are the necessary changes (tested all the 3 pipelines, excluding mlm)

natuan (approved):

bfineran (approved):

PR #738, opened 2022-04-28
Added flag to exclude quantization of embedding activations.

anmarques:
Embeddings were being quantized twice: as weights before the gather op, and as activations after the op. This PR removes the latter.


bfineran:
looks like there's some failing tests

natuan (approved):

PR #749, opened 2022-05-04
custom data column name support in text_classification train script

bfineran:
this PR adds support for custom data column name support in the `sparseml.transformers.text_classification` train script.

`input_column_names` allows for one or two column names in the input csv/json file to be specified as inputs to the model/tokenizer
`label_column_name` gives the user the ability to override the default `label` name for model targets

existing behavior should not be modified based on these changes

### DIFF
--- a/src/sparseml/transformers/text_classification.py
+++ b/src/sparseml/transformers/text_classification.py
@@ -183,6 +183,27 @@ class DataTrainingArguments:
         default=None,
         metadata={"help": "A csv or a json file containing the test data."},
     )
+    input_column_names: Optional[str] = field(
+        default=None,
+        metadata={
+            "help": (
+                "name of column to read model input data from. may also be comma "
+                "separated list of two columns to use as inputs. examples include "
+                "'sentence' for single column and 'sentence_1,sentence_2' for two. "
+                "Default behavior is to read columns based on task name or infer from "
+                "non 'label' columns if sentence_column_names and task name not"
+                "provided"
+            )
+        },
+    )
+    label_column_name: str = field(
+        default="label",
+        metadata={
+            "help": (
+                "column in dataset where input labels are located. Default is 'label'"
+            )
+        },
+    )
 
     def __post_init__(self):
         if self.task_name is not None:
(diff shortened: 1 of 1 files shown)


src/sparseml/transformers/text_classification.py
```diff
@@ -183,6 +183,27 @@ class DataTrainingArguments:
         default=None,
         metadata={"help": "A csv or a json file containing the test data."},
     )
+    input_column_names: Optional[str] = field(
+        default=None,
+        metadata={
+            "help": (
+                "name of column to read model input data from. may also be comma "
+                "separated list of two columns to use as inputs. examples include "
```
KSGulin, line 191: nit
```suggestion
                "name of column to read model input data from. May also be comma "
                "separated list of two columns to use as inputs. Examples include "
```

src/sparseml/transformers/text_classification.py
```diff
@@ -464,13 +486,27 @@ def main():
     )
 
     # Preprocessing the datasets
+    if data_args.input_column_names is not None:
+        if "," in data_args.input_column_names:
+            # two input columns
+            columns = data_args.input_column_names
```
KSGulin, line 492: It looks like the original CLI string is getting assigned here, which would fail the logic below
```suggestion
            columns = data_args.input_column_names.replace(' ', '').split(',')
```

mgoin, line 492: doesn't this need to split input_column_names on ","?

src/sparseml/transformers/text_classification.py
```diff
@@ -464,13 +486,27 @@ def main():
...
+            if len(columns) != 2:
+                raise ValueError(
+                    "input_column_names may only specify up to two columns "
+                    f"{len(columns)} provided: {columns}"
+                )
+            sentence1_key, sentence2_key = columns
+        else:
+            # one input column
+            sentence1_key = data_args.input_column_names
+            sentence2_key = None
```
KSGulin, line 502: In the lines following this one, it seems that the keys get overwritten if a task name is set. Is this the expected behavior?

KSGulin (commented):
Left a few suggestions. Otherwise LGTM!

src/sparseml/transformers/text_classification.py
bfineran: going to keep spaces in in case a dataset may have a space in a column name

bfineran: yeah, missed this on commit

src/sparseml/transformers/text_classification.py
bfineran: switching below to elif

bfineran: input_column_names should take priority

bfineran:
thanks @mgoin @KSGulin responded

mgoin (approved):
thanks lgtm!

InquestGeronimo (approved):

Issue #750, opened 2022-05-04
Sparsity miscalculated


eldarkurtic:
**Describe the bug**
I think that the property `_last_applied_sparsity` of the `PruningParamsScorer` class is miscalculated at https://github.com/neuralmagic/sparseml/blob/a6477f900b55afd555734fe6cf784e0137d815a5/src/sparseml/pytorch/sparsification/pruning/mask_params.py#L376, because of the three reasons:
1. applying newly created masks happens in the next line (i.e. https://github.com/neuralmagic/sparseml/blob/a6477f900b55afd555734fe6cf784e0137d815a5/src/sparseml/pytorch/sparsification/pruning/mask_params.py#L380), so at the time when sparsity of `self.params_data[0]` is calculated those params are not yet masked
2. sparsity should be calculated from masks, not from param values. This is mostly because sometimes optimizers might set some param values to zero and as such they would be considered as pruned.
3. when `global_sparsity` is used (for example with `GlobalMagnitudePruningModifier`), getting the sparsity information from `self.params_data[0]` will result with incorrect number as it considers only the first param in the list of all params

PR #751, opened 2022-05-04
Fix sparsity calculation via masks-sparsity

eldarkurtic:
Should fix https://github.com/neuralmagic/sparseml/issues/750 by calculating sparsity from newly created masks.

### DIFF
--- a/src/sparseml/pytorch/sparsification/pruning/mask_params.py
+++ b/src/sparseml/pytorch/sparsification/pruning/mask_params.py
@@ -26,7 +26,7 @@ from torch.nn import Module, Parameter
 
 from sparseml.pytorch.sparsification.pruning.mask_creator import PruningMaskCreator
 from sparseml.pytorch.sparsification.pruning.scorer import PruningParamsScorer
-from sparseml.pytorch.utils import mask_difference, tensor_sparsity
+from sparseml.pytorch.utils import mask_difference, tensor_list_sparsity
 
 
 __all__ = [
@@ -373,9 +373,7 @@ class ModuleParamPruningMask(object):
         )
 
         if self._scorer:
-            self._scorer.update_last_applied_sparsity(
-                tensor_sparsity(self.params_data[0])
-            )
+            self._scorer.update_last_applied_sparsity(tensor_list_sparsity(masks))
 
         return self.set_param_masks(masks)
 
--- a/src/sparseml/pytorch/utils/helpers.py
+++ b/src/sparseml/pytorch/utils/helpers.py
@@ -658,7 +658,8 @@ def tensor_list_sparsity(tensors: List[Tensor]) -> float:
     :param tensors: the list of tensors to calculate the sparsity for
     :return: the total sparsity of all tensors in the list
     """
-    zeros, numel = 0, 0
+    zeros = 0
+    numel = 0
     for tensor in tensors:
         zeros += (tensor == 0).sum().item()
         numel += tensor.numel()
--- a/src/sparseml/pytorch/utils/helpers.py
+++ b/src/sparseml/pytorch/utils/helpers.py
@@ -70,6 +70,7 @@ __all__ = [
     "tensors_export",
     "tensor_density",
     "tensor_sparsity",
+    "tensor_list_sparsity",
     "tensor_sample",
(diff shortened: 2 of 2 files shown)


src/sparseml/pytorch/sparsification/pruning/mask_params.py
```diff
@@ -373,9 +373,7 @@ def update_param_masks(self, target: Union[float, List[float]]) -> List[Tensor]:
         )
 
         if self._scorer:
-            self._scorer.update_last_applied_sparsity(
-                tensor_sparsity(self.params_data[0])
-            )
+            self._scorer.update_last_applied_sparsity(tensor_list_sparsity(masks))
 
         return self.set_param_masks(masks)
```
KSGulin, line 378: Updated to use parameters instead of masks
```suggestion
        mask_diff = self.set_param_masks(masks)

        if self._scorer:
            self._scorer.update_last_applied_sparsity(
                tensor_list_sparsity(self.params_data)
            )

        return mask_diff
```

src/sparseml/pytorch/sparsification/pruning/mask_params.py
```diff
@@ -373,9 +373,7 @@ def update_param_masks(self, target: Union[float, List[float]]) -> List[Tensor]:
         )
 
         if self._scorer:
```
KSGulin, line 375: See next comment for full suggestion
```suggestion
```

KSGulin (commented):
Great fix, thanks for putting this up @eldarkurtic! The one suggestion I have is to use the true sparsity (via params) rather than the masks. Although it's true that the optimizer can set some weights to zero, we expect this to be fairly rare later in training and the design standard is to report the true sparsity

src/sparseml/pytorch/sparsification/pruning/mask_params.py
eldarkurtic: @KSGulin the true sparsity is the sparsity that the pruner imposed on masks and this is what is measured/compared. Making use of weights sparsity is not reliable because of the two following use-cases (we had issues with both of them before and unfortunately learned this the hard way):
1. optimizer can set some weights to zero (this actually happened a lot in experiments with Yolo models) and the sparsity will increase a bit, but since masks are not zero, those weights can become non-zero in the following step and the reported sparsity will decrease again. This creates a super weird behaviour and it doesn't work well with sparsity schedulers which rely on the fact that sparsity is a monotonically increasing function.
2. structured pruners (for example 4-block pruners) rely on the fact that the number of non-zero weights is always divisible by the block-size (in this case 4), and then the weights that are set to zero by optimizer will ruin that structure (i.e. block-pruning won't work anymore)

eldarkurtic: Using the true sparsity from params would definitely be the good idea but in the final stage,  after the pruning run has finished and the model is being exported for deployment (and when final sparsities are being reported). At that point we don't have to care about any structure in the masks as the model will be used just for inference.

bfineran: great points @eldarkurtic, especially for (2) we should be respecting any mask structure. `tensor_list_sparsity` looks like a clean solution

src/sparseml/pytorch/utils/helpers.py
```diff
@@ -652,6 +653,18 @@ def tensor_sample(
     return samples
 
 
+def tensor_list_sparsity(tensors: List[Tensor]) -> float:
+    """
+    :param tensors: the list of tensors to calculate the sparsity for
+    :return: the total sparsity of all tensors in the list
+    """
+    zeros, numel = 0, 0
```
bfineran, line 661:
```suggestion
    zeros = 0
    numel = 0
```

bfineran: small nit based on codebase

bfineran (approved):
using global sparsity based on masks makes sense, thanks @eldarkurtic

bfineran (approved):

KSGulin (approved):
Good points @eldarkurtic. I agree then, calculating sparsity on masks is the better route

Issue #763, opened 2022-05-05
Collecting gradients seems like the process is hanging


eldarkurtic:
**Is your feature request related to a problem? Please describe.**
When collecting gradients for pruning, there are no messages printed/logged which makes one think that the process crashed or is hanging and needs to be restarted.

**Describe the solution you'd like**
Some updates/logs to know that gradients are being collected in the background properly.

eldarkurtic:
Collecting gradients can sometimes take a lot of time (for example up to 40 minutes), so it's useful to have some indicators/logs that the process is running properly.

PR #764, opened 2022-05-05
Log status while collecting grads

eldarkurtic:
A solution to https://github.com/neuralmagic/sparseml/issues/763
It's currently hardcoded to report every 100 steps, as printing at each step might clutter the stdout completely. 
We can also add it as a recipe argument so that it's configurable, but I think it will just introduce a new param that has to be explained/set while creating recipes so probably better to leave it hardcoded to some reasonable value.


bfineran (commented):
thanks @eldarkurtic, looks like a reasonable resolution to #763.

Thoughts on adding this as a modifier debug log in the MFACPruningModifier (when `computed_grads` is yielded)?

eldarkurtic:
@bfineran that would actually be much better, as we would tie the log to the specific modifier. Thanks!

PR #753, opened 2022-05-05
[MFAC] Change logging level from debug to info

eldarkurtic:
Should fix https://github.com/neuralmagic/sparseml/issues/752

PR #766, opened 2022-05-05
[M-FAC] Logging Update

### LINKED ISSUE #752: Important messages during MFAC pruning not printed by default
**Feature request**
There are a lot of things happening in the MFAC pruner ("under the hood") and they are currently logged at the debug level, which means they won't be printed during the standard run. I think these should be printed by default as info messages, as they are very important for anyone running any kind of MFAC pruning to be aware of what is going on. 
This will prevent and inform the user when situations like these occur, for example:
- when `num_grads` is huge, the standard output will look like experiment is hanging while in the background MFAC pruner is just collecting a lot of gradients which can take up to an hour in some cases
- some exceptions (regarding the usage of resources) will be raised and handled silently, and then attempt a new config to fit on a device while the user is completely unaware of these things happening in the background (for example to configure them better in the future)

### LINKED ISSUE #763: Collecting gradients seems like the process is hanging

KSGulin:
M-FAC Modifier logging update to address #753 and #763

Changes:
- Optional progress bar for the GradSampler, which is turned on by default
- Log M-FAC modifier debug messages as modifier logs

Not changed:
- Woodfisher calculation logs. The classes containing these calculations are separate from the modifier and are also invoked by the scorer rather than the modifier. It's non-trivial to have their logs included in the modifier logging. Potential future solution is to include a more global access to modifier logging so that auxiliary routines can make use of them as well.

### DIFF
--- a/src/sparseml/pytorch/sparsification/modifier.py
+++ b/src/sparseml/pytorch/sparsification/modifier.py
@@ -701,18 +701,17 @@ class ScheduledModifier(Modifier, BaseScheduled):
         loggers: Optional[LoggerManager] = None,
         epoch: Optional[float] = None,
         steps_per_epoch: Optional[int] = None,
+        level: Optional[int] = None,
     ):
-        if not tag:
-            tag = type(self).__name__
-        if not loggers:
-            loggers = self.loggers
-        if epoch and steps_per_epoch:
-            step = loggers.epoch_to_step(epoch, steps_per_epoch)
-        else:
-            step = None
-        loggers.log_string(
-            tag=tag, string=string, step=step, level=LOGGING_LEVELS["debug"]
+        tag = tag or type(self).__name__
+        loggers = loggers or self.loggers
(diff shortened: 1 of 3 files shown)


bfineran (commented):
looks good to me overall @KSGulin. can you look into the failing test?

bfineran:
@KSGulin looks like the issue is that `contextlib.nullcontext` is not available in python 3.6. I think your logic will still work without needing a context (ie just updating pbar if it's not None), can you take a quick look?

KSGulin:
@bfineran found a solution simpler than my previous approach. Ready for re-review.

bfineran (approved):

rahul-tuli (approved):

PR #782, opened 2022-05-12
Allow GradSampler to be of type Callable

KSGulin:


src/sparseml/pytorch/utils/sparsification.py
```diff
@@ -223,9 +233,12 @@ def iter_module_backwards(
             total=num_grads, desc="Collecting gradients", disable=not progress_bar
         )
 
+        data_loader = (
+            self._data_loader() if callable(self._data_loader) else self._data_loader
+        )
         with pbar:
             while computed_grads < num_grads:
-                for forward_args, forward_kwargs, loss_target in self._data_loader:
+                for forward_args, forward_kwargs, loss_target in data_loader:
```
bfineran, line 241: shouldn't we "refresh" the data_loader if `callable` between each of the inner loops?

bfineran (changes requested):

src/sparseml/pytorch/utils/sparsification.py
KSGulin: The refreshing happens on line 236, so that data_loader is always a generator and never a callable

bfineran: but when the generator runs out, shouldn't we need to refresh again before line 241?

rahul-tuli: I agree with @bfineran , I think line 265 should be inside the while loop but before the for loop;

The idea being, for loop over an Iterable would start from the first element(for every while loop iteration), but the generator once consumed would just die out and throw StopIterations

KSGulin: The dataset does actually refresh every time the generator is iterated over (line 236 is run again). But your suggestion is a good one because it covers edge cases such as when the number of gradients > the number of training samples (which may come up in test runs). Updated 🙂

bfineran: is the update pushed?

bfineran (dismissed):

src/sparseml/pytorch/utils/sparsification.py
KSGulin: Ah, it is now.

dbogunowicz (approved):

bfineran (approved):

PR #786, opened 2022-05-13
support for separate teacher model tokenizer in text_classification training

bfineran:
this PR adds support for using a separate tokenizer for the teacher model (from the teacher's name or path) in the `sparseml.transformers.text_classification` training script using a `--use_teacher_tokenizer` kwarg.  This may be useful when the student and teacher models have different tokenizers.

after approval, support in other scripts may be added - as is, the sparseml trainer integration will look for columns marked `distill_teacher:` for inputs to teacher models if available. This additional identifier serves two purposes 1) to prevent naming conflicts 2) to prevent additional args from leaking between teacher/student if there are valid optional kwargs that exist for both models but only are provided by one of the tokenizers

The following toy command demonstrates usage for sparsifying a BERT model with RoBERTa teacher - notice that the command will fail if `--use_teacher_tokenizer` is not set

```bash
sparseml.transformers.train.text_classification \
    --output_dir delphai-models/sparse_quantized \
    --model_name_or_path zoo:nlp/masked_language_modeling/bert-base/pytorch/huggingface/wikipedia_bookcorpus/12layer_pruned80_quant-none-vnni \
    --recipe zoo:nlp/masked_language_modeling/bert-base/pytorch/huggingface/wikipedia_bookcorpus/12layer_pruned80_quant-none-vnni?recipe_type=transfer-text_classification \
    --recipe_args '{"init_lr":6e-5}' \
    --distill_teacher roberta-base \
    --task_name sst2  \
    --max_seq_length 128 --per_device_train_batch_size 12 --per_device_eval_batch_size 12 --preprocessing_num_workers 6 \
    --do_train --do_eval --evaluation_strategy epoch --fp16 --seed 2819  \
    --save_strategy epoch --save_total_limit 1 \
    --overwrite_output_dir
```


dbogunowicz (dismissed):
lgtm

src/sparseml/transformers/text_classification.py
```diff
@@ -563,14 +585,31 @@ def main():
...
             f"Using max_seq_length={tokenizer.model_max_length}."
         )
+    if teacher_tokenizer and max_seq_length > teacher_tokenizer.model_max_length:
+        _LOGGER.warning(
+            f"The max_seq_length passed ({max_seq_length}) is larger than "
+            "the maximum length for the teacher model "
+            f"({teacher_tokenizer.model_max_length}). "
+            f"Using max_seq_length={teacher_tokenizer.model_max_length}."
+        )
     max_seq_length = min(data_args.max_seq_length, tokenizer.model_max_length)
```
anmarques, line 602: Contrary to the warning above, max_seq_length is not taking into consideration teacher_tokenizer.model_max_length

anmarques (changes requested):
Please see comment about max_seq_length. LGTM other than that for text_classification. Any reason we don't contemplate other tasks in the PR?

src/sparseml/transformers/text_classification.py
bfineran: good catch, removed

bfineran:
thanks @anmarques. we can add for other tasks in a follow up. only added for text classification because 1) it came from a specific user request 2) wanted to get this reviewed for one task before making the same changes to all others (simplify review process)

anmarques (approved):

rahul-tuli (approved):

PR #797, opened 2022-05-18
Integrations testing for transformers

KSGulin:


bfineran (commented):
LGTM so far, even though it's not complete here, `test_transformers` shows the flexibility of the framework given how compact it is.  Starting to notice some code that can be shared across integrations for the test impls.

Assuming we are adding in export test impls before merging?

bfineran (commented):
Let's add a config that tests export. Also worried if the deepsparse random data generator is sufficient given items such as attention mask need a specific format. Might be best to have sample tokenizer input given in the config then use the tokenizer to generate sample data. (Can see how this flow may work in our transformers onnx export script)

dbogunowicz (approved):
Can we remove my (+ Konstantin's) transformers tests from `tests/sparsezoo/transformers` then? They seem to be redundant now.

tests/integrations/transformers/test_transformers.py
```diff
@@ -0,0 +1,247 @@
...
+
+    sample_input = {}
+    output_names = []
+
+    for input_a, input_b in zip(model_a.graph.input, model_b.graph.input):
+        assert input_a.name == input_b.name
+        input_a_shape = _get_tensor_shape(input_a)
+        assert input_a_shape == _get_tensor_shape(input_b)
+        sample_input[input_a.name] = numpy.random.randn(*input_a_shape).astype(
+            numpy.float32
```
bfineran, line 228: inputs to the BERT models are INTs I believe, will this run as expected?

bfineran (changes requested):

rahul-tuli (approved):

bfineran (approved):

PR #807, opened 2022-05-21
set end_comparator of LayerPruningModifier to -1

bfineran:
Under the current `end_comparator` the end epoch of `LayerPruningModifier` must be greater than the start epoch. This would make a schedule like `start_epoch=2.0, end_epoch=-1` invalid which is a bug, since we should be able to start at an arbitrary epoch (`start_epoch=2.0`) and run indefinitely (`end_epoch=-1.0`)

`end_comparator` specs: 
https://github.com/neuralmagic/sparseml/blob/29119f73596fbea01a0b82133696305995a5b72c/src/sparseml/optim/modifier.py#L630
https://github.com/neuralmagic/sparseml/blob/29119f73596fbea01a0b82133696305995a5b72c/src/sparseml/optim/modifier.py#L631

### DIFF
--- a/src/sparseml/pytorch/sparsification/pruning/modifier_pruning_layer.py
+++ b/src/sparseml/pytorch/sparsification/pruning/modifier_pruning_layer.py
@@ -70,6 +70,7 @@ class LayerPruningModifier(ScheduledUpdateModifier):
             start_epoch=start_epoch,
             end_epoch=end_epoch,
             update_frequency=-1.0,
+            end_comparator=-1,
         )
         self._layers = validate_str_iterable(
             layers, "{} for layers".format(self.__class__.__name__)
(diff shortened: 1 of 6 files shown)


mgoin (approved):

dbogunowicz (approved):

spacemanidol (approved):

src/sparseml/pytorch/sparsification/pruning/modifier_pruning_layer.py
```diff
@@ -70,6 +70,7 @@ def __init__(
             start_epoch=start_epoch,
             end_epoch=end_epoch,
             update_frequency=-1.0,
+            end_comparator=-1,
```
bfineran, line 73: noting that we should not merge this PR yet. this fix uncovered potential issues with the logic surrounding the implementation of `end_comparator=-1`

bfineran:
blocking landing - see unresolved comment

dbogunowicz:
@bfineran what needs to be completed prior to this PR being merged?

markurtz (approved):

rahul-tuli (approved):

bfineran:
@dbogunowicz see latest non merge commit - updated `end_comparator=-1` to only allow `end_epoch < start_epoch` if `end_epoch==-1`. a few more changes will be needed on the recipe composition side to be added in a future PR

PR #810, opened 2022-05-23
[Integrations-Testing] Add abridged mode

KSGulin:
Add an abridged mode which automatically shortens the run. The currently shortening logic sets the number of train and eval steps to a default value (10 was chosen arbitrarily), but the logic can be expanded to include abridged deploy runs once implemented. Adding this keyword allows for the maintenance of a standardized short run mode.


dbogunowicz:
What's the point of the abridged mode? Is it to facilitate testing on the local machines? Is it somehow linked to test frequency (running one test weekly, but abridged test daily)? I am just missing the context ;)

KSGulin:
@dbogunowicz the abridged mode allows us to run a full training run, which may take hours, in minutes. This would be used for testing on commit or on nightly, when we don't want to commit to such long tests, but we still want to test the validity of our pipelines. The same can be achieved by setting args such as "max-train-samples" under the commands fields, but the benefit of the abridged mode is having one standardized definition of the arg values for a shortened run.

tests/integrations/image_classification/args.py
```diff
@@ -120,6 +120,16 @@ class ImageClassificationTrainArgs(_ImageClassificationBaseArgs):
     recipe_args: str = Field(
         default=None, description="json string for recipe constructor"
     )
+    max_train_steps: int = Field(
+        default=-1,
+        description="The maximum number of training steps to run per epoch. If "
+        "negative, will run for the entire dataset",
```
dbogunowicz, line 126:
```suggestion
        "negative, will run for the entire training set",
```

tests/integrations/image_classification/args.py
```diff
@@ -120,6 +120,16 @@ class ImageClassificationTrainArgs(_ImageClassificationBaseArgs):
...
+    )
+    max_eval_steps: int = Field(
+        default=-1,
+        description="The maximum number of eval steps to run per epoch. If negative, "
+        "will run for the entire dataset",
```
dbogunowicz, line 131:
```suggestion
        "will run for the entire validation set",
```

tests/integrations/object_detection/args.py
```diff
@@ -137,3 +137,13 @@ class Yolov5ExportArgs(BaseModel):
     remove_grid: bool = Field(
         default=False, description="remove export of Detect() layer grid"
     )
+    max_train_steps: int = Field(
+        default=-1,
+        description="Set the maximum number of training steps per epoch. if negative,"
+        "the entire dataset will be used, default=-1",
```
dbogunowicz, line 143:
```suggestion
        "the entire training set will be used, default=-1",
```

tests/integrations/object_detection/args.py
```diff
@@ -137,3 +137,13 @@ class Yolov5ExportArgs(BaseModel):
...
+    )
+    max_eval_steps: int = Field(
+        default=-1,
+        description="Set the maximum number of eval steps per epoch. if negative,"
+        "the entire dataset will be used, default=-1",
```
dbogunowicz, line 148:
```suggestion
        "the entire validation set will be used, default=-1",
```

tests/integrations/transformers/test_transformers.py
```diff
@@ -72,6 +72,11 @@ def capture_pre_run_state(self):
         if train_args:
             train_args.output_dir = self.save_dir.name
 
+    def add_abridged_configs(self):
+        if "train" in self.command_types:
+            self.configs["train"].max_steps = 10
```
bfineran, line 77:
```suggestion
            self.configs["train"].max_train_samples = 10
```

bfineran (changes requested):

PR #811, opened 2022-05-23
[Integrations-Testing] Re-organize support functions

KSGulin:
This PR reorganizes and updates the integrations testing in the following ways:

- Move Config class to its own module (config.py)
- Move configs for each integrations into a configs subdirectory (only subdirectory now searched for config files)
- Consolidate common auxiliary testing functions in helpers.py
- Simplify Transformers args by moving args common between integrations into the base args class
- Update syntax and address various nits found throughout


bfineran (approved):
LGTM, please confirm in PR description tests are passing locally before merging

KSGulin:
@bfineran for all the integrations test to pass we need this and #848 to land. Once these land on [integrations-testing-updates](https://github.com/neuralmagic/sparseml/tree/integrations-testing-updates) they will be tested locally. #849 will add automatic integrations testing, so we'll be sure all tests are passing before landing on main.

kylesayrs (approved):

tests/integrations/config.py
```diff
@@ -0,0 +1,99 @@
...
+        args_dict = self.run_args.dict()
+        args_string_list = []
+        for key, value in args_dict.items():
+            key = "--" + key
+            key = key.replace("_", "-") if self.dashed_keywords else key
+            # Handles bool type args (e.g. --do-train)
+            if isinstance(value, bool):
+                if value:
+                    args_string_list.append(key)
+            elif isinstance(value, List):
```
rahul-tuli, line 68: nit: can be ignored but I find it more readable to use builtin objects

```suggestion
            elif isinstance(value, list):
```

rahul-tuli, line 68: Also maybe we can focus on behavior rather than actual type, as in check if value is Iterable rather than an actual List

```
from collections.abc import Iterable
isinstance(value, Iterable)
```

tests/integrations/config.py
```diff
@@ -0,0 +1,99 @@
...
+                # Handles args that are both bool and value based (see evolve in yolov5)
+                if isinstance(value[0], bool):
```
rahul-tuli, line 70: Potential Error if value is []

```suggestion
                if value and isinstance(value[0], bool):
```

tests/integrations/config.py
```diff
@@ -0,0 +1,99 @@
...
+                    if value[0]:
```
rahul-tuli, line 71: Could we also check if `len(value) >=2`; `value[1]` is referenced on line 72

tests/integrations/config.py
```diff
@@ -0,0 +1,99 @@
...
+        pre_args = self.pre_args.split(" ") if self.pre_args else []
+        return pre_args + [self.command_stub] + args_string_list
+
+    def _validate_config(self):
+        # Check that all provided run args correspond to expected run args. Mainly
+        # meant to catch typos
+        unknown_run_args = [
+            arg
+            for arg in self.raw_config["command_args"].keys()
+            if arg not in self.run_args.__fields_set__
```
rahul-tuli, line 93: noice!

rahul-tuli (commented):
Few Small nits!

PR #813, opened 2022-05-24
[BugFix] AttributeError when fetching val loader for transformers sample input/output exports

rahul-tuli:
Wrap getting val data loader in a try except clause to deal with errors while fetching val data loader, rely on train_loader if can't get the val loader

### DIFF
--- a/src/sparseml/transformers/sparsification/trainer.py
+++ b/src/sparseml/transformers/sparsification/trainer.py
@@ -499,7 +499,11 @@ class RecipeManagerTrainerInterface:
         os.makedirs(sample_out_dir, exist_ok=True)
         device = self.model.device
 
-        dataloader = self.get_val_dataloader() or self.get_train_dataloader()
+        try:
+            dataloader = self.get_val_dataloader()
+        except Exception:
+            dataloader = self.get_train_dataloader()
+
         _LOGGER.info(f"Exporting {num_samples_to_export} samples to {output_dir}")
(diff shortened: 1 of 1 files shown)


src/sparseml/transformers/sparsification/trainer.py
```diff
@@ -499,7 +499,11 @@ def save_sample_inputs_outputs(
         os.makedirs(sample_out_dir, exist_ok=True)
         device = self.model.device
 
-        dataloader = self.get_val_dataloader() or self.get_train_dataloader()
+        try:
```
dbogunowicz, line 502: What's the intention here? The original code will pick either of two, with the priority given for `val_dataloader`. The refactor imo does the same, but also makes the logic more error-tolerant, is this the intention?

dbogunowicz (approved):

src/sparseml/transformers/sparsification/trainer.py
rahul-tuli: Yes! some `Trainer`(s) might not have `get_val_dataloader(...)` attribute, which will raise a `AttributeError` in original code, The `try ... except` clause catches such errors and delegates fetching dataloader to `get_train_dataloader(...)` in case of errors; if `get_train_dataloader(...)` also throws an error then that should be reported to the end user. Does that make sense?

KSGulin (approved):

bfineran (approved):

PR #814, opened 2022-05-25
[WIP] Refactor transformers trainer classes for customed save checkpoint

natuan:
This change is to have a single trainer class, derived from HF Trainer class, with customized behaviors that could be shared by all SparseML transformers trainer classes. In addition to save checkpoint in this PR, the func `_remove_unused_columns` could be moved up to avoid being repeated.


natuan:
replaced by https://github.com/neuralmagic/sparseml/pull/861

PR #789, opened 2022-05-16
Onnxruntime as a separate dependency (asserting compatibility of onnxruntime-gpu)

dbogunowicz:
The goal is to assert that:
- if you run any ORT dependent code on `sparseml/src` without ORT installed you get an error telling you to install `sparseml[onnxruntime]`.
- otherwise run normally. 

Also: removes `onnxruntime` from the standard `setup.py` deps.

Relates to issue: https://github.com/neuralmagic/sparseml/issues/704
Overrides PR: https://github.com/neuralmagic/sparseml/pull/723

Manual testing:
1. Pull this branch and initialize a fresh venv (`python -m pip install -e ./[dev]`).
2. Remove onnxruntime from the venv
3. In python console run:
`import src.sparseml.onnx.utils.helpers`
Output:
```
In [1]: import src.sparseml.onnx.utils.helpers
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
File ~/sparseml_copy/sparseml/src/sparseml/onnx/base.py:31, in <module>
     30 try:
---> 31     import onnxruntime
     33     onnxruntime_err = None

ModuleNotFoundError: No module named 'onnxruntime'

During handling of the above exception, another exception occurred:

ModuleNotFoundError                       Traceback (most recent call last)
Input In [1], in <cell line: 1>()
----> 1 import src.sparseml.onnx.utils.helpers

File ~/sparseml_copy/sparseml/src/sparseml/onnx/__init__.py:21, in <module>
     15 """
     16 Functionality for working with and sparsifying Models in the ONNX/ONNXRuntime framework
     17 """
     19 # flake8: noqa
---> 21 from .base import *
     22 from .benchmark import *
     23 from .framework import detect_framework, framework_info, is_supported

File ~/sparseml_copy/sparseml/src/sparseml/onnx/base.py:36, in <module>
     33     onnxruntime_err = None
     35 except ModuleNotFoundError as error:
---> 36     raise error.__class__(
     37         f"{error.msg}. To fix this error, install sparseml[onnxruntime]."
     38     )
     40     onnxruntime = object()  # TODO: populate with fake object for necessary imports
     41     onnxruntime_err = error

ModuleNotFoundError: No module named 'onnxruntime'. To fix this error, install sparseml[onnxruntime].
```

`import src.sparseml.onnx.utils.model`
Output: 
```
In [1]: import src.sparseml.onnx.utils.helpers
---------------------------------------------------------------------------
[...]

PR #815, opened 2022-05-25
delay onnxruntime import error until invocation of dependent functions

bfineran:
#789 makes `onnxruntime` an optional dependency of sparseml. This PR fixes two issues that led to import of non ORT dependent functions (such as `sparseml.pytorch.optim.ScheduledModifierManager`) run into import errors

1) removes the addition of a hard check of ORT import on any import of `sparseml.onnx`
2) moves onnxruntime import in the two `src/` functions that require ORT to inside the functions, using the `require_onnxruntime` decorator as a checker

### DIFF
--- a/src/sparseml/base.py
+++ b/src/sparseml/base.py
@@ -239,6 +239,7 @@ def check_version(
     max_version: Optional[str] = None,
     raise_on_error: bool = True,
     alternate_package_names: Optional[List[str]] = None,
+    extra_error_message: Optional[str] = None,
 ) -> bool:
     """
     :param package_name: the name of the package to check the version of
@@ -255,6 +256,9 @@ def check_version(
     :param alternate_package_names: List of alternate names to look for the package
         under if package_name is not found. Useful for nightly builds.
     :type alternate_package_names: Optional[List[str]]
+    :param extra_error_message: optional string to append to error message if error
+        is raised
+    :type extra_error_message: Optional[str]
     :return: If raise_on_error, will return False if the package is not installed
         or the version is outside the accepted bounds and True if everything is correct.
     :rtype: bool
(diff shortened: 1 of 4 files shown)


src/sparseml/onnx/base.py
```diff
@@ -32,11 +32,7 @@
 
     onnxruntime_err = None
 
-except ModuleNotFoundError as error:
-    raise error.__class__(
-        f"{error.msg}. To fix this error, install sparseml[onnxruntime]."
-    )
-
+except Exception as error:
```
dbogunowicz, line 35: Looking good, just tested it and I like how it really pushes the error instantiation as late as possible. However, I would keep the message for the user that points to `sparseml[onnxruntime]` installation. This way the user will know that this is not some random bug off the happy path, but something which we account for and have a solution to.

bfineran: good call, added something in

dbogunowicz (approved):

src/sparseml/base.py
```diff
@@ -255,6 +256,9 @@ def check_version(
     :param alternate_package_names: List of alternate names to look for the package
         under if package_name is not found. Useful for nightly builds.
     :type alternate_package_names: Optional[List[str]]
+    :param extra_error_message: optional string to append to error message if error
+        is raised
+    :type extra_error_message: Optional[str]
```
rahul-tuli, line 261: Is this a new standard we're following? We already have the types annotated in signature, is this required?

rahul-tuli (approved):

src/sparseml/base.py
bfineran: not a standard - just keeping consistent with the existing style in this docstring

PR #819, opened 2022-05-26
Add support for optional dist backend in param scoring

eldarkurtic:
Currently, PruningParamsGradScorer handles gradient communication between processes in DDP runs by creating a gloo backend. This was initially done this way to overcome limitations of the nccl backend when passing around gradients as cpu-tensors. Some pruners (for example oBERT) need to communicate only one gpu-tensor, and thus can use the default distributed backend that comes with DDP (i.e. nccl). This implementation should allow creating groups with any distributed backend, and should also take care about cases when additional group is not need (i.e. in the oBERT implementation, with `dist_backend=None`).

Remaining ToDos if this gets accepted: pruning modifiers that are using this scorer should specify which dist. backend to use. 
@KSGulin I assume we will need to update only MFAC modifiers, which should initialize scorer with `dist_backend='gloo'`


bfineran:
hi @eldarkurtic, code looks good overall, any reason why the default is currently set to `None` rather than `'gloo'` to keep existing functionality the default?

eldarkurtic:
@bfineran the gloo backend is a special case used currently only by MFAC pruners to pass around huge cpu-tensors. I think it's not something a user would expect to have as a default behaviour. The default proposed in the PR (with `None`) is going to use whatever the default distributed backend is already initialized by the pytorch DDP launcher, and won't create additional groups unless explicitly specified with the `dist_backend` argument.

bfineran (dismissed):

KSGulin (changes requested):
Good idea and good code @eldarkurtic. The movement pruning modifier should be updated here as well. Otherwise looks good to go

bfineran (approved):

PR #821, opened 2022-05-27
Delay casting masks to tensor dtype

### LINKED ISSUE #820: Unnecessary mask flipping due to rounding errors
**Describe the bug**
When summing up a huge number of ones in torch.float32 type, it seems that rounding errors accumulate and cause unnecessary flipping of masks. More specifically, for a huge `tensor` with size of for example 80M and `torch.float32` data type:
```
mask = (tensor > threshold).type(tensor.dtype)
assert torch.sum(mask) != torch.sum(tensor > threshold)
```
The LHS sums over a huge tensor of ones and zeros in float32, while the RHS sums over a huge tensor of booleans which are casted in int64. The RHS will always have the correct result, and unfortunately the LHS will differ by 1 almost always for huge tensors. 

In the code this makes
https://github.com/neuralmagic/sparseml/blob/bd64454b3ae7b40bccd9057957d5910f45eda4cb/src/sparseml/pytorch/sparsification/pruning/mask_creator.py#L142
to be True and trigger unnecessary flipping of masks, because the `target_num_masked` represents the correct value obtained from `tensor.numel()` call.

**To Reproduce**
Run pruning of a huge model (for example bert-base-uncased) with `global_sparsity=True`.

eldarkurtic:
Fixes https://github.com/neuralmagic/sparseml/issues/820 by keeping the `threshold_mask` values in boolean type (which will be casted to int64 when used in `torch.sum` and result with the correct and expected value).


bfineran (approved):
good catch @eldarkurtic

markurtz (approved):

PR #822, opened 2022-05-27
Update type hint for loss_fn

eldarkurtic:
Given the way `loss_fn` is used https://github.com/neuralmagic/sparseml/blob/bd64454b3ae7b40bccd9057957d5910f45eda4cb/src/sparseml/pytorch/utils/sparsification.py#L247
the type hint should probably be updated to `Callable[[Arg1Type, Arg2Type], ReturnType]`


bfineran (approved):

src/sparseml/pytorch/utils/sparsification.py
```diff
@@ -200,7 +200,7 @@ class GradSampler:
     def __init__(
         self,
         data_loader: Union[Iterator[Tuple[List[Any], Dict[str, Any], Any]], Callable],
-        loss_fn: Callable[[Any], Any],
+        loss_fn: Callable[[Any, Any], Any],
```
kylesayrs, line 203: There are loss functions that take 3 arguments: https://pytorch.org/docs/stable/generated/torch.nn.TripletMarginLoss.html#torch.nn.TripletMarginLoss

Make sure that only loss functions with 2 arguments are passed to `loss_fn`

kylesayrs: Never mind, function is only used in one place, lgtm

kylesayrs (changes requested):

kylesayrs (approved):

markurtz (approved):

Issue #827, opened 2022-05-27
Sorting huge tensors prohibitively expensive


eldarkurtic:
This issue is related to usage of the `torch.sort` in https://github.com/neuralmagic/sparseml/blob/bd64454b3ae7b40bccd9057957d5910f45eda4cb/src/sparseml/pytorch/sparsification/pruning/mask_creator.py#L189
The main problem arises when `tensor` is huge, which happens for example when global_sparsity is enabled or when the model simply has a large number of weights in a layer. In these cases, the run is going to crash with cuda OOM exception because the `torch.sort` won't be able to fit on the device. 
More details about this high memory usage of `torch.sort`: https://github.com/pytorch/pytorch/issues/77049#issuecomment-1125249066
Given that the `torch.sort` is called to determine the threshold value for a given sparsity target, we don't need the entire tensor to be sorted.

PR #828, opened 2022-05-27
Use kthvalue instead of sort

eldarkurtic:
Fixes https://github.com/neuralmagic/sparseml/issues/827

Important notes:
- the proposed solution in this PR doesn't produce any memory overhead while returning the desired threshold value, so pruning experiments with huge models are runnable again and won't crash with cuda OOM
- unfortunately, it's much slower than `torch.sort` (see [https://github.com/pytorch/pytorch/issues/75599#issuecomment-1120483478](https://github.com/pytorch/pytorch/issues/75599#issuecomment-1120483478)) and this might affect pruners that are pruning with high frequencies
- for example, running a GMP 10x per epoch over 30 epochs adds ~3 hours of run-time
 - `torch.topk` function might seem like a solution for both problems, memory with `torch.sort` and speed with `torch.kthvalue`, but unfortunately the source code of the `torch.topk` calls `torch.sort` when the given tensor is huge which gets us back to the memory issue

Given that both solutions have some drawbacks, I think that the `torch.kthvalue` should still be preferred as it makes experiments with large models runnable without cuda OOM issues. `torch.sort` is simply too memory-expensive (check examples in the linked pytorch issue: https://github.com/pytorch/pytorch/issues/77049).

To discuss how to get the best of the both worlds @bfineran  @markurtz :
- `torch.sort` should be preferred when memory is not the issue, which happens for example when global_sparsity=False or when the model is simply small, or when the GPU device is large enough to fit the `sort` call
- `torch.kthvalue` should be preferred in all other cases to make experiments runnable on a GPU (so I think it's should be a default solution)
- a potential solution to get the best of both worlds: we could introduce an env variable `MEMORY_BOUNDED` (or something similar) which we can set when we know that the model is small enough for the device and this can trigger usage of `torch.sort` instead of the slower `torch.kthvalue`


bfineran:
thanks for the detailed description @eldarkurtic.  Agreed that we should try and keep `sort` in when memory is not the issue.  is this error occurring in a wide variety of experiments for you with a reasonable amount of GPU memory? 

using an env var sounds reasonable, we could also surround the `sort` call in a try block and raise an exception pointing to the var, or just set the var then so future runs can use kth element instead

eldarkurtic:
@bfineran unfortunately it occurs quite often. It's impossible to run pruning with `global_sparsity=True` on the BERT model with "only" 85M params on the 24GB RTX 3090. Also, unfortunately, the peak memory usage grows quickly with the number of params to sort, so pruning larger models (>100M params like BERT-Large/T5) is probably going to be impossible even with the simple magnitude pruner.

A few questions regarding the implementation:
1. Any preference on the name of the env variable?
2. Would this be a reasonable flow:
- check the env variable and do as it says
- if not set (for example a user is not aware of its existence), call sort in a try-block and if it fails print the warning message to set the env variable for future runs

rahul-tuli (dismissed):

dbogunowicz (dismissed):

bfineran (changes requested):
(2) sounds like a good approach, let's make sure to get the env variable added to some documentation (can help find a good spot)

src/sparseml/utils/helpers.py
```diff
@@ -820,3 +823,32 @@ def json_to_jsonl(json_file_path: str, overwrite: bool = True):
...
+#
+# torch helper functions
+#
+##############################
+
+
+MEMORY_BOUNDED = "MEMORY_BOUNDED"
+
+
+def memory_aware_threshold(tensor: torch.Tensor, idx: int):
```
bfineran, line 838: nit: a pydoc string here

src/sparseml/utils/helpers.py
```diff
@@ -820,3 +823,32 @@ def json_to_jsonl(json_file_path: str, overwrite: bool = True):
...
+    try:
+        if (
+            MEMORY_BOUNDED in os.environ
+            and os.environ[MEMORY_BOUNDED].lower() == "true"
+        ):
+            return torch.kthvalue(tensor.view(-1), idx + 1)[0]
+        else:
+            return torch.sort(tensor.view(-1))[0][idx]
+    except RuntimeError:
+        _LOGGER.warning(
```
bfineran, line 848: thoughts on setting `MEMORY_BOUNDED=True` here so we don't run into this warning in future steps?

eldarkurtic: great idea, thanks Ben!

src/sparseml/utils/helpers.py
```diff
@@ -836,6 +836,14 @@ def json_to_jsonl(json_file_path: str, overwrite: bool = True):
 
 
 def memory_aware_threshold(tensor: torch.Tensor, idx: int):
+    """
+    Finds a threshold at the lookup idx in the most efficient way with available
+    resources. Will be phased out when GPU-memory overhead of torch.sort reduces,
+    or when torch.kthvalue becomes faster than torch.sort.
+    :param tensor: A tensor to find a k-th smallest value in, where k=idx+1
```
bfineran, line 843:
```suggestion
    or when torch.kthvalue becomes faster than torch.sort.

    :param tensor: A tensor to find a k-th smallest value in, where k=idx+1
```

bfineran (approved):
LGTM @eldarkurtic

eldarkurtic:
@bfineran any suggestions on where to document the `MEMORY_BOUNDED` env variable?

bfineran (dismissed):
thanks @eldarkurtic!

kylesayrs (approved):
small nits, otherwise a very cool find

tests/sparseml/pytorch/utils/test_helpers.py
```diff
@@ -880,3 +882,23 @@ def test_thin_model_from_checkpoint(model, state_dict, test_input):
...
+        (torch.rand(1000), 123),
+        (torch.rand(10_000), 4321),
+        (torch.rand(100_000), 12345),
+    ],
+)
+def test_memory_aware_threshold(tensor, idx):
+    dev = torch.device("cuda:0") if torch.cuda.is_available() else torch.device("cpu")
+    tensor = tensor.to(dev)
+
+    os.environ[MEMORY_BOUNDED] = "True"
```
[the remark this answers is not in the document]
eldarkurtic: good catch, thanks @kylesayrs !

kylesayrs (approved):

bfineran (approved):

Issue #823, opened 2022-05-27
Delegate creating the GradSampler to PruningModifiers


eldarkurtic:
Currently, the GradSampler is instantiated in https://github.com/neuralmagic/sparseml/blob/bd64454b3ae7b40bccd9057957d5910f45eda4cb/src/sparseml/transformers/sparsification/trainer.py#L158

We should consider delegating this process to the PruningModifier class that makes use of it, because:
- now, all pruning modifiers will instantiate it, even though they might not need it (for example MagnitudePruningModifier)
- currently `RecipeManagerTrainerInterface` class is creating it with `self._mfac_data_loader()` and `self._mfac_loss_function`, but `MFAC` is not the only pruner that might use the GradSampler, so we probably shouldn't call it `mfac`
- by delegating this process to the PruningModifier class we also have the opportunity to modify hyperparams when dataloader and loss function are created
- this is very important for `MFAC` and `oBERT` pruning modifiers because the batch-size for GradSampler is an important hyperparam to be configurable from the recipe
- currently this is hardcoded to use `train_batch_size/2` https://github.com/neuralmagic/sparseml/blob/bd64454b3ae7b40bccd9057957d5910f45eda4cb/src/sparseml/transformers/sparsification/trainer.py#L667 which will be a problem when running experiments where `train_batch_size` is huge (for example 256); it's going to be very hard to fit that on a GPU

PR #831, opened 2022-05-27
[transformers] Delegate creating GradSampler to PruningModifiers

eldarkurtic:
Attempts to fix https://github.com/neuralmagic/sparseml/issues/823 by passing `trainer` to PruningModifier classes, which should use it to instantiate dataloader and loss for GradSampler with desired hyperparams (given by the recipe).
This addresses only transformers use-case, and similar modifications should probably be done for other models/domains too.


bfineran:
thanks for the PR @eldarkurtic. where would you propose we move the task specific logic required for the dataloader and loss function?

eldarkurtic:
@bfineran my previous solution was bad since it would work only for transformers use-case, so I have reverted the changes and pushed a new version which is a slightly modified version of the already existing code. 

The notable updates would be:
- delegate creating the grad sampler to the PruningModifier class, so that the default data_loader's hyperparams can be overridden by kwargs given in the recipe
-  this should be generic enough to handle other domains/use-cases too

KSGulin (commented):
@eldarkurtic these are good suggestions. I'm in support of them, we just also need to update the logic in the [mfac pruner](https://github.com/neuralmagic/sparseml/blob/9402e573acb9e4e4677f795367845696bd28221c/src/sparseml/pytorch/sparsification/pruning/modifier_pruning_mfac.py#L263) and [yolov5](https://github.com/neuralmagic/yolov5/blob/master/utils/sparse.py) to correspond.

eldarkurtic:
https://github.com/neuralmagic/yolov5/pull/75 complements this PR with required updates in the yolov5 repo

bfineran (dismissed):

src/sparseml/pytorch/sparsification/pruning/modifier_pruning_mfac.py
```diff
@@ -229,6 +235,13 @@ def available_devices(self) -> Optional[List[str]]:
         """
         return self._available_devices
 
+    @ModifierProp(serializable=True)
+    def grad_sampler_kwargs(self) -> Optional[Dict[str, Any]]:
+        """
+        Return dict of training dataloader's overriden configs for gradient sampling
```
kylesayrs, line 241: nit: apostrophe in docstring

bfineran (approved):

src/sparseml/pytorch/sparsification/pruning/modifier_pruning_mfac.py
```diff
@@ -130,6 +132,8 @@ class MFACPruningModifier(BaseGradualPruningModifier):
     :param mask_type: String to define type of sparsity to apply. May be 'unstructred'
         for unstructured pruning or 'block4' for four block pruning or a list of two
         integers for a custom block shape. Default is 'unstructured'
+    :param grad_sampler_kwargs: kwargs to override default train dataloader config
+        for pruner's gradient sampling.
```
kylesayrs, line 136: nit: apostrophe

src/sparseml/pytorch/sparsification/pruning/modifier_pruning_mfac.py
```diff
@@ -229,6 +235,13 @@ def available_devices(self) -> Optional[List[str]]:
         """
         return self._available_devices
 
+    @ModifierProp(serializable=True)
+    def grad_sampler_kwargs(self) -> Optional[Dict[str, Any]]:
+        """
+        Return dict of training dataloader configs overriden for gradient sampling
```
kylesayrs, line 241: nit: overridden

src/sparseml/pytorch/sparsification/pruning/modifier_pruning_mfac.py
```diff
@@ -259,14 +272,21 @@ def initialize(
...
+            if (
+                "data_loader_builder" not in kwargs["grad_sampler"]
+                or "loss_function" not in kwargs["grad_sampler"]
+            ):
+                raise RuntimeError(
+                    "grad_sampler dict with data_loader_builder and loss_function "
+                    "must be provided to initialize GradSampler"
                 )
-            self._grad_sampler = grad_sampler
+            self._grad_sampler = GradSampler(
```
kylesayrs, line 283: Why not keep the same kwarg names as GradSampler(data_loader, loss_fn)? That way it's clearer to the user how to name their kwargs and you can pass GradSampler(**self._grad_sampler_kwargs) rather than passing them as args manually.

src/sparseml/pytorch/sparsification/pruning/modifier_pruning_mfac.py
```diff
@@ -130,6 +132,8 @@ class MFACPruningModifier(BaseGradualPruningModifier):
     :param mask_type: String to define type of sparsity to apply. May be 'unstructred'
         for unstructured pruning or 'block4' for four block pruning or a list of two
         integers for a custom block shape. Default is 'unstructured'
+    :param grad_sampler_kwargs: kwargs to override default train dataloader config
```
kylesayrs, line 135: It's very confusing to tell the difference between `grad_sampler_kwargs` and `grad_sampler`, since `grad_sampler_kwargs` is a list of kwargs for to pass to the data loader builder and `grad_sampler` is a list of kwargs that are used to build the data loader and a loss fn pass to the `GradSampler()`. If it has to stay this way, I propose renaming

`grad_sampler_kwargs` -> `data_loader_kwargs` or `grad_sampler_data_loader_kwargs`
`grad_sampler` -> `grad_sampler_kwargs`

Otherwise, I'd make the argument `grad_sampler_kwargs` which contains a key for the data loader, the data loader's arguments, and the loss fn

src/sparseml/transformers/sparsification/trainer.py
```diff
@@ -274,7 +264,10 @@ def create_optimizer(self):
                 wrap_optim=self.scaler,
                 loggers=self.logger_manager,
                 distillation_teacher=self.teacher,
-                grad_sampler=self.grad_sampler,
+                grad_sampler={
```
kylesayrs, line 267: There's a type change here, make sure this is should be grad_sampler and not grad_sampler_kwargs

kylesayrs (changes requested):

src/sparseml/pytorch/sparsification/pruning/modifier_pruning_mfac.py
eldarkurtic: Not sure if I understand the proposed idea, but `self._grad_sampler_kwargs` can't be used in the context `GradSampler(**self._grad_sampler_kwargs)` as it represents a dictionary of user-defined configs for gradient sampling procedure. `GradSampler` constructor expects an iterable dataloader and a callable loss function.

src/sparseml/pytorch/sparsification/pruning/modifier_pruning_mfac.py
eldarkurtic: @kylesayrs thanks for the review and useful comments!
In the initial versions of MFAC pruner class these gradient sampling kwargs were divided in two groups: one for dataloader and one for loss function. But there was a pushback as it was considered to be too complicated for end users to configure (don't know if these views have changed now). 

I agree that the naming should be improved, but I would try to keep `grad_sampler_kwargs` in the yaml recipe as changing it would require updating all existing recipes from oBERT runs. I would rather try to adopt a different naming for `grad_sampler = {"data_loader_builder": ..., "loss_function": ...}` dictionary. What do you think?

src/sparseml/transformers/sparsification/trainer.py
eldarkurtic: Yes this should be a dictionary with keys: `data_loader_builder` and `loss_function`, as we now delegate instantiating `GradSampler` class to the pruner's `initialize` procedure. That way we make it possible to override some of the default gradient-sampling configs. For example in upstream BERT runs, we don't want to use the default training batchsize=256 to sample gradients for pruning, and we want to set this to something much smaller via yaml recipe.

src/sparseml/pytorch/sparsification/pruning/modifier_pruning_mfac.py
bfineran: agreed on keeping the naming as is, especially given relationship between the recipe args, name and actual python object

bfineran (dismissed):

kylesayrs (approved):

bfineran (approved):

kylesayrs (approved):

PR #832, opened 2022-05-29
Implement OBSPruningModifier

eldarkurtic:
- Implementation of the pruning algorithm described in https://arxiv.org/abs/2203.07259
- Supports experiments on a single device and multiple-devices via DDP
- Optimized for unstructured and block4 version introduced in the paper
- Gradual pruning experiments on a single-device and multiple-devices via DDP are tested for both mask types (unstructured and block4), and the results match the numbers reported in the paper (+- stdev)
- With FP16 mode and the gradual pruning recipe described in the paper (12-layer BERT-base model, 1024 gradients, block-size 50, 12-layer BERT-base teacher), the experiment fits on a single 24GB RTX 3090 card
- the OBSPruningModifier can be used to prune other models too, it's not constrained to BERT models


bfineran (changes requested):
looks great overall @eldarkurtic. left a few nits to conform to codebase standards. let's make sure to have a plan to test the implementation vs the paper results

eldarkurtic:
@bfineran all code review updates are pushed; the only thing left is to figure out how to fix the pytorch-tests to handle kwargs when calling `initialize`. That should resolve the:
`RuntimeError: grad_sampler dict with data_loader_builder and loss_function must be provided to initialize GradSampler`

bfineran (approved):
LGTM. thank you @eldarkurtic! added a check for torch version compatibility

bfineran (approved):

src/sparseml/pytorch/sparsification/pruning/modifier_pruning_obs.py
```diff
@@ -0,0 +1,580 @@
...
+        fisher_block_size: int,
+        mask_type: str,
+    ):
+        super().__init__(params)
+        self._num_grads = num_grads
+        self._damp = damp
+        self._fisher_block_size = fisher_block_size
+        self._mask_type = mask_type
+
+        self._finvs = None  # type: List[EmpiricalBlockFisherInverse]
```
dbogunowicz, line 351: Why can't we directly attach the `type` to the variable?

src/sparseml/pytorch/sparsification/pruning/modifier_pruning_obs.py
```diff
@@ -0,0 +1,580 @@
...
+
+        # batched f_inv x g: (batch, B, B) x (batch, B) -> (batch, B)
+        finv_g = torch.einsum("bij,bj->bi", self.f_inv, g)
+
+        # scalar denominator for each batch: (batch)
+        alpha = (self.m + torch.einsum("bi,bi->b", g, finv_g)).sqrt().unsqueeze(1)
+        finv_g /= alpha
+
+        # update f_inv with new outer product: (batch, B) x (batch, B) -> (batch, B, B)
+        self.f_inv.baddbmm_(finv_g.unsqueeze(2), finv_g.unsqueeze(1), alpha=-1)
```
dbogunowicz, line 560: Can we change the method name `baddbmm` to something more verbose?

eldarkurtic: `baddbmm` is a method from PyTorch https://pytorch.org/docs/stable/generated/torch.baddbmm.html

src/sparseml/pytorch/sparsification/pruning/modifier_pruning_obs.py
eldarkurtic: do you mean like `self._finvs: List[EmpiricalBlockFisherInverse]` or something else?

bfineran: the `type:` in-line comment is used throughout sparseml. resolving comment

dbogunowicz (approved):

PR #842, opened 2022-06-02
Fix: allow non-string keys in recipes

KSGulin:
This fix adds back compatibility for numeric recipe keys, as seen in the following example:

```
target_sparsities: &target_sparsities
  0.90: ['sections.1.0.point.conv.weight', 'sections.1.1.point.conv.weight', 'sections.2.0.point.conv.weight']
  0.95: ['sections.2.1.point.conv.weight', 'sections.3.0.point.conv.weight', 'sections.3.1.point.conv.weight', 'sections.3.5.point.conv.weight']
  0.97: ['sections.3.2.point.conv.weight', 'sections.3.3.point.conv.weight', 'sections.3.4.point.conv.weight', 'sections.4.0.point.conv.weight', 'sections.4.1.point.conv.weight']
```

### DIFF
--- a/src/sparseml/optim/helpers.py
+++ b/src/sparseml/optim/helpers.py
@@ -348,7 +348,7 @@ def check_if_staged_recipe(container: dict) -> bool:
     for k, v in container.items():
         if isinstance(v, dict):
             if any(
-                [key for key in v.keys() if isinstance(key, str) and "modifiers" in key]
+                key for key in v.keys() if isinstance(key, str) and "modifiers" in key
             ):
                 return True
     return False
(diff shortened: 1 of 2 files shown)


bfineran (dismissed):

src/sparseml/optim/helpers.py
```diff
@@ -347,7 +347,9 @@ def check_if_staged_recipe(container: dict) -> bool:
     """
     for k, v in container.items():
         if isinstance(v, dict):
-            if any([key for key in v.keys() if "modifiers" in key]):
+            if any(
+                [key for key in v.keys() if isinstance(key, str) and "modifiers" in key]
```
rahul-tuli, line 351: nit:can be absolutely ignored, but would be good to take advantage of `any(...)` method's early stopping by using generator

```suggestion
                key for key in v.keys() if isinstance(key, str) and "modifiers" in key
```

rahul-tuli (dismissed):

bfineran (approved):

rahul-tuli (approved):

PR #849, opened 2022-06-06
[Integrations-Testing] Github actions workflow

KSGulin:
Adds a GHA workflow for integrations testing, with 2 possible triggers (or cadences, as we refer to them internally).
Pre-commit is the only cadence which checks for directories changed in code and only tests the corresponding integrations. The rest of the cadences test all integrations each time they're invoked. 

Commit - triggered on push to main or release branch
Pre-commit - triggered on pull request

Note: `integrations-testing-updates` is set as a trigger branch in this PR. This is for testing purposes and will be removed before `integrations-testing-updates` lands on `main`

### DIFF
--- a/.github/workflows/integrations-check.yaml
+++ b/.github/workflows/integrations-check.yaml
@@ -1,0 +1,81 @@
+name: Integrations Testing
+on: 
+  push:
+    branches:
+      - main
+      - 'release/*'
+  pull_request:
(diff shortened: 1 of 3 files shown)


bfineran (changes requested):
let's take out weekly and nightly for now then we're good to go

PR #861, opened 2022-06-10
Transformers refactor

natuan:
Currently it's not possible for the two trainers in sparseml transformers both extending a functionality from the upstream HuggingFace Trainer and sharing common code. This PR is to fix that, for now enabling them share saving best model after a specified epoch and removing unused column.

Additionally, create common training args for different training flow, and let them share distill teacher and recipe args. Same things could be done when needed for data and model args.

### DIFF
--- a/src/sparseml/transformers/question_answering.py
+++ b/src/sparseml/transformers/question_answering.py
@@ -49,8 +49,8 @@ from transformers.utils.versions import require_version
 
 from sparseml.transformers.sparsification import (
     QuestionAnsweringTrainer,
-    postprocess_qa_predictions,
     TrainingArguments,
+    postprocess_qa_predictions,
 )
 from sparseml.transformers.utils import SparseAutoModel, get_shared_tokenizer_src
 
@@ -89,10 +89,6 @@ class ModelArguments:
                 "huggingface.co/models"
             )
         }
-    )
-    distill_teacher: Optional[str] = field(
-        default=None,
-        metadata={"help": "Teacher model which needs to be a trained QA model"},
     )
     config_name: Optional[str] = field(
         default=None,
@@ -141,21 +137,6 @@ class DataTrainingArguments:
     Arguments pertaining to what data to input to our model for training and eval
     """
 
-    recipe: Optional[str] = field(
-        default=None,
-        metadata={
-            "help": (
-                "Path to a SparseML sparsification recipe, see "
-                "https://github.com/neuralmagic/sparseml for more information"
-            )
-        },
-    )
-    recipe_args: Optional[str] = field(
-        default=None,
-        metadata={
-            "help": "Recipe arguments to be overwritten",
-        },
-    )
     dataset_name: Optional[str] = field(
         default=None,
         metadata={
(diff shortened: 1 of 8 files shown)


src/sparseml/transformers/question_answering.py
```diff
@@ -771,8 +752,8 @@ def compute_metrics(p: EvalPrediction):
     trainer = QuestionAnsweringTrainer(
         model=model,
         model_state_path=model_args.model_name_or_path,
-        recipe=data_args.recipe,
-        recipe_args=data_args.recipe_args,
+        recipe=training_args.recipe,
```
KSGulin, line 755: Looks like this should be mimicked in the `masked_language_modeling.py` and `token_classification.py`

src/sparseml/transformers/sparsification/trainer.py
```diff
@@ -831,39 +832,25 @@ def _generate_apply_manager_params(self, kwargs) -> Tuple[Optional[str], float]:
         return checkpoint, epoch
 
 
-class Trainer(TrainerInterface, TransformersTrainer):
+class TransformersTrainer(HFTransformersTrainer):
```
KSGulin, line 835: Would be good to have an explicit init function here

src/sparseml/transformers/sparsification/trainer.py
```diff
@@ -831,39 +832,25 @@ def _generate_apply_manager_params(self, kwargs) -> Tuple[Optional[str], float]:
...
-        used to load config and tokenizer settings
-    :param recipe: the recipe, if any, to apply to the modle and training
-        process
-    :param recipe_args: A json string, csv key=value string, or dictionary containing
-        arguments to override the root arguments within the recipe such as
-        learning rate or num epochs
-    :param teacher: teacher model for distillation. Set to 'self' to distill
-        from the loaded model or 'disable' to turn of distillation
-    :param kwargs: key word arguments passed to the parent class
+    A transformers trainer class with customed behaviors that can be shared
```
KSGulin, line 837: nit
```suggestion
    A transformers trainer class with custom behavior that can be shared
```

KSGulin (commented):
Looks good @natuan. Just had a few comments

src/sparseml/transformers/question_answering.py
natuan: Fixed

src/sparseml/transformers/sparsification/trainer.py
rahul-tuli: QQ: What benefit would that have?

rahul-tuli:
Could we update the description with the code for loading best model for QA before landing?

rahul-tuli (dismissed):

src/sparseml/transformers/sparsification/trainer.py
natuan: Currently we're not creating any instance from this class.

anmarques (approved):
LGTM!

rahul-tuli (approved):

PR #879, opened 2022-06-16
finalize stage modifiers after stage end epoch in staged torch runs

bfineran:
currently in runs of composed staged recipes, modifier finalization only occurs at the end of the entire run (after all stages). this may cause issues because when a stage is composed through multiple runs, model states (and recipe assumptions) are built assuming the previous stage has been finalized. This PR introduces a change to finalize the modifiers in a stage immediately after the maximum epoch for the stage passes (on the start of the next epoch)


jeanniefinks:

PR #909, opened 2022-06-23
Upload data tar to tag and download upon integration install

KSGulin:
This PR adds a path for automated maintenance and pulldown of integration-specific data as tag assets.

The full lifecycle:

**Data upload**
- Trigger upon push to a release branch, only if data files are modified (currently, only yolov5 data files are checked)
- Determine the appropriate tag to upload to (e.g. push to release/0.12 - > tag = v.0.12.0)
- Check if the tag already exists. If it doesn't, skip the steps below
- Create a `tar.gz` file of the data files (right now only targets yolov5 data)
- Push the asset to the tag

**Data download**
- Upon installation of integration (again, only yolov5 supported), download data archive
- Uncompress it into the sparseml install directory

Notes:
- Data will only be stored on and downloaded from `major.minor` releases. Patches will not have data assets attached to them
- This PR includes a minor comment change to one of the data files, so that when it's cherry picked to the release branch, it will trigger the data upload

### DIFF
--- a/.github/workflows/upload-data-tar.yaml
+++ b/.github/workflows/upload-data-tar.yaml
@@ -2,11 +2,11 @@ name: Upload Data Tar
 
 on:
   push:
-    branches: ["include-data-files"]
+    branches: ["release/*"]
     paths:
-      - 'src/sparseml/yolov5/models'
-      - 'src/sparseml/yolov5/models_v5.0'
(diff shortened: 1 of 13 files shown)


KSGulin:
Closing this PR, as there is no good way to test it prior to release. Reverting to including yolov5 data files in the wheel installable.

PR #912, opened 2022-06-24
add TensorBoardLogger to transformers integration

bfineran:
adds a sparseml `TensorBoardLogger` to the transformers integration to enable tensorboard logging of sparsification modifier values.

implementation is currently setup so that `TensorBoardLogger` is only created if the default transformers tensorboard logging callback is setup and will share a logging directory with it so the sparsification and default transformers training values will both appear on the board.

default transformers tensorboard logged values include run arguments, epoch, loss, and learning rate.
the tensorboard logs are written to `--logging_dir`, so the tensorboard server can be launched with
```bash
tensorboard --logdir <LOGGING_DIR>
```

additionally investigating if there is any further information to be gained by setting `sync_tensorboard=True` in the `wandb` logger


dbogunowicz (approved):

src/sparseml/transformers/sparsification/trainer.py
```diff
@@ -154,7 +156,7 @@ def __init__(
         self.criterion = torch.nn.CrossEntropyLoss()
         self.callback_disable_fp16 = DisableHalfPrecisionCallback(self)
         self.callback_handler.add_callback(self.callback_disable_fp16)
-
+        self._add_tensorboard_logger_if_available()
```
corey-nm, line 159: Could you add TensorboardLogger to the list of loggers next to `[WANDBLogger()`] in the logger manager block above? Or is the issue that tensorboard is added to callbacks in some other piece of code, so you need to grab that instance from that list instead of instantiating a new one?

corey-nm (approved):

src/sparseml/transformers/sparsification/trainer.py
bfineran: we want to group it with some additional logic around the callback, so broke it into a separate function

PR #929, opened 2022-07-04
[WIP][SparseZoo v2 Bridge] Save yolo training artifacts to ModelDirectory directory

dbogunowicz:


src/sparseml/yolov5/helpers.py
```diff
@@ -0,0 +1,82 @@
...
+    (the directory where the pipeline saves its training artifacts),
+    and saves the training artifacts to `output_dir` in the `ModelDirectory` structure.
+
+    :param output_dir: The output path where the artifacts are saved
+        (adhering to the in `ModelDirectory` structure)
+    :param training_outputs_dir: The path to the existing directory
+        with the saved training artifacts
+    :param model_file_torch: name of the final .pth/.pt file to be saved
+    """
+    training_outputs_files = glob.glob(os.path.join(training_outputs_dir, "*"))
```
bfineran, line 41: nit: use `os.listdir(training_outputs_dir)`

src/sparseml/yolov5/helpers.py
```diff
@@ -0,0 +1,82 @@
...
+    logs_path = [
+        file
+        for file in training_outputs_files
+        if os.path.basename(file).startswith("events.out.")
+    ]
+    for root_file in ["sample_inputs.tar.gz", "sample_outputs.tar.gz"]:
+        root_file_path = os.path.join(training_outputs_dir, root_file)
+        if not os.path.exists(root_file_path):
+            raise ValueError(
```
bfineran, line 50: let's make this a warning instead - required for research/zoo but shouldn't block for now

src/sparseml/yolov5/helpers.py
```diff
@@ -0,0 +1,82 @@
...
+                f"File {root_file_path} missing. To create this file, "
+                "make sure that the training script is being ran with"
+                "`--num_export_samples` argument."
+            )
+
+    model_onnx_path = os.path.join(
+        training_outputs_dir, "weights", model_file_torch.replace(".pt", ".onnx")
```
bfineran, line 57: what is the purpose here/what are the potential `onnx` names here?

bfineran (changes requested):
LGTM pending comments and resolution on the error in sample function output

src/sparseml/yolov5/helpers.py
dbogunowicz: Yeah, I need to suppress my unproductive `glob.glob(os.path.join(whathever, "*"))` habit.

src/sparseml/yolov5/helpers.py
dbogunowicz: Very good question. This is something that also got me confused - so I decided to give more flexibility to the user.

When you look at the directory with the training artifacts:
image
There are two types of models in `/weights`: `best.pt` and `last.pt`. Being unsure whether the potential ML researcher would choose one of them by default, I gave a possibility to point at the one which should eventually make it to the `ModelDirectory`.  But maybe we are always choosing one over the other? Happy to get your input on that @bfineran .

src/sparseml/yolov5/helpers.py
```diff
@@ -0,0 +1,81 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import os
+
+from sparsezoo.v2.helpers import setup_model_directory
```
dbogunowicz, line 18:
```suggestion
from sparsezoo import setup_model
```

src/sparseml/yolov5/helpers.py
```diff
@@ -0,0 +1,81 @@
...
+
+def get_model_directory(
+    output_dir: str,
+    training_outputs_dir: str,
+    model_file_torch: str,
+) -> None:
+    """
+    Takes the `training_outputs_dir`
+    (the directory where the pipeline saves its training artifacts),
+    and saves the training artifacts to `output_dir` in the `ModelDirectory` structure.
```
dbogunowicz, line 32:
```suggestion
    and saves the training artifacts to `output_dir` in as the sparsezoo`Model` class object.
```

dbogunowicz, line 32:
```suggestion
    and saves the training artifacts to `output_dir` as a sparsezoo Model class object.
```

src/sparseml/yolov5/helpers.py
```diff
@@ -0,0 +1,81 @@
...
+        file
+        for file in training_outputs_files
+        if os.path.basename(file).startswith("events.out.")
+    ]
+    for root_file in ["sample_inputs.tar.gz", "sample_outputs.tar.gz"]:
+        root_file_path = os.path.join(training_outputs_dir, root_file)
+        if not os.path.exists(root_file_path):
+            logging.warning(
+                f"File {root_file_path} missing. To create this file, "
+                "make sure that the training script is being ran with"
```
dbogunowicz, line 51:
```suggestion
                "make sure that the export script is ran with"
```

dbogunowicz, line 51:
```suggestion
                "make sure that the export script has been ran with"
```

src/sparseml/yolov5/helpers.py
```diff
@@ -0,0 +1,81 @@
...
+        training_outputs_dir, "weights", model_file_torch.replace(".pt", ".onnx")
+    )
+    if not os.path.exists(model_onnx_path):
+        raise ValueError(
+            f"File {model_onnx_path} missing. To create this file, "
+            "make sure that the `export` script (for exporting "
+            "yolo model) has been evoked."
+        )
+
+    setup_model_directory(
```
dbogunowicz, line 65:
```suggestion
    setup_model(
```

bfineran (approved):

KSGulin (dismissed):

src/sparseml/yolov5/helpers.py
bfineran: Good catch, however we shouldn't be exporting both. Either best or last will be exported to model.onnx, would recommend removing this functionality

dbogunowicz: this is why I leave it as an argument, to choose `best `or `last` :D

KSGulin (approved):

src/sparseml/yolov5/helpers.py
bfineran: we should just be exporting `model.onnx` though right?

bfineran: as long as the deployment directory ends up with `model.onnx` (and can keep existing behavior outside of this) we should be good

bfineran (approved):

bfineran (approved):

rahul-tuli (approved):

PR #932, opened 2022-07-06
Add data download scripts to yolact integration

rahul-tuli:
This PR adds data downloading scripts to sparseml's yolact integration;

Recommended usage:
```bash
sh COCO.sh [OPTIONAL_DATA_DIRECTORY]
```

The path to the downloaded train + validation info(json) and images can then be passed on to the respective train scripts.

### DIFF
--- a/src/sparseml/yolact/COCO_test.sh
+++ b/src/sparseml/yolact/COCO_test.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+
+start=`date +%s`
+
+# handle optional download dir
+if [ -z "$1" ]
+  then
+    # navigate to ./data
+    echo "navigating to ./data/ ..."
+    mkdir -p ./data
+    cd ./data/
+    mkdir -p ./coco
+    cd ./coco
+    mkdir -p ./images
+    mkdir -p ./annotations
+  else
+    # check if specified dir is valid
+    if [ ! -d $1 ]; then
+        echo $1 " is not a valid directory"
+        exit 0
+    fi
+    echo "navigating to " $1 " ..."
+    cd $1
+fi
+
+if [ ! -d images ]
+  then
+    mkdir -p ./images
+fi
+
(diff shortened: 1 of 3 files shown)


src/sparseml/yolact/COCO_test.sh
```diff
@@ -0,0 +1,62 @@
...
+fi
+
+# Download the annotation data.
+cd ./annotations
+echo "Downloading MSCOCO test info ..."
+curl -LO http://images.cocodataset.org/annotations/image_info_test2017.zip
+echo "Finished downloading. Now extracting ..."
+
+# Unzip data
+echo "Extracting train images ..."
```
dbogunowicz, line 49:
```suggestion
echo "Extracting test images ..."
```

dbogunowicz (commented):
Any reason for having two scripts; one for train and val and one for test?
I guess there is a possibility that the user would like only to train (therefore only train and val), but I would probably opt for having one script to pull everything in one go.

src/sparseml/yolact/COCO_test.sh
rahul-tuli: These are the original scripts from DBolya, I think separate scripts provide logical separation b/w training and test times, but I don't have a strong argument for either approaches.

rahul-tuli:
> Any reason for having two scripts; one for train and val and one for test? I guess there is a possibility that the user would like only to train (therefore only train and val), but I would probably opt for having one script to pull everything in one go.

These are the original scripts from DBolya, I think separate scripts provide logical separation b/w training and test times, but I don't have a strong argument for either approaches

bfineran (changes requested):
let's make sure to add `src/sparseml/yolact/*.sh` to `MANIFEST.in`

should be good after this, correct @KSGulin?

bfineran (approved):

PR #940, opened 2022-07-13
Enable QATWrapper on modules that contain weights

anmarques:
PR motivated by mobileBERT. It requires using QATWrapper on a Linear module. Previous QATWrapper didn't allow weights of wrapped module to be quantized (only inputs and outputs). This PR applies standard weight quantization to any wrapped module by defining a qconfig for the wrapped module.


mgoin (approved):

src/sparseml/pytorch/sparsification/quantization/helpers.py
```diff
@@ -256,6 +256,13 @@ def __init__(
         num_input_quant_stubs = num_inputs + len(self.kwarg_input_names)
 
         self.forward_fn = forward_fn
+        # Add weight qconfig to forward_fn (in case it has weights)
+        qconfig_ = get_qat_qconfig(qproperties)
+        qconfig = torch_quantization.QConfig(
+            activation=torch.nn.Identity,
+            weight=qconfig_.weight,
+        )
+        self.forward_fn.qconfig = qconfig
```
natuan, line 265: @anmarques Would it be better to have the new QATEmbeddingTransformation implement the `configure_qconfig()` method with its own `qconfig` above? That way, if I understand correctly, this layer will be attached with its appropriate qconfig by following the main flow here: https://github.com/neuralmagic/sparseml/blob/main/src/sparseml/pytorch/sparsification/quantization/modifier_quantization.py#L653

Right now I feel that it's ugly that all module will have the above special qconfig, and only later on will they be changed back to the default (and then some might get their customed config back). By design they should have empty qconfig at this stage.

natuan (changes requested):

src/sparseml/pytorch/sparsification/quantization/helpers.py
```diff
@@ -256,6 +256,13 @@ def __init__(
         num_input_quant_stubs = num_inputs + len(self.kwarg_input_names)
 
         self.forward_fn = forward_fn
+        # Add weight qconfig to forward_fn (in case it has weights)
+        qconfig_ = get_qat_qconfig(qproperties)
+        qconfig = torch_quantization.QConfig(
```
natuan, line 261: Since qconfig_ is already of type torch_quantization.QConfig, would it be simpler just to set its activation to identity?

natuan (changes requested):

src/sparseml/pytorch/sparsification/quantization/helpers.py
natuan: Are you sure the wrapped forward_fn will not have qconfig? I think they will here: https://github.com/neuralmagic/sparseml/blob/main/src/sparseml/pytorch/sparsification/quantization/modifier_quantization.py#L649

The QATWrapper does not attach qconfig to forward_fn because it's designed for inputs and outputs only (as seen in its doc and constructor's signature). IMO we should keep the design clean that way and following the recipe that customed modules implements their own qconfig.

src/sparseml/pytorch/sparsification/quantization/helpers.py
anmarques: Attempted that and the activation cannot be changed after it's defined: AttributeError: can't set attribute

src/sparseml/pytorch/sparsification/quantization/helpers.py
anmarques: Not sure why PyTorch does not propagate qconfig for forward_fn, but it doesn't. If this were true a wrapped Linear module would quantize it's weights, but that doesn't happen. That's the main motivation for this PR.

natuan: My main interest here is trying to keep the code clean if possible by letting your custom module defines its own qconfig, and then things will work (w/o having to mess up the QATWrapper). Otherwise I'm concerned that if we have to do the same for other modules in the future, we wouldn't have a good way/example to follow.

natuan (changes requested):

src/sparseml/pytorch/sparsification/quantization/helpers.py
natuan: ok, thanks for trying

src/sparseml/pytorch/sparsification/quantization/helpers.py
natuan: What we agreed on after discussion:
- Verify to make sure this PR won't affect other models (BERT, DistilBERT)
- If the qconfig is correctly propagated down to the linear module of the QATEmbeddingTransformation, we probably wouldn't need this PR. It's still not clear as to why this is not working as expected
- The configure_qconfig method from QATEmbeddingTransformation, hence transformers side, seems hard to do without making transformers depends on sparseml (since we would need sparseml.QConfigProperty).

natuan (approved):

bfineran (approved):

PR #946, opened 2022-07-19
Setting max python version in setup.py to 3.9.0

corey-nm:
This will give an error like:
```
ERROR: Package 'sparseml-nightly' requires a different Python: 3.10.5 not in '<=3.9.0,>=3.6.0'
```
if someone tries to install with unsupported python version


corey-nm:
Closing in preference to #947

PR #949, opened 2022-07-21
Option to eval w/o amp during training

natuan:
The evaluation in transformers currently still runs with fp16 during QAT even though this mode is turned off. This creates bigger gap in accuracy compared to deepsparse, especially for small datasets such as SST2. This PR is to provide the option `--no_amp_eval` to disable mixed precision during evaluation (default False for not affecting existing training commands).
Qualification: used the flag during QAT and separate evaluation, checked use_amp status, checked model accuracy w/ vs w/o the flag.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -7,8 +7,8 @@ FROM nvidia/cuda:${CUDA_VERSION}-devel-ubuntu18.04 as base
 # As of 05/05/22 nvidia images are broken. Two lines below are a temporary fix.
 # Source: https://github.com/NVIDIA/nvidia-docker/issues/1632
 
-RUN rm /etc/apt/sources.list.d/cuda.list
-RUN rm /etc/apt/sources.list.d/nvidia-ml.list
+RUN if [ -f /etc/apt/sources.list.d/cuda.list ] ; then rm /etc/apt/sources.list.d/cuda.list ; fi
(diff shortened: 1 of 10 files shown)


dbogunowicz (commented):
Looking good @natuan , could you provide short manual testing example in the PR description?

natuan:
> Looking good @natuan , could you provide short manual testing example in the PR description?

Updated with testing

bfineran (commented):
LGTM overall. discussed this with @anmarques yesterday, but what are your thoughts on making no amp eval the default behavior for consistency then providing an option to disable the disable

bfineran (commented):
can we keep in an arg to disable this? I don't think we'd need to do fp16 eval, but since this is transformers default, we should probably allow it?

natuan:
> can we keep in an arg to disable this? I don't think we'd need to do fp16 eval, but since this is transformers default, we should probably allow it?

There's already the option from transformers `fp16_full_eval` for that purpose

bfineran:
awesome, and that overrides these changes?

natuan:
> awesome, and that overrides these changes?

natuan:
> > awesome, and that overrides these changes?

Good catch. I've updated the code to take existing flags into account

rahul-tuli (approved):

corey-nm (approved):

PR #951, opened 2022-07-21
Moving `integrations/pytorch/notebooks` to `examples/`

corey-nm:
These are really good examples of how to use sparseml. Moving them under `examples/` makes them much easier to find for users.

Are there other nice scripts/notebooks/showcases that can go under examples?


bfineran (changes requested):
we've intentionally migrated our examples all into integrations to limit the number of entrypoints. open to taking a look at this again maybe when we add in more/update examples, but let's close for now

PR #966, opened 2022-08-03
Fix TrainableParamsModifier

### LINKED ISSUE #965: TrainableParamsModifier doesn't freeze params
**Describe the bug**
When `TrainableParamsModifier` is used to freeze params (i.e. to set their `requires_grad=False`), the parameters are not frozen and optimizer is updating them. 
This happens because the `TrainableParamsModifier` is called for the first time to set `requires_grad=False` in the `optimizer.step()` call, but at this point it's already late to set it. The forward-backward pass happened and the PyTorch autograd recorded all these params (as by default they had `requires_grad=True`) and attached `.grad` to them. This will be a signal for optimizer to optimize those params too. 

**Expected behavior**
To have these params frozen and not updated during the run.

**To Reproduce**
Run any experiment with a recipe that includes `TrainableParamsModifier` and compare the weights before the run and after a few steps.

eldarkurtic:
Fixes: https://github.com/neuralmagic/sparseml/issues/965

### DIFF
--- a/src/sparseml/pytorch/sparsification/training/modifier_params.py
+++ b/src/sparseml/pytorch/sparsification/training/modifier_params.py
@@ -126,6 +126,8 @@ class TrainableParamsModifier(BaseTrainableParamsModifier, ScheduledModifier):
         for layer_name, layer, param_name, param in layers_names_and_params:
             self._module_params.append(param)
 
+        self._check_update(epoch, steps_per_epoch=1)
+
     def update(
         self, module: Module, optimizer: Optimizer, epoch: float, steps_per_epoch: int
     ):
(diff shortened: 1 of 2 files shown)


src/sparseml/pytorch/sparsification/training/modifier_params.py
```diff
@@ -149,6 +149,8 @@ def update(
             for param in self._module_params:
                 self._original.append(param.requires_grad)
                 param.requires_grad = self._trainable
+                if not self._trainable:
+                    param.grad = None  # clear .grad to prevent optimizer updates
```
natuan, line 153: you mind put a short explanation here as to why we need this extra step (that the (Adam) optimizer looks at .grad and ignores requires_grad when updating params)?

natuan (dismissed):

src/sparseml/pytorch/sparsification/training/modifier_params.py
```diff
@@ -142,16 +144,23 @@ def update(
...
-            self._original.clear()
+    def _enable(self, param: Parameter):
+        param.requires_grad_(True)
+
+    def _disable(self, param: Parameter):
+        param.requires_grad_(False)
+        param.grad = None  # clear to prevent optimizer updates
 
+    def _check_update(self, epoch: float, steps_per_epoch: int):
+        if self.start_pending(epoch, steps_per_epoch) and not self._original:
```
natuan, line 157: I'm curious: do we actually need the `not self._original` condition?

eldarkurtic: Yes we do for cases when modifier kicks in immediately during `def initialize(...)` (e.g. gradual pruning with `start_epoch=0`). `not self._original` prevents enabling/disabling again when `def update(...)` is called for the first time in the gradual run.

natuan (approved):

KSGulin (approved):
Great catch!

PR #968, opened 2022-08-04
Fix qat convert mobilebert

anmarques:
This PR introduces changes to QAT export to support mobileBERT:
1. Moves ReLU folding to after linear modules processing (it was after conv processing before)
2. Moves embedding quantization up in the list of conversions (does not affect other conversions)
3. Adds a conversion to propagate a dequantize linear module down through a gather op. MobileBERT has embedding quantizations followed by a gather and linear module. This propagation allows the quantization to be picked up by the linear module


src/sparseml/pytorch/sparsification/quantization/quantize_qat_export.py
```diff
@@ -1719,3 +1720,116 @@ def skip_onnx_input_quantize(
...
+    |         |   |   |
+    |         | Slice Slice
+    |         |   |   |
+    |         |  Pad Pad
+    |         |   |   |
+    |           Concat
+    |             |
+    |       DequantizeLinear
+    |             |
+    |           OUTPUT
```
mgoin, line 1752: note that the engine can do something like this currently for ops like reshapes and transposes, would be interesting to see if we could handle this as well

mgoin (approved):

rahul-tuli (dismissed):

src/sparseml/pytorch/sparsification/quantization/quantize_qat_export.py
anmarques: If I understand it correctly, the issue w/ that is that the exported onnx would not convert the Linear module that follows the embeddings w/ a QMatMul.

In the ONNX export the above conversion precedes the Linear conversion to QMatMul, and the Linear conversion needs the dequantize node to be directly at the input. So this has to be done at the export level.

rahul-tuli (approved):
LGTM!

mgoin (approved):

bfineran (approved):

Issue #970, opened 2022-08-05
Manager miscalculates start_epoch when composing stages (recipes)


eldarkurtic:
**Describe the bug**
Some modifiers (for example DistillationModifier) have `start_epoch=-1.0` and when composed with the previous stage with `start_epoch += previous_stage_end_epoch`, the `start_epoch` will be off by -1.0.

PR #971, opened 2022-08-05
Set start_epoch to zero when -1.0 before composing

eldarkurtic:
Fixes https://github.com/neuralmagic/sparseml/issues/970


bfineran (approved):

bfineran:
good fix!

natuan (approved):

PR #964, opened 2022-08-02
Draft: Supporting torch 1.12

corey-nm:
Todos:

- Guard specific torch function calls in torch version checks:
    - [ ] fuse_modules
    - [ ] onnx.export 
- [ ] Get all the unit tests passing (there are some that still fill due to onnx node name changes)

PR #979, opened 2022-08-09
Upgrading ubuntu in workflows and setting python to 3.9

corey-nm:
This is necessary for upgrading to torch 1.12. Originally done in #964 , but moving these changes to a separate PR to make reviewing easier

### DIFF
--- a/.github/workflows/Integrations-post-merge-check.yaml
+++ b/.github/workflows/Integrations-post-merge-check.yaml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   test-setup:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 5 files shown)


rahul-tuli (approved):

KSGulin (commented):
Thoughts of updating to ubuntu 22.04 rather than latest, so that we have a static test environment? Or do we reap more benefit from staying up to date with the latest image?

corey-nm:
> Thoughts of updating to ubuntu 22.04 rather than latest, so that we have a static test environment? Or do we reap more benefit from staying up to date with the latest image?

Good idea, I think static test env would gain us more and be slightly easier to reproduce/debug. I'll change to 22.04

KSGulin (dismissed):
LGTM 😃

corey-nm:
Hmm may need to downgrade to 20.04 for package versioning. Looks like python 3.8 (keras) and 3.7 (for tf 1) aren't available via actions/setup-python.

KSGulin (approved):

rahul-tuli (approved):

PR #980, opened 2022-08-09
Using torch.linalg.vector_norm over norm

corey-nm:
Progress on upgrades to torch 1.12. Main pr is #964.

Testing was done by running unit tests in python 3.10 and torch 1.12.

### DIFF
--- a/src/sparseml/pytorch/sparsification/distillation/modifier_distillation.py
+++ b/src/sparseml/pytorch/sparsification/distillation/modifier_distillation.py
@@ -18,6 +18,7 @@ Modifier for performing model distillation
 
 
 import logging
+from collections.abc import Mapping
 from copy import deepcopy
 from typing import Any, Dict, Iterable, List, Optional, Union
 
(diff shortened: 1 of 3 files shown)


rahul-tuli (dismissed):

src/sparseml/pytorch/sparsification/distillation/modifier_distillation.py
```diff
@@ -388,7 +387,7 @@ def _kldiv_output_loss(self, student_outputs, teacher_outputs):
             distill_head_output_losses.append(
                 self._calc_distill_head_output_loss(student_outputs, teacher_outputs)
             )
-        elif isinstance(student_outputs, Mapping):
+        elif isinstance(student_outputs, Dict):
```
KSGulin, line 390: What's the motivation for restricting the type accepted here?

corey-nm: `Mapping` was moved to `from collections.abc import Mapping` in python 3.10. Wasn't sure if we wanted to do weird import stuff depending on python version or just change this check.

Do you know what other types are handled by this elif branch?

corey-nm: Actually looking more into it apparnetly it was deprecated in 3.3, so I think I can actually just change the import and it should be fine

KSGulin (approved):

rahul-tuli (approved):

PR #982, opened 2022-08-10
Freeze setuptools at 59.5.0

corey-nm:
To address issue discussed in https://github.com/pytorch/pytorch/issues/69894#issuecomment-1080635462

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ version_nm_deps = f"{version_major_minor}.0"
 _PACKAGE_NAME = "sparseml" if is_release else "sparseml-nightly"
 
 _deps = [
-    "setuptools==59.5.0",
+    "setuptools<=59.5.0",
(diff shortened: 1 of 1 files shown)


bfineran (commented):
Can we add a more generic dependency rather than pinning?

rahul-tuli:
I agree with Ben!

corey-nm:
Sure I can make `<=59.5.0`

rahul-tuli (approved):
LGTM! As a side note could we also check if a later/latest version of setup tools fixes the bug.

corey-nm:
> LGTM! As a side note could we also check if a later/latest version of setup tools fixes the bug.

The issue started with setuptools==59.6.0. It's fixed in later pytorch versions

KSGulin (approved):

PR #990, opened 2022-08-12
don't save epoch on IC one_shot checkpoints

bfineran:
default saved epoch for `one_shot` in the IC flows is `-1` due to `Trainer` initialization. This will cause issues on model load since the checkpoint recipe will be initialized to epoch `-1`, not applying any of the optimizations. This PR removes saving the epoch for one shot models so the entire one shot recipe will be applied on checkpoint load.

**test_plan:**
@anmarques to verify

### DIFF
--- a/src/sparseml/pytorch/image_classification/utils/helpers.py
+++ b/src/sparseml/pytorch/image_classification/utils/helpers.py
@@ -409,7 +409,7 @@ def save_model_training(
     manager: BaseManager,
     save_name: str,
     save_dir: str,
-    epoch: int,
+    epoch: Optional[int],
     val_res: Optional[ModuleRunResults],
     checkpoint_manager: Optional[BaseManager] = None,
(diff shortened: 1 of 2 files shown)


anmarques (approved):
Verified in a one-shot flow for quantization

src/sparseml/pytorch/image_classification/utils/helpers.py
```diff
@@ -420,7 +420,7 @@ def save_model_training(
     :param manager: manager created from the training recipe
     :param save_name: name to save model to
     :param save_dir: directory to save results in
-    :param epoch: integer representing umber of epochs to
+    :param epoch: integer representing epoch at which model is saved at
```
kylesayrs, line 423: nit: specify what `None` does

kylesayrs (approved):

rahul-tuli (approved):

PR #992, opened 2022-08-15
Add gradient accumulation for IC

KSGulin:
This PR adds the CLI arg `gradient_accum_steps` to IC, allowing control over gradient accumulation.

**Manual Testing**
main branch run (baseline). Top1 acc: 97.39998626708984
```
sparseml.image_classification.train     --recipe-path zoo:cv/classification/resnet_v1-50/pytorch/sparseml/imagenette/pruned-conservative?recipe_type=original     --pretrained True     --arch-key resnet50     --dataset imagenette     --dataset-path /home/konstantin/.fastai/data/imagenette2      --train-batch-size 128 --test-batch-size 256     --loader-num-workers 8     --model-tag resnet50-imagenette-pruned-conservative
```

feature/grad-accum-ic run, base command. Top1 acc: 98.20000457763672
```
sparseml.image_classification.train     --recipe-path zoo:cv/classification/resnet_v1-50/pytorch/sparseml/imagenette/pruned-conservative?recipe_type=original     --pretrained True     --arch-key resnet50     --dataset imagenette     --dataset-path /home/konstantin/.fastai/data/imagenette2      --train-batch-size 128 --test-batch-size 256     --loader-num-workers 8     --model-tag resnet50-imagenette-pruned-conservative
```

feature/grad-accum-ic run, with gradient accumulation. Top1 acc: 98.19998168945312 
```
sparseml.image_classification.train     --recipe-path zoo:cv/classification/resnet_v1-50/pytorch/sparseml/imagenette/pruned-conservative?recipe_type=original     --pretrained True     --arch-key resnet50     --dataset imagenette     --dataset-path /home/konstantin/.fastai/data/imagenette2      --train-batch-size 32 --test-batch-size 256     --loader-num-workers 8     --model-tag resnet50-imagenette-pruned-conservative --gradient_accum_steps 4
```

**Automated Testing**
Added relevant checks to `ModuleTrainer` unit tests


src/sparseml/pytorch/utils/module.py
```diff
@@ -975,6 +976,7 @@ def _runner_batch(
 
             # loss calculation
             losses = self._loss(data, pred)
+            losses[DEFAULT_LOSS_KEY] /= self._num_accumulated_batches
```
corey-nm, line 979: is this a bugfix since it didn't divide before?

src/sparseml/pytorch/utils/module.py
```diff
@@ -987,13 +989,17 @@ def _runner_batch(
...
+            self._accumulated == self._num_accumulated_batches
+            or self._accumulated == counter_len
+        ):
             if self.device_context.use_mixed_precision:
                 self._scaler.step(self._optimizer)
                 self._scaler.update()
             else:
                 self._optimizer.step(closure=self._optim_closure)
             self._accumulated = 0
+            self._optimizer.zero_grad()
```
corey-nm, line 1002: this changes the order of step/zero_grad right? before zero_grad would happen before step, and now it seems like it happens after it?

src/sparseml/pytorch/utils/module.py
KSGulin: Yes. Gradient accumulation was in essence a ghost feature before, as it wasn't utilized or exposed to the user. But this does fix the logic. Also since your comment I've made an update to reset the loss to the un-normalized value after gradient calculations. This only impacts the value of the loss logged and makes sure it's printed correctly

src/sparseml/pytorch/utils/module.py
KSGulin: I'm not sure why it is currently happening before. For `_num_accumulated_batches > 1` this would lead to issues, since it would clear the gradients from the previous updates before they were used to update. Maybe the check was meant to be `if self._accumulated == (self._num_accumulated_batches + 1):`. I suppose there might be a potential use for the gradients between `_runner_batch()` calls. I'll update to clear gradients at the beginning

bfineran (approved):

dbogunowicz (approved):
lgtm

PR #1003, opened 2022-08-18
Document classification with IMDB dataset 

natuan:
This PR extends the text classification training flow to allow for IMDB dataset.

The additional changes inside optim/manager, optim/modifier and modifier_quantization are for the composed recipe at the end of quantization, needed for separating the pruning and quantization into two training stages. (They're posted in this PR https://github.com/neuralmagic/sparseml/pull/987)


spacemanidol (approved):
looks good!

KSGulin (approved):

PR #1008, opened 2022-08-19
Yolact doc updates

rahul-tuli:
This PR removes the old unsupported pathway to create your own baseline Yolact model from a `darknet53.pth` backbone. Instead users are recommended to leverage and use yolact models(including baseline) from sparsezoo. The docs have been updated accordingly


bfineran (approved):
please confirm testing of commands before landing

kylesayrs (approved):

PR #1019, opened 2022-08-23
Update torch versions to support 1.9.1+cuda

KSGulin:
Currently, torch 1.9.1+cuda... and torchvision 0.10.1+cuda... do not fall within the torch requirements for sparseml. This will cause in an error thrown when another library has sparseml as a dependency, as sparsify.auto does. This PR adds a simple conditional change to include these unsupported versions

**Testing**
Running various sparsify.auto commands, as an example: `sparsify.auto --task yolov5 --dataset coco128.yaml`

Running within sparseml the command `pip install -e .[torchvision]` and confirming versions installed are
```
torch                         1.9.1
torchvision                   0.10.1
```

Sanity checks through python console:
```
>>> from packaging import version
>>> version.parse('1.9.1+cu111') <= version.parse('1.9.1')
False
>>> version.parse('1.9.1+cu111') < version.parse('1.9.2')
True
```


corey-nm (approved):
Super clean fix for this, nice!

bfineran (approved):

mgoin (approved):

PR #1029, opened 2022-08-31
Fix paths for training_args.py

eldarkurtic:
Tiny fix for paths to the `training_args.py` script.


natuan (approved):

rahul-tuli (approved):
And the best PR award for the second time in a row goes to @eldarkurtic

eldarkurtic:
Gotta fix 'em all!

PR #1039, opened 2022-09-15
Fix max_train_samples, stratified dataset split

natuan:
Fix max_train_samples, max_eval_samples for train/val split from using validation_ratio.

Fix stratified split (used currently for IMDB) when the datasets package does not support this feature. Keep the original code for reproducibility of shipped IMDB models.

This is to resolve the ticket: https://app.asana.com/0/1201652400520766/1202932644760687/f

### DIFF
--- a/src/sparseml/transformers/text_classification.py
+++ b/src/sparseml/transformers/text_classification.py
@@ -35,6 +35,7 @@ import datasets
 import numpy as np
 import transformers
 from datasets import load_dataset, load_metric
+from sklearn.model_selection import StratifiedShuffleSplit
 from transformers import (
     AutoConfig,
     AutoTokenizer,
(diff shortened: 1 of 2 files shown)


src/sparseml/transformers/text_classification.py
```diff
@@ -35,6 +35,7 @@
 import numpy as np
 import transformers
 from datasets import load_dataset, load_metric
+from sklearn.model_selection import StratifiedShuffleSplit
```
rahul-tuli, line 38: Is `sklearn` installed as a part of our transformers fork? if not could we add it to `setup.py`

src/sparseml/transformers/text_classification.py
```diff
@@ -674,19 +667,26 @@ def _create_train_dataset(raw_datasets, max_train_samples):
...
-                )
+            train_dataset = (
+                raw_datasets["train"] if train_dataset is None else train_dataset
+            )
             train_dataset, eval_dataset = _split_train_val(
                 train_dataset, data_args.validation_ratio
             )
+            if (
+                data_args.max_train_samples is not None
+                and len(train_dataset) > data_args.max_train_samples
```
rahul-tuli, line 678: Should we also support cases where `len(train_dataset) < data_args.max_train_samples` ?

rahul-tuli: Or would that qualify as a usage error

natuan: No (if max_train_samples >= len of the dataset then we just keep using it)

src/sparseml/transformers/text_classification.py
natuan: Added scikit-learn to setup.py

rahul-tuli (approved):

anmarques (approved):

PR #1047, opened 2022-09-26
Update recipe arg overwrite to support list types

KSGulin:
Currently, utilizing `--recipe_args` with a recipe that has a [2, 4] mask or a list-valued entry for any of the fields, will result in an error. This is because the recipe update logic makes the assumption that any list-valued entry is a list of modifiers (represented by dictionaries). This PR fixes this with a minor logic change.

**Testing**
```bash
sparseml.yolov5.train --cfg models_v5.0/yolov5l.yaml --recipe-args '{\"quantization_start_epoch\":1}' --recipe recipes/test_recipe.md
```
Where `recipes/test_recipe.md` is

```yaml
---
# General Hyperparams
init_lr: 0.01
weights_warmup_lr: 0

# Pruning Hyperparams
pruning_update_frequency: 0.5
mask_type: [1, 4]

# Quantization Hyperparams
quantization_start_epoch: 120

training_modifiers:
    start_epoch: 0
    end_epoch: 3
    lr_func: linear
    init_lr: eval(weights_warmup_lr)
    final_lr: eval(init_lr)
    param_groups: [0, 1]
        
quantization_modifiers:
  - !QuantizationModifier
    start_epoch: eval(quantization_start_epoch)
---
```

### DIFF
--- a/src/sparseml/optim/helpers.py
+++ b/src/sparseml/optim/helpers.py
@@ -255,7 +255,7 @@ def _update_recipe_variable(var_name: str, var_value, container):
     key_found = False
 
     for key, value in container.items():
-        if not isinstance(value, list) and not all(
+        if not isinstance(value, list) or not all(
             [isinstance(val, dict) for val in value]
         ):
             if var_name == key:
--- a/src/sparseml/optim/helpers.py
+++ b/src/sparseml/optim/helpers.py
@@ -256,7 +256,7 @@ def _update_recipe_variable(var_name: str, var_value, container):
 
     for key, value in container.items():
         if not isinstance(value, list) or not all(
-            [isinstance(val, dict) for val in value]
+            isinstance(val, dict) for val in value
         ):
             if var_name == key:
                 container[var_name] = var_value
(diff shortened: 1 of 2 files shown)


bfineran (dismissed):

src/sparseml/optim/helpers.py
```diff
@@ -255,7 +255,9 @@ def _update_recipe_variable(var_name: str, var_value, container):
     key_found = False
 
     for key, value in container.items():
-        if not isinstance(value, list):
+        if not isinstance(value, list) or not all(
+            [isinstance(val, dict) for val in value]
```
rahul-tuli, line 259:
```suggestion
            isinstance(val, dict) for val in value
```

rahul-tuli (dismissed):
Go `Nittany Lions`

src/sparseml/optim/helpers.py
```diff
@@ -255,7 +255,9 @@ def _update_recipe_variable(var_name: str, var_value, container):
     key_found = False
 
     for key, value in container.items():
-        if not isinstance(value, list):
+        if not isinstance(value, list) or not all(
```
corey-nm, line 258: Can we add some small unit tests for this just to be sure that this functionality isn't lost in the future?

KSGulin: Added coverage to existing unit tests and confirmed it fails on main without this PR

bfineran (approved):

corey-nm (approved):

PR #1051, opened 2022-09-28
Implement invHess recomputations for OBS

eldarkurtic:
This feature improves OBSPruningModifier by allowing to recompute inverse Hessian `num_recomputations` times when doing a pruning step. 

For example, in the standard pipeline we would ask OBSPruningModifier to prune a model to 50% and it would do so by estimating inverse Hessian once and then use it to prune weights. This recomputation feature enables doing the same pruning step but with multiple estimations of the inverse Hessian. For example, `num_recomputations=2` would estimate the inverse Hessian, prune `50%/num_recomputations = 25%` of weights, and then re-estimate it again (with newly pruned weights taken into consideration) and prune the remaining 25% to reach the pruning target of 50%.

This feature helps a lot in situations where a lot of weights are being removed in one pruning step (e.g. 50%, 70%). OBS method relies on the fact that the model is in the vicinity of the local minimum, and when removing many weights at once this assumption breaks.

To demonstrate benefits of recomputations, here are some one-shot pruning results of the fine-tuned BERT-base model on the SQuADv1.1 task:

| num\_recomputations | 70% sparsity | 80% sparsity | 90% sparsity |
| ------------------- | ------ | ------ | ------ |
| 1, default          | 83.38  | 57.48  | 14.57  |
| 4                   | 85.17  | 72.76  | 29.23  |
| 8                   | 85.34  | 77.85  | 39.58  |
| 16                  | 85.49  | 78.27  | 45.03  |
| 32                  | 85.49  | 78.65  | 47.67  |

`1, default` is the standard approach without any additional re-estimations, i.e. what we have been using so far.

### DIFF
--- a/src/sparseml/pytorch/sparsification/pruning/modifier_pruning_base.py
+++ b/src/sparseml/pytorch/sparsification/pruning/modifier_pruning_base.py
@@ -347,7 +347,12 @@ class BasePruningModifier(ABC, ScheduledUpdateModifier):
         self.check_mask_update(module, epoch, steps_per_epoch)
 
     def check_mask_update(
-        self, module: Module, epoch: float, steps_per_epoch: int, **kwargs
+        self,
+        module: Module,
+        epoch: float,
+        steps_per_epoch: int,
+        recomputation_sparsity: Optional[float] = None,
+        **kwargs,
     ):
         """
         Update mask values if necessary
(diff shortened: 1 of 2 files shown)


bfineran (approved):
looks great, thanks @eldarkurtic

PR #1052, opened 2022-09-28
Efficientnet v2

anmarques:
This PR adds support for EfficientNetV2 models. Summary of changes:
1. Creates the _FusedInvertedBottleneckBlock (used in v2 models)
2. Allows batchnorm configs to be passed to the EfficientNet class (need to set eps value)
3. Create definitions for efficientnet_v2_s, efficientnet_v2_m and efficientnet_v2_l
3. Adds support to set rgb_means and rgb_stds to dataset (needed for efficientnet_v2_l)

Torchvision checkpoints were ported to match these model definitions and accuracy was verified.

### DIFF
--- a/src/sparseml/pytorch/models/classification/efficientnet.py
+++ b/src/sparseml/pytorch/models/classification/efficientnet.py
@@ -19,8 +19,7 @@ Further info can be found in the paper `here <https://arxiv.org/abs/1905.11946>`
 
 import math
 from collections import OrderedDict
-from dataclasses import dataclass
-from typing import List, Optional, Tuple, Union
+from typing import List, Optional, Tuple, Union, Mapping
 
 import torch
 from torch import Tensor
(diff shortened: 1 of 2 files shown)


src/sparseml/pytorch/models/classification/efficientnet.py
```diff
@@ -862,4 +1003,334 @@ def efficientnet_b7(
...
+    ],
+    input_shape=(3, 480, 480),
+    domain="cv",
+    sub_domain="classification",
+    architecture="efficientnet_v2",
+    sub_architecture="m",
+    default_dataset="imagenet",
+    default_desc="base",
+    def_ignore_error_tensors=["classifier.fc.weight", "classifier.fc.bias"],
+    desc_args={"optim-perf": ("se_mod", True)},
```
bfineran, line 1128: maybe not for this PR, but we should consider adding some default dataloader kwargs here, so the default loaded weights are compatible with the loader

bfineran (dismissed):

src/sparseml/pytorch/models/classification/efficientnet.py
```diff
@@ -220,6 +231,93 @@ def forward(self, inp: Tensor):
...
 
+class _FusedInvertedBottleneckBlock(Module):
+    def __init__(
+        self,
+        in_channels: int,
+        out_channels: int,
+        kernel_size: int,
+        expansion_ratio: int,
+        stride: int,
+        bn_kwargs: Optional[Mapping] = None,
```
rahul-tuli, line 242:
```suggestion
        bn_kwargs: Optional[Mapping[str, Any]] = None,
```

rahul-tuli (dismissed):
LGTM! I really like the use of `typing.Mapping` over `typing.Dict`; I have one request, could we update the typing of `bn_kwargs` to `Optional[Mapping[str, Any]]` to make it more explicit? Thanks!

src/sparseml/pytorch/models/classification/efficientnet.py
anmarques: That's a great call @bfineran

src/sparseml/pytorch/models/classification/efficientnet.py
```diff
@@ -98,21 +101,26 @@ def __init__(
         kernel_size: int,
         expansion_ratio: int,
         stride: int,
-        se_ratio: Union[float, None],
+        se_ratio: Optional[float],
```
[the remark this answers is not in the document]
anmarques: 😄

rahul-tuli (approved):

bfineran (approved):

PR #1053, opened 2022-09-29
Add a registry for fetching Modifier Info

rahul-tuli:
This PR adds a `ModifierBuildInfo` class with state + utility methods for building a recipe from `modifier(s)`
This also adds a registry for various pruning and quantization modifiers, that will be used to fetch modifier info

### DIFF
--- a/src/sparseml/recipe_template/utils.py
+++ b/src/sparseml/recipe_template/utils.py
@@ -1,0 +1,144 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Any, Dict, Optional
+
+from sparseml.pytorch.sparsification import (
+    ACDCPruningModifier,
+    GMPruningModifier,
+    MagnitudePruningModifier,
+    Modifier,
+    QuantizationModifier,
+)
+from sparseml.sparsification import ModifierYAMLBuilder
+
+
+__all__ = [
+    "ModifierBuildInfo",
+]
+
+
+class ModifierBuildInfo:
+    """
+    A class with state and helper methods for building a recipe from
+    modifier(s)
+
+    :param modifier: class of the Modifier in question
+    :param modifier_name: str A string name to be prefixed with all attribute
+        variables defined for this modifier, defaults to "mod"
+    :param fields: An Optional dictionary mapping modifier attribute names to their
+        corresponding values
+    """
+
+    def __init__(
+        self,
+        modifier: Modifier,
+        modifier_name: str = "mod",
+        fields: Optional[Dict[str, Any]] = None,
+    ):
+        self.modifier = modifier
+        self.modifier_name = modifier_name
+        self.modifier_builder = ModifierYAMLBuilder(modifier)
+        self.fields = fields or {}
+        self.__modifier_recipe_variables = {}
+        self.update(updated_fields=self.fields)
+
+    def update(self, updated_fields: Optional[Dict[str, Any]] = None):
+        """
+        Utility method to update the fields for current modifier
+
+        :param updated_fields: An Optional dictionary mapping modifier attribute names
+            to their corresponding values
+        """
+        for key, value in updated_fields.items():
+            variable_name = f"{self.modifier_name}_{key}"
+            setattr(self.modifier_builder, key, f"eval({variable_name})")
+            self.__modifier_recipe_variables[variable_name] = updated_fields[key]
+
+    @property
+    def modifier_recipe_variables(self) -> Dict[str, Any]:
+        """
+        :returns: A dict mapping b/w recipe variable names --> values
+            for the current object
+        """
+        return self.__modifier_recipe_variables
+
+
+# PRUNING MODIFIERS INFO
+
+_PRUNING_MODIFIER_INFO_REGISTRY = {
+    "false": None,
+    "true": ModifierBuildInfo(
+        modifier=MagnitudePruningModifier,
+        modifier_name="pruning",
+        fields={
+            "init_sparsity": 0.05,
+            "final_sparsity": 0.8,
+            "start_epoch": 0.0,
+            "end_epoch": 10.0,
+            "update_frequency": 1.0,
+            "params": "__ALL_PRUNABLE__",
+            "leave_enabled": True,
+            "inter_func": "cubic",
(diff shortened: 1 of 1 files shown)


src/sparseml/recipe_template/utils.py
```diff
@@ -0,0 +1,144 @@
...
+    "ModifierBuildInfo",
+]
+
+
+class ModifierBuildInfo:
+    """
+    A class with state and helper methods for building a recipe from
+    modifier(s)
+
+    :param modifier: class of the Modifier in question
```
dbogunowicz, line 37: Would be great to have some documentation (pseudocode in this docstring, alternatively unit test that would have a side-effect of being a form of documentation) of how the potential use of this class in code would be.

src/sparseml/recipe_template/utils.py
```diff
@@ -0,0 +1,144 @@
...
+        self,
+        modifier: Modifier,
+        modifier_name: str = "mod",
+        fields: Optional[Dict[str, Any]] = None,
+    ):
+        self.modifier = modifier
+        self.modifier_name = modifier_name
+        self.modifier_builder = ModifierYAMLBuilder(modifier)
+        self.fields = fields or {}
+        self.__modifier_recipe_variables = {}
```
dbogunowicz, line 54: what is the reason to use `__` here?

src/sparseml/recipe_template/utils.py
```diff
@@ -0,0 +1,144 @@
...
+)
+from sparseml.sparsification import ModifierYAMLBuilder
+
+
+__all__ = [
+    "ModifierBuildInfo",
+]
+
+
+class ModifierBuildInfo:
```
dbogunowicz, line 32: Shouldn't this class also have a method of actually building a recipe? So far it is only updating its state.

src/sparseml/recipe_template/utils.py
```diff
@@ -0,0 +1,144 @@
...
+        """
+        :returns: A dict mapping b/w recipe variable names --> values
+            for the current object
+        """
+        return self.__modifier_recipe_variables
+
+
+# PRUNING MODIFIERS INFO
+
+_PRUNING_MODIFIER_INFO_REGISTRY = {
```
dbogunowicz, line 80: All those registers do not have any relationship to `ModifierBuildInfo`, nor are they designed to be exported from this script. Maybe we should move them to a separate script?

src/sparseml/recipe_template/utils.py
```diff
@@ -0,0 +1,144 @@
...
+    def update(self, updated_fields: Optional[Dict[str, Any]] = None):
+        """
+        Utility method to update the fields for current modifier
+
+        :param updated_fields: An Optional dictionary mapping modifier attribute names
+            to their corresponding values
+        """
+        for key, value in updated_fields.items():
+            variable_name = f"{self.modifier_name}_{key}"
+            setattr(self.modifier_builder, key, f"eval({variable_name})")
```
dbogunowicz, line 66: are we not worried about using the (generally unsafe) `eval()` function here? Not sure what's the relationship of this class to the general UI, but if the user can manually specify the `fields` argument, using `eval()` may not be the best choice.

src/sparseml/recipe_template/utils.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+# PRUNING MODIFIERS INFO
+
+_PRUNING_MODIFIER_INFO_REGISTRY = {
+    "false": None,
+    "true": ModifierBuildInfo(
+        modifier=MagnitudePruningModifier,
+        modifier_name="pruning",
+        fields={
```
dbogunowicz, line 85: the fields here are default files for the modifier?

dbogunowicz (commented):
Looks good, left couple of suggestions.

src/sparseml/recipe_template/utils.py
dbogunowicz: Ok, not a separate script, but I would probably bring them to the top of the script and make them public (why are they private? there are not used in the code by `ModifierBuildInfo`) and add them to `__all__`

corey-nm: +1 they seem to be used in #1054 , so i think they should be moved there (and honestly could probably just be functions instead of global dicts)

src/sparseml/recipe_template/utils.py
corey-nm: I think this is putting the string eval into the recipe (which we do all over the place in recipes).

corey-nm, line 66: Why use setattr here? Should these go in self.fields?

src/sparseml/recipe_template/utils.py
```diff
@@ -0,0 +1,144 @@
...
+            "inter_func": "cubic",
+            "mask_type": "unstructured",
+        },
+    ),
+}
+
+# QUANTIZATION MODIFIERS INFO
+
+_QUANTIZATION_MODIFIER_INFO_REGISTRY = {
+    "false": None,
```
corey-nm, line 129: Recommend remove "false" and do `.get(name, None)` on retrieval.

src/sparseml/recipe_template/utils.py
```diff
@@ -0,0 +1,144 @@
...
+    "true": ModifierBuildInfo(
```
corey-nm, line 130: Can we come up with a better name for this that's more specified to what it does? Could totally see adding other options for this in the future

src/sparseml/recipe_template/utils.py
```diff
@@ -0,0 +1,144 @@
...
+            for the current object
+        """
+        return self.__modifier_recipe_variables
+
+
+# PRUNING MODIFIERS INFO
+
+_PRUNING_MODIFIER_INFO_REGISTRY = {
+    "false": None,
+    "true": ModifierBuildInfo(
```
corey-nm, line 82: maybe rename to `magnitude`?

src/sparseml/recipe_template/utils.py
corey-nm: Ahh so we could just insert the actual value here instead of evaling?

src/sparseml/recipe_template/utils.py
KSGulin: I would prefer to keep them in the current dictionary form or as formattable strings. It's easier to read and the functionality/intention is more straightforward

src/sparseml/recipe_template/utils.py
dbogunowicz: +1

src/sparseml/recipe_template/utils.py
dbogunowicz: no, what I am saying is, we would probably not like to have the user put something weird in the recipe_template config, that can later end up as an argument to `eval()` and get executed.

src/sparseml/recipe_template/utils.py
rahul-tuli: [name mangling](https://medium.com/analytics-vidhya/python-name-mangling-and-how-to-use-underscores-e67b529f744f), if a user tries to change `__modifier_recipe_variables` it will be more difficult.

dbogunowicz: yeah i know, but why this specific variables gets it?

rahul-tuli:
closing this one out due to a design change that does not need this code

PR #1057, opened 2022-09-29
Adding pass to qat that folds Identity nodes

corey-nm:
Currently in torch 1.12, the onnx export creates identity nodes with a single initializer (See image below). This PR adds a pass that folds the initializers from these Identity nodes into the actual nodes. I.e. it removes the identity nodes.

Test plan: unit tests

Before:
image


src/sparseml/pytorch/sparsification/quantization/quantize_qat_export.py
```diff
@@ -1556,6 +1585,7 @@ def quantize_torch_qat_export(
 
     _fold_qat_conv_bns(model)
     _convert_single_constants_to_initializers(model)
+    _fold_identity_initializers(model)
```
corey-nm, line 1588: Does it matter where i put this?

src/sparseml/pytorch/utils/exporter.py
```diff
@@ -633,6 +634,35 @@ def _save_label_to_class_mapping(
...
+        if not is_match(match):
+            continue
+        matches.append(match)
+
+        # find any node in the graph that uses the output of `match`
+        # as an input. replace the input with `match`'s input
+        for other in model.graph.node:
+            for i in range(len(other.input)):
+                if other.input[i] == match.output[0]:
+                    other.input[i] = match.input[0]
```
bfineran, line 660: can clean up with `ONNXGraph`
https://github.com/neuralmagic/sparseml/blob/d86283f4402842bb3c53b1c1f6461c3a8991799e/src/sparseml/onnx/utils/graph_editor.py#L140

src/sparseml/pytorch/utils/exporter.py
```diff
@@ -633,6 +634,35 @@ def _save_label_to_class_mapping(
     )
 
 
+def _fold_identity_initializers(model: onnx.ModelProto):
+    # folds any Identity nodes that have a single input (which is an initializer)
```
dbogunowicz, line 638: Why identity nodes get created in the first place?

src/sparseml/pytorch/utils/exporter.py
```diff
@@ -633,6 +634,35 @@ def _save_label_to_class_mapping(
...
+
+    def is_match(node: onnx.NodeProto) -> bool:
+        return (
+            node.op_type == "Identity"
+            and len(node.input) == 1
+            and len(node.output) == 1
+            and any(node.input[0] == init.name for init in model.graph.initializer)
+        )
+
+    for match in model.graph.node:
```
dbogunowicz, line 650:
```suggestion
    for node in model.graph.node:
```

src/sparseml/pytorch/utils/exporter.py
```diff
@@ -633,6 +634,35 @@ def _save_label_to_class_mapping(
...
+        if not is_match(match):
+            continue
+        matches.append(match)
```
dbogunowicz, line 653: more simple --> if is match --> append

dbogunowicz (commented):
few comments

dbogunowicz (dismissed):

src/sparseml/pytorch/utils/exporter.py
```diff
@@ -633,6 +634,36 @@ def _save_label_to_class_mapping(
...
+    # and a single output
+    matches = []
+
+    def is_match(node: onnx.NodeProto) -> bool:
+        return (
+            node.op_type == "Identity"
+            and len(node.input) == 1
+            and len(node.output) == 1
+            and any(node.input[0] == init.name for init in model.graph.initializer)
+        )
```
rahul-tuli, line 648: Appreciate you abstracting the actual check into a function, makes it super easy to understand the code

src/sparseml/pytorch/utils/exporter.py
```diff
@@ -633,6 +634,36 @@ def _save_label_to_class_mapping(
...
+
+    for node in model.graph.node:
+        if not is_match(node):
+            continue
+        matches.append(node)
+
+        # find any node in the graph that uses the output of `node`
+        # as an input. replace the input with `node`'s input
+        for other in model.graph.node:
+            for i in range(len(other.input)):
```
rahul-tuli, line 658: nit pick:
```suggestion
            for index in range(len(other.input)):
```

rahul-tuli: also would `enumerate` be cleaner here?

rahul-tuli (approved):

src/sparseml/pytorch/utils/exporter.py
corey-nm: Ya i'll update that rq!

src/sparseml/pytorch/utils/exporter.py
```diff
@@ -655,9 +655,9 @@ def is_match(node: onnx.NodeProto) -> bool:
         # find any node in the graph that uses the output of `node`
         # as an input. replace the input with `node`'s input
         for other in model.graph.node:
-            for i in range(len(other.input)):
+            for i, other_input_i in enumerate(other.input):
```
rahul-tuli, line 658: Could we rename `i` ?

rahul-tuli (dismissed):

src/sparseml/pytorch/utils/exporter.py
corey-nm: why?

src/sparseml/pytorch/utils/exporter.py
bfineran: it doesn't get rid of the swap - replaces the need for the double for loop (ie finding nodes that take `match.output[0]` as input

bfineran:
LGTM pending comment

src/sparseml/pytorch/utils/exporter.py
rahul-tuli: I'm good either way, our codebase tries to use something like `index` over `i` where ever we can

for index: 31 matches in 20 files
for i: 29 matches in 16 files

So it's upto you!

src/sparseml/pytorch/utils/exporter.py
corey-nm: I think we still need the double loop to only replace inputs that are equal, but will do less work. Updated

src/sparseml/pytorch/utils/exporter.py
corey-nm: I choose `i` 😁

rahul-tuli: Accepted!

rahul-tuli (approved):

bfineran (approved):

PR #1058, opened 2022-09-30
Feature branch for ModuleExporter Refactor

corey-nm:
# TODOS
- [ ] Add unit tests for existing passes
- [ ] Add utility `match_sequences` which will return all sequences of specified NodeMatches. For example the following would iterate all sequences of Conv->Div->BatchNorm nodes:
```python
match_sequences(model, [Conv, Div, BatchNorm])
```
- [ ] Add helper properties for `NodeMatch` with params in attributes (Conv/BatchNorm/etc)
- [ ] Add new passes:
    - [ ] fold transpose with initializer
- [ ] Combine Add/Cast/Mul passes (QDQ -> Gemm + bias, MatMul -> Add, Conv + bias -> QDQ)
- [ ] Combine Gemm/Conv passes (QDQ -> Gemm, Conv -> QDQ)
- [ ] use `List[ModelSimplifier]` in `export_onnx`
    - [ ] if convert_qat=True, then extend the list of simplifiers with QAT passes

# Summary

The main goals of this refactor are:
1. Harden onnx export pipeline (tests, etc)
2. Simplify matching logic of passes
3. Simplify logic of passes
4. Combine as many passes as possible

There are two new main base classes to support this:
1. `NodeMatch`, which will match against specific nodes in the graph, and also make it easy to access expected attributes from onnx nodes
2. `ModelSimplifier` which represents a single optimization pass

### DIFF
--- a/src/sparseml/onnx/simplify/__init__.py
+++ b/src/sparseml/onnx/simplify/__init__.py
@@ -1,0 +1,13 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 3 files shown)


corey-nm:
Closing - will re-open later when actively worked on

PR #1061, opened 2022-09-30
[QuantizationModifier] pydantic classes for defining quantization schemes to generate QConfigs

bfineran:
a core feature of the QuantizationModifier refactor is the ability for users to have both more simple and more fine grained control over how quantization is applied at large and in small pieces of the model.  To support this, this PR introduces two simple pydantic models that can be used to define how a particular layer's inputs, weights, and outputs are quantized

intended use:
`QuantizationArgs` defines how quantization should be applied for a particular component via `num_bits`, `symmetric` and optional kwargs that map to pytorch quantization `Observer` args.
`QuantizationScheme` builds on `QuantizationArgs` to allow for args to be set separately for inputs, weights, and outputs. If any of these values are set to None, that particular component of the layer will not be quantized.

helper methods are included to create `Observer` and `QConfig` objects from the pydantic models respectively.

these classes will be built on for parsing inputs to recipes - recipe users will not interface with these objects directly:

a post-refactor modifier YAML may look something like
```yaml
!QuantizationModifier
  ...
  submodule_schemes:
    "bert.encoder":
        input_activations:
          num_bits: 8
          symmetric: False
        weights:
          num_bits: 8
          symmetric: True
        output_activations: None
```

**test_plan**
unit tests included

### DIFF
--- a/src/sparseml/pytorch/sparsification/quantization/quantize.py
+++ b/src/sparseml/pytorch/sparsification/quantization/quantize.py
@@ -1,0 +1,140 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Tooling for applying quantization to pytorch modules via
+structured configurations
+"""
+
+from typing import Any, Dict, Optional
+
+import torch
+from pydantic import BaseModel, Field
+
+from sparseml.pytorch.sparsification.quantization.helpers import compute_range
+
+
+try:
+    from torch import quantization as torch_quantization
+except Exception:
+    torch_quantization = None
+
+
+__all__ = [
+    "QuantizationArgs",
+    "QuantizationScheme",
+]
+
+
+class QuantizationArgs(BaseModel):
+    """
+    Class representing user facing arguments to define quantization Observers of
+    activations or weights in a network
+    """
+
+    num_bits: int = Field(default=8, help="number of bits to target for quantization")
+    symmetric: bool = Field(
+        default=False, help="set True to use symmetric quantization. Default False"
+    )
+    kwargs: Dict[str, Any] = Field(
+        default_factory=dict,
+        help=(
+            "optional dict of kwargs to be passed directly to torch quantization "
+            "Observers constructor excluding quantization range or symmetry"
+        ),
+    )
+
+    @classmethod
+    def default_activation_args(cls):
+        """
+        :return: default 8 bits asymmetric settings
+        """
+        return cls(num_bits=8, symmetric=False)
+
+    @classmethod
+    def default_weight_args(cls):
+        """
+        :return: default 8 bits symmetric settings
+        """
+        return cls(num_bits=8, symmetric=True)
+
+    def get_observer(self) -> "torch.quantization.FakeQuantize":
+        """
+        :return: torch quantization FakeQuantize built based on these QuantizationArgs
+        """
+        qscheme = (
+            torch.per_tensor_symmetric if self.symmetric else torch.per_tensor_affine
+        )
+        target_dtype = torch.qint8
+        quant_min, quant_max = compute_range(target_dtype, self.num_bits)
+        return torch_quantization.FakeQuantize.with_args(
+            observer=torch_quantization.MovingAverageMinMaxObserver,
+            quant_min=quant_min,
+            quant_max=quant_max,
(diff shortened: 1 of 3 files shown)


tests/sparseml/pytorch/sparsification/quantization/test_quantize.py
```diff
@@ -0,0 +1,100 @@
...
+            QuantizationArgs(num_bits=4, symmetric=True),
+        ),
+        (
+            dict(num_bits=8, kwargs=dict(reduce_rage=True)),
+            QuantizationArgs(num_bits=8, kwargs=dict(reduce_rage=True)),
+        ),
+    ],
+)
+def test_quantization_args_from_dict(quantization_args_dict, target_quantization_args):
+    loaded_quantization_args = QuantizationArgs.parse_obj(quantization_args_dict)
```
rahul-tuli, line 41: Could we rather create the object using kwargs, like `QuantizationArgs(**quantization_args_dict)` I don't think users will use pydantic's `parse_obj` to create `QuantizationArgs` objects

rahul-tuli: Feel free to ignore this though

tests/sparseml/pytorch/sparsification/quantization/test_quantize.py
```diff
@@ -0,0 +1,100 @@
...
+    [
+        (QuantizationArgs(num_bits=8, symmetric=False), -128, 127),
+        (QuantizationArgs(num_bits=4, symmetric=True), -8, 7),
+    ],
+)
+def test_quantization_args_get_observer(
+    qunatization_args, target_quant_min, target_quant_max
+):
+    observer = qunatization_args.get_observer()
+    assert hasattr(observer, "with_args")
```
rahul-tuli, line 98: Is this really necessary? What does this line check?

src/sparseml/pytorch/sparsification/quantization/quantize.py
```diff
@@ -0,0 +1,140 @@
...
+]
+
+
+class QuantizationArgs(BaseModel):
+    """
+    Class representing user facing arguments to define quantization Observers of
+    activations or weights in a network
+    """
+
+    num_bits: int = Field(default=8, help="number of bits to target for quantization")
```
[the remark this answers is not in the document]
bfineran: definitely, oversight

tests/sparseml/pytorch/sparsification/quantization/test_quantize.py
bfineran: type check isn't great here because a partial is technically returned. this checks that the partial has expected format

tests/sparseml/pytorch/sparsification/quantization/test_quantize.py
bfineran: end users will never create `QuantizationArgs` on their own - these classes are used to parse modifier arguments.

`parse_obj` is used because it will recursively parse any nested pydantic classes

src/sparseml/pytorch/sparsification/quantization/quantize.py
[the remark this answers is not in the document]
KSGulin: It looks like this hasn't been addressed yet

src/sparseml/pytorch/sparsification/quantization/quantize.py
```diff
@@ -0,0 +1,140 @@
...
+class QuantizationScheme(BaseModel):
+    """
+    Class composed of QuantizationArgs to build QConfig and QuantWrapper objects for
+    quantizing models. Provides a simple user interface for defining how inputs,
+    weights, and outputs should be quantized
+    """
+
+    input_activations: Optional[QuantizationArgs] = Field(
+        default_factory=QuantizationArgs.default_activation_args,
+        help=(
```
dbogunowicz, line 100: let's address @rahul-tuli comment (swap help for a description) across multiple Classes.

src/sparseml/pytorch/sparsification/quantization/quantize.py
```diff
@@ -0,0 +1,140 @@
...
+    def get_qconfig(self) -> "torch.quantization.QConfig":
+        """
+        :return: QConfig for Modules (output activations used,
+            use QuantWrapper for inputs)
+        """
+        return _get_qconfig(self.output_activations, self.weights)
+
+    def get_wrapper_qconfig(self) -> "torch.quantization.QConfig":
+        """
+        :return: QConfig for QuantWrapper objets (input activations used)
```
dbogunowicz, line 129:
```suggestion
        :return: QConfig for QuantWrapper objects (input activations used)
```

src/sparseml/pytorch/sparsification/quantization/quantize.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+        return _get_qconfig(self.input_activations, self.weights)
+
+
+def _get_qconfig(
+    activation_args: Optional[QuantizationArgs], weight_args: Optional[QuantizationArgs]
```
dbogunowicz, line 135: If the argument is optional, its default value should be specified as `None`.

```suggestion
    activation_args: Optional[QuantizationArgs] = None, weight_args: Optional[QuantizationArgs] = None
```

rahul-tuli (approved):
LGTM!

PR #1066, opened 2022-10-04
No longer removing initializer if it is still in use in onnx graph

corey-nm:
The current method of keep_params causes a crash in torch 1.12, because initializers can be used by multiple nodes. Therefore removing an initializer is only safe if it is not used.


bfineran (approved):
LGTM - definitely exposes some of the issues we are addressing in the refactor

rahul-tuli (approved):

PR #1068, opened 2022-10-04
Simplifying and fixing fix_batch_norm_names

corey-nm:
In torch 1.12, some of the initializers are used by multiple nodes. For this fix_batch_norm_names function, this was causing a bug where:
1. A node's input & initializer were updated by removing ".module"
2. Another node that also used that initializer was not updated because the check for `init_name not in name_to_inits` failed, so the node was skipped

This PR simplifies the logic of the above by changing initializer names and node input/output names separately.

### DIFF
--- a/src/sparseml/pytorch/utils/exporter.py
+++ b/src/sparseml/pytorch/utils/exporter.py
@@ -747,8 +747,6 @@ def _delete_trivial_onnx_adds(model: onnx.ModelProto):
 
 
 def _unwrap_batchnorms(model: onnx.ModelProto):
-    onnx.checker.check_model(model)
-
     for init in model.graph.initializer:
         init.name = init.name.replace(".bn_wrapper_replace_me", "")
     for node in model.graph.node:
--- a/src/sparseml/pytorch/utils/exporter.py
+++ b/src/sparseml/pytorch/utils/exporter.py
@@ -530,9 +530,10 @@ def export_onnx(
     # onnx file fixes
     onnx_model = onnx.load(file_path)
     _fold_identity_initializers(onnx_model)
-    # fix changed batch norm names
-    _remove_module_from_names(onnx_model)
     if batch_norms_wrapped:
+        # fix changed batch norm names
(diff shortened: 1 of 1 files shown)


KSGulin (approved):

src/sparseml/pytorch/utils/exporter.py
```diff
@@ -745,25 +745,15 @@ def _delete_trivial_onnx_adds(model: onnx.ModelProto):
             continue
 
 
-def _fix_batch_norm_names(model: onnx.ModelProto):
-    name_to_inits = {init.name: init for init in model.graph.initializer}
+def _remove_module_from_names(model: onnx.ModelProto):
+    onnx.checker.check_model(model)
```
bfineran, line 749: why do we need to check before?

src/sparseml/pytorch/utils/exporter.py
```diff
@@ -745,25 +745,15 @@ def _delete_trivial_onnx_adds(model: onnx.ModelProto):
...
+
+    for init in model.graph.initializer:
+        init.name = init.name.replace(".module", "")
     for node in model.graph.node:
-        if node.op_type != "BatchNormalization":
```
bfineran, line 751: would prefer to keep scope still just to BN ops, `module` is still a valid name

src/sparseml/pytorch/utils/exporter.py
corey-nm: for more checks 😁 i can remove it

src/sparseml/pytorch/utils/exporter.py
corey-nm: and i guess only initializers that go into batch norms?

corey-nm: Okay I changed the name `module` to something else that should be way more specific to this wrapper. Prefer to keep the simpler logic of this function because the previous version was harder to understand than it needs to be

bfineran (approved):

KSGulin (approved):

PR #1070, opened 2022-10-05
Adding Softmax to to_exclude in quantization modifier

corey-nm:
In torch 1.12, softmax gets a qconfig added to it by the quantization modifier. Thus it ends up with QDQ nodes after the softmax node in the onnx graph:
image

This PR adds Softmax to a to_exclude list in Quantization modifier so this does not occur.

Test plan: ran resnet50 export for quant & non-quant versions and verified that softmax no longer has QDQ nodes after it in 1.12


bfineran (commented):
we should have some way to override this default behavior
(will be resolved after quantization modifier refactor regardless)

bfineran (dismissed):

PR #1080, opened 2022-10-13
Deprecating dynamic_axes and always setting it

corey-nm:
Reasoning: in torch 1.12 any shape operations *without* batch dimension as dynamic will be treated as a constant, and therefore constant folding operations will apply to them. For example in resnet there is a flatten operation towards the end of the graph.

In torch 1.12, without dynamic_axes="batch" set, you get this graph:

image

With dynamic_axes="batch set, you get this graph:

image

### DIFF
--- a/src/sparseml/pytorch/utils/exporter.py
+++ b/src/sparseml/pytorch/utils/exporter.py
@@ -484,13 +484,18 @@ def export_onnx(
     if "output_names" not in export_kwargs:
         export_kwargs["output_names"] = _get_output_names(out)
 
-    if dynamic_axes == "batch":
-        dynamic_axes = {
-            tensor_name: {0: "batch"}
-            for tensor_name in (
-                export_kwargs["input_names"] + export_kwargs["output_names"]
-            )
-        }
+    if dynamic_axes is not None:
+        warnings.warn(
+            "Setting dynamic_axes is deprecated, now it is always set",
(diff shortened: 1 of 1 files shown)


KSGulin (approved):
LGTM!

src/sparseml/pytorch/utils/exporter.py
```diff
@@ -484,13 +484,18 @@ def export_onnx(
...
-    if dynamic_axes == "batch":
-        dynamic_axes = {
-            tensor_name: {0: "batch"}
-            for tensor_name in (
-                export_kwargs["input_names"] + export_kwargs["output_names"]
-            )
-        }
+    if dynamic_axes is not None:
+        warnings.warn(
+            "Setting dynamic_axes is deprecated, now it is always set",
```
bfineran, line 489: let's make this warning a bit more specific to show that the value the user set `dynamic_axes` as is no longer respected and we are setting the 0th index across all inputs as dynamic

bfineran (changes requested):

src/sparseml/pytorch/utils/exporter.py
corey-nm: Updated

rahul-tuli (approved):

bfineran (approved):

PR #1062, opened 2022-10-03
python 3.10/torch 1.12 support

corey-nm:
- update github workflows to use python 3.10
- add backwards compatible github workflow to run with 3.9 and torch 1.9
- updates onnx opset defaults to be based on torch version
- raising error at runtime if torch 1.10/1.11 are installed. users can bypass with an env variable
- raising max torch version
- adding different calls for torch.ao.quantization/torch.onnx.export based on torch versions

# Tests

### Training runs
the following training runs have been executed on this branch without errors

1. distilbert on sst2
```bash
sparseml.transformers.train.text_classification \
  --output_dir sparse_distilbert-sentiment_analysis_sst2 \
  --model_name_or_path zoo:nlp/masked_language_modeling/distilbert-none/pytorch/huggingface/wikipedia_bookcorpus/pruned80_quant-none-vnni \
  --recipe zoo:nlp/sentiment_analysis/distilbert-none/pytorch/huggingface/sst2/pruned80_quant-none-vnni \
  --distill_teacher zoo:nlp/sentiment_analysis/bert-base/pytorch/huggingface/sst2/base-none \
  --task_name sst2 \
  --do_train \
  --do_eval \
  --fp16 \
  --evaluation_strategy epoch \
  --logging_steps 2000 \
  --logging_first_step \
  --per_device_train_batch_size 32 \
  --per_device_eval_batch_size 32 \
  --preprocessing_num_workers 6 \
  --gradient_accumulation_steps 1 \
  --max_seq_length 128 \
  --seed 21868 \
  --save_strategy epoch \
  --save_total_limit 1
```
2. yolov5 on voc
```bash
sparseml.yolov5.train \
    --data VOC.yaml \
    --cfg models_v5.0/yolov5s.yaml \
    --weights zoo:cv/detection/yolov5-s/pytorch/ultralytics/coco/pruned_quant-aggressive_94?recipe_type=transfer \
    --hyp data/hyps/hyp.finetune.yaml \
    --recipe zoo:cv/detection/yolov5-s/pytorch/ultralytics/coco/pruned_quant-aggressive_94?recipe_type=transfer \
    --max-train-steps 20
```
3. resnet50 on imagenette
```bash
sparseml.image_classification.train \
    --recipe-path zoo:cv/classification/resnet_v1-50/pytorch/sparseml/imagenet/pruned_quant-moderate?recipe_type=transfer_learn	\
[...]

PR #1082, opened 2022-10-14
Reworking analyzer model and extract_node_shape tests

corey-nm:
This addresses test failures from #1080, and also addresses test failures in #1062 

Test Plan: Any suggestions for how to validate the data inside the model analysis & node shape files?


bfineran (commented):
are there any old test files we should delete now?

corey-nm:
> are there any old test files we should delete now?

AFAIK no, the stuff used for the old tests was all specified in python, which I deleted from the helpers file

bfineran (approved):

rahul-tuli (approved):

PR #1084, opened 2022-10-17
One shot export [Image classification]

rahul-tuli:
This PR adds support to apply a recipe in `one-shot` manner while exporting Image classification models

Test plan: Automated test to start with a vanilla(non-quantized) `resnet50`, export checkpoint in one-shot manner with a quantization recipe and then check the onnx graph for quantized nodes

Note the failing test should pass after #1122 lands

### DIFF
--- a/src/sparseml/pytorch/image_classification/export.py
+++ b/src/sparseml/pytorch/image_classification/export.py
@@ -97,6 +97,8 @@ Options:
   --num-classes, --num_classes INTEGER
                                   number of classes for model; must be set if
                                   a dataset is not provided
+  --one-shot, --one_shot TEXT     recipe to be applied in one-shot manner,
+                                  before exporting
   --help                          Show this message and exit.
 
 
@@ -106,6 +108,13 @@ sparseml.image_classification.export_onnx \
     --arch-key resnet50 --dataset imagenet \
     --dataset-path ~/datasets/ILSVRC2012 \
     --checkpoint-path ~/checkpoints/resnet50_checkpoint.pth
+
+Example command for exporting ResNet50 with one-shot:
+sparseml.image_classification.export_onnx \
+    --arch-key resnet50 --dataset imagenet \
+    --dataset-path ~/datasets/ILSVRC2012 \
+    --checkpoint-path ~/checkpoints/resnet50_checkpoint.pth \
+    --one-shot ~/recipes/resnet_recipe.md
 """
 import json
 import os
@@ -291,6 +300,14 @@ LOGGER = get_main_logger()
     show_default=True,
     help="number of classes for model; must be set if a dataset is not provided",
 )
+@click.option(
+    "--one-shot",
+    "--one_shot",
+    type=str,
+    default=None,
+    show_default=True,
+    help="recipe to be applied in one-shot manner, before exporting",
+)
 def main(
     dataset: Optional[str] = None,
     dataset_path: Optional[str] = None,
@@ -310,6 +327,7 @@ def main(
     recipe: Optional[str] = None,
     convert_qat: bool = True,
     num_classes: Optional[int] = None,
+    one_shot: Optional[str] = None,
 ):
     """
     SparseML-PyTorch Integration for exporting image classification models to
@@ -369,7 +387,13 @@ def main(
 
     if recipe is not None:
         ScheduledModifierManager.from_yaml(recipe).apply_structure(model)
+
+    if checkpoint_path:
         load_model(checkpoint_path, model, strict=True)
+
+    if one_shot is not None:
+        ScheduledModifierManager.from_yaml(file_path=one_shot).apply(module=model)
+
     export(
         model=model,
         val_loader=val_loader,
--- a/tests/sparseml/pytorch/image_classification/test_export.py
+++ b/tests/sparseml/pytorch/image_classification/test_export.py
@@ -1,0 +1,97 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import math
+from pathlib import Path
+
+import onnx
+import pytest
+import torch
+
+from click.testing import CliRunner
+from sparseml.pytorch.image_classification.export import main
+from sparseml.pytorch.models import resnet18
+from sparsezoo.analysis import ModelAnalysis
+
+
+@pytest.fixture()
+def temp_dir(tmp_path):
+    """
+    return: a temporary directory
+    """
+    test_dir = tmp_path / "test_dir"
+    test_dir.mkdir()
+    yield test_dir
+
+
+@pytest.fixture()
+def resnet_checkpoint(temp_dir):
+    """
+    return: path to a ResNet checkpoint
+    """
+    checkpoint_path = str(temp_dir / "model.pth")
+    model = resnet18()
+    torch.save(model, checkpoint_path)
+    yield checkpoint_path
+
+
+@pytest.fixture()
+def recipe_path():
+    """
+    return: path to a resnet recipe
+    """
+    yield str(Path(__file__).resolve().parent / "resnet_test_recipe.yaml")
+
+
+def test_export_one_shot(resnet_checkpoint, recipe_path, temp_dir):
+    runner = CliRunner()
+    arch_key = "resnet18"
+    result = runner.invoke(
+        main,
+        [
+            "--arch-key",
+            arch_key,
+            "--checkpoint-path",
+            resnet_checkpoint,
+            "--num-samples",
+            0,
+            "--save-dir",
+            temp_dir,
+            "--num-classes",
+            1000,
+            "--one-shot",
+            recipe_path,
+        ],
+    )
+    assert result.exit_code == 0
+
+    expected_onnx_path = str(temp_dir / arch_key / "deployment" / "model.onnx")
+
+    # exported file exists
+    assert Path(expected_onnx_path).exists()
+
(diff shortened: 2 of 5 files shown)


src/sparseml/pytorch/image_classification/export.py
```diff
@@ -290,6 +297,13 @@ class strings. Expected to be a path to a
     show_default=True,
     help="number of classes for model; must be set if a dataset is not provided",
 )
+@click.option(
+    "--one-shot",
+    "--one_shot",
+    is_flag=True,
+    show_default=True,
```
corey-nm, line 304: missing default value?

src/sparseml/pytorch/image_classification/export.py
```diff
@@ -366,9 +381,18 @@ def main(
...
         labels_to_class_mapping = helpers.label_to_class_mapping_from_dataset(dataset)
 
-    if recipe is not None:
-        ScheduledModifierManager.from_yaml(recipe).apply_structure(model)
-        load_model(checkpoint_path, model, strict=True)
+    # training recipe must be a part of checkpoint
+    load_model(checkpoint_path, model, strict=True)
+
+    if one_shot:
+        if not recipe:
```
corey-nm, line 388: Do we need to cover the case where `recipe==True` and `one_shot==False`?

tests/sparseml/pytorch/image_classification/test_export.py
```diff
@@ -0,0 +1,90 @@
...
+import pytest
+import torch
+
+from click.testing import CliRunner
+from sparseml.pytorch.image_classification.export import main
+from sparseml.pytorch.models import resnet50
+
+
+@pytest.fixture()
+def temp_dir():
```
corey-nm, line 28: you can just use the built in `tmp_path` fixture instead of this 😀

tests/sparseml/pytorch/image_classification/test_export.py
```diff
@@ -0,0 +1,90 @@
...
+
+    # exported file exists
+    assert os.path.exists(expected_onnx_path)
+
+    # exported file is valid onnx
+    model = onnx.load(expected_onnx_path)
+    onnx.checker.check_model(model)
+
+    # check onnx model is quantized
+    assert any(node.op_type == "QuantizeLinear" for node in model.graph.node)
```
corey-nm, line 90: nice test!! 🎉

src/sparseml/pytorch/image_classification/export.py
rahul-tuli: Flag always has a default value of `False`

src/sparseml/pytorch/image_classification/export.py
rahul-tuli: Good catch! Updating.

tests/sparseml/pytorch/image_classification/test_export.py
rahul-tuli: Thanks!

src/sparseml/pytorch/image_classification/export.py
```diff
@@ -368,7 +386,13 @@ def main(
 
     if recipe is not None:
         ScheduledModifierManager.from_yaml(recipe).apply_structure(model)
+
+    if checkpoint_path:
         load_model(checkpoint_path, model, strict=True)
+
+    if one_shot is not None:
```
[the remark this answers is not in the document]
rahul-tuli: Yes, both can exist, recipe is generally a training recipe, at some point in our codebase we moved to saving the training recipe in the checkpoint itself like `{'state_dict': ..., 'recipe': ...}` the training recipe being provided separately is to support older flows;

This one shot export flow, is meant so that someone can quickly test how the model will look like after application of another recipe.

for example, the original training recipe could quantize some parts of the graph; later if a user wants to apply another recipe on top that quantizes different layers or weights, they could pass in the `new` recipe with the one shot argument; if they saved the checkpoint without training recipe they could pass in the `old` training recipe with the `--recipe` argument;
like the following:
```
--recipe old_training_recipe.md --one-shot new_one_shot_recipe.md
```

What would happen here essentially is that the training recipe will be used to apply structure to the model then load weights, after which the one shot recipe will be applied and the model will be exported. Does this make sense?

corey-nm (approved):
🎉

src/sparseml/pytorch/image_classification/export.py
```diff
@@ -368,7 +386,13 @@ def main(
 
     if recipe is not None:
         ScheduledModifierManager.from_yaml(recipe).apply_structure(model)
+
+    if checkpoint_path:
```
bfineran, line 390: what is the reason for this change? could this potentially fail before?

tests/sparseml/pytorch/image_classification/test_export.py
```diff
@@ -0,0 +1,89 @@
...
+def quant_recipe_path():
+    """
+    return: path to a resnet50 recipe with quantization
+    """
+    recipe_path = "./resnet_test_recipe.md"
+    recipe_path = os.path.abspath(recipe_path)
+    yield recipe_path
+
+
+def test_export_one_shot(non_quant_resnet_checkpoint, quant_recipe_path, temp_dir):
```
bfineran, line 57: we should be checking pruning here as well, especially since you're using a pruning recipe. (use the `tensor_sparsity` helper, see comments in the recipe template PR)

tests/sparseml/pytorch/image_classification/resnet_test_recipe.md
```diff
@@ -0,0 +1,207 @@
+<!--
```
bfineran, line 1: this file is unnecessary IMO - copied from a zoo recipe contains bloat such as non-relevant modifiers and hyper parameters as well as the markdown.

We should be able to just move this into a constant string in python

```yaml
_TEST_RECIPE_STR = """
modifiers:
  - !ACDCPruningModifier
    compression_sparsity: 0.9
    start_epoch: 0.0
    end_epoch: 1.0
    global_sparsity: True 
    update_frequency: 1.0
    params: __ALL_PRUNABLE__

  - !QuantizationModifier
    start_epoch: 1.0
    disable_quantization_observer_epoch: 2.0
    freeze_bn_stats_epoch: 2.0
"""
```
should suffice

bfineran (changes requested):
few quick changes to test then LGTM

src/sparseml/pytorch/image_classification/export.py
rahul-tuli: Before, the code wasn't loading in checkpoint weights if the recipe was not provided which was a bug; this fixes that

tests/sparseml/pytorch/image_classification/test_export.py
rahul-tuli: Tensor sparsity takes in a torch tensor, I used ModelAnalysis tool in upcoming commit to assert sparsity

tests/sparseml/pytorch/image_classification/resnet_test_recipe.md
rahul-tuli: Thanks for the minimal recipe, I would argue we should still keep this as a test file, point being a recipe file path is expected by the export script, we can always write the recipe to a temporary file on the fly during testing, let me know your thoughts on that

tests/sparseml/pytorch/image_classification/test_export.py
rahul-tuli: Thanks @corey-nm!

corey-nm (dismissed):
lgtm

tests/sparseml/pytorch/image_classification/test_export.py
```diff
@@ -0,0 +1,95 @@
...
+
+    # check onnx model is quantized
+    assert any(node.op_type == "QuantizeLinear" for node in model.graph.node)
+
+    # check onnx model is sparse
+    model_analysis = ModelAnalysis.from_onnx(onnx_file_path=expected_onnx_path)
+    block = model_analysis.operation_summary.ops.block_structure
+
+    assert block["single"].sparsity > 0.99
+    assert block["block4"].sparsity > 0.95
```
bfineran, line 95: this doesn't seem right to me - the recipe is only targeting 0.9% sparsity, which means the single and block sparsity should be much lower. additionally since the recipe isn't targeting block sparsity explicitly, we shouldn't be testing for it. any block sparsity is a side effect

bfineran (changes requested):

tests/sparseml/pytorch/image_classification/test_export.py
rahul-tuli: Updated!

KSGulin (approved):

tests/sparseml/pytorch/image_classification/test_export.py
```diff
@@ -0,0 +1,94 @@
...
+    onnx.checker.check_model(model)
+
+    # check onnx model is quantized
+    assert any(node.op_type == "QuantizeLinear" for node in model.graph.node)
+
+    # check onnx model is sparse
+    model_analysis = ModelAnalysis.from_onnx(onnx_file_path=expected_onnx_path)
+    block = model_analysis.operation_summary.ops.block_structure
+
+    assert block["single"].sparsity > 0.90
```
bfineran, line 94: this still doesn't quite add up (doesn't check the previous issue that the model was way too sparse, likely all 0s) - sparsity shouldn't be greater than 0.9 since we are targeting 0.9. should test that we are within a margin of error.

bfineran (changes requested):
another testing comment, also please look into the failing tests

tests/sparseml/pytorch/image_classification/test_export.py
rahul-tuli: What's an acceptable margin of error?

bfineran (approved):

bfineran:
bumping this

tests/sparseml/pytorch/image_classification/test_export.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def test_export_one_shot(non_quant_resnet_checkpoint, quant_recipe_path, temp_dir):
+    runner = CliRunner()
+    arch_key = "resnet50"
+    result = runner.invoke(
+        main,
+        [
+            "--arch-key",
+            "resnet50",
```
bfineran, line 64: let's change to resnet18 or mobilenet

bfineran (approved):

corey-nm (approved):
🔥

PR #1088, opened 2022-10-19
Allowing torch 1.10 and 1.11, but raising error with quantized exports

corey-nm:
Addresses comment in #1062

### DIFF
--- a/src/sparseml/pytorch/__init__.py
+++ b/src/sparseml/pytorch/__init__.py
@@ -17,6 +17,7 @@ Functionality for working with and sparsifying Models in the PyTorch framework
 """
 
 import os
+import warnings
 
 from packaging import version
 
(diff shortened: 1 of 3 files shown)


src/sparseml/pytorch/utils/exporter.py
```diff
@@ -571,6 +571,18 @@ def export_onnx(
...
+        if (
+            _PARSED_TORCH_VERSION.major == 1
+            and _PARSED_TORCH_VERSION.minor in [10, 11]
+            and "NM_FORCE_QUANT_EXPORT" not in os.environ
+        ):
+            raise RuntimeError(
+                "Quantized exports are not supported in torch==1.10.* or 1.11.*. "
+                f"Found torch version {torch.__version__}. "
+                "To override this error, set environment variable "
+                "`NM_FORCE_QUANT_EXPORT` to anything."
```
rahul-tuli, line 583: The condition would pass if `NM_FORCE_QUANT_EXPORT` is set to anything, but could we specify a value, like `NM_FORCE_QUANT_EXPORT=true` or `1` to make it a less confusing for users? Does that make sense?

rahul-tuli (approved):

src/sparseml/pytorch/utils/exporter.py
corey-nm: Updated

bfineran (changes requested):
we need to move this exception up - right now it's failing at the last minute possible.
* let's make this in general failing because torch 1.10 and 1.11 are not supported by sparseml (change env var name to something like ignore torch version and update message accordingly)
* move fail from right before export (after a user has already trained their model on an unsupported, unexportable version) to either the init.py or base.py file for spasreml/pytorch

src/sparseml/pytorch/__init__.py
```diff
@@ -16,6 +16,29 @@
...
+
+
+try:
+    import torch
+
+    _PARSED_TORCH_VERSION = version.parse(torch.__version__)
+    if (
+        _PARSED_TORCH_VERSION.major == 1
+        and _PARSED_TORCH_VERSION.minor in [10, 11]
+        and os.environ.get("NM_BYPASS_TORCH_VERSION", "").lower() == "true"
```
bfineran, line 31: shouldn't his be a not?
also in general for environment variables, we use integers for env var flags (ie 1/0)
```suggestion
        and not os.environ.get("NM_BYPASS_TORCH_VERSION", "")
```

bfineran (commented):
let's also raise a warning when the bypass is set that quantized onnx export may not work

src/sparseml/pytorch/__init__.py
corey-nm: Ahh okay I can change to int. `not` won't work because any non-empty string will resolve to True. so "0" would be True.

KSGulin (approved):

bfineran (approved):

PR #1089, opened 2022-10-19
Making onnx opset 13 default

corey-nm:
For torch-1.12, the default onnx opset should be 13. Unclear what the default should be for other versions.

### DIFF
--- a/integrations/keras/classification.py
+++ b/integrations/keras/classification.py
@@ -573,7 +573,7 @@ def parse_args():
         "--onnx-opset",
         type=int,
         default=13,
-        help="The onnx opset to use for export. Default is 11",
+        help="The onnx opset to use for export. Default is 13",
     )
     export_parser.add_argument(
         "--export-debug-mode",
--- a/integrations/keras/classification.py
+++ b/integrations/keras/classification.py
@@ -572,7 +572,7 @@ def parse_args():
     export_parser.add_argument(
         "--onnx-opset",
         type=int,
-        default=11,
+        default=13,
         help="The onnx opset to use for export. Default is 11",
     )
     export_parser.add_argument(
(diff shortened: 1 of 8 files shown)


integrations/keras/classification.py
```diff
@@ -572,7 +572,7 @@ def parse_args():
     export_parser.add_argument(
         "--onnx-opset",
         type=int,
-        default=11,
+        default=13,
         help="The onnx opset to use for export. Default is 11",
```
rahul-tuli, line 576:
```suggestion
        help="The onnx opset to use for export. Default is 13",
```

corey-nm:
> A missed 11 -> 13, LGTM otherwise

Great catch haha 🚀

src/sparseml/tensorflow_v1/utils/exporter.py
```diff
@@ -38,7 +38,7 @@
 
 
 def default_onnx_opset() -> int:
-    return 9 if onnx.__version__ < "1.6" else 11
+    return 9 if onnx.__version__ < "1.6" else 13
```
bfineran, line 41: is this a tested pathway? if not would prefer to not change

integrations/keras/classification.py
```diff
@@ -572,8 +572,8 @@ def parse_args():
     export_parser.add_argument(
         "--onnx-opset",
         type=int,
-        default=11,
-        help="The onnx opset to use for export. Default is 11",
+        default=13,
```
bfineran, line 575: let's keep this as is for now unless tested

integrations/tensorflow_v1/classification.py
```diff
@@ -546,8 +546,8 @@ def parse_args():
             par.add_argument(
                 "--onnx-opset",
                 type=int,
-                default=11,
-                help="The onnx opset to use for export. Default is 11",
+                default=13,
```
bfineran, line 549: let's keep this as is for now unless tested

bfineran (commented):
the tf and keras onnx exporters rely on third party libraries that may have not been updated in a bit, hesitant on updating without further testing

integrations/tensorflow_v1/classification.py
corey-nm: Will do

corey-nm:
Closing, will re-open a different one

Issue #1090, opened 2022-10-19
PruningModifiers invoked two times when start_epoch=0


eldarkurtic:
**Describe the bug**
When a pruning modifier has `start_epoch: 0`, it is invoked two times. The first time during the `def initialize(...)` call at https://github.com/neuralmagic/sparseml/blob/1360d075b4f61a3c536ab296dd63a87bcd577219/src/sparseml/pytorch/sparsification/pruning/modifier_pruning_base.py#L322
and the second time during the standard (expected) training-aware flow (i.e. when called from optimizer.step()).

The first call is meant to be used when modifiers are invoked in a one-shot manner only (i.e. when `epoch=inf` and `steps_per_epoch=1`).

**Expected behavior**
Pruner should be called only once when `start_epoch: 0`.

**To Reproduce**
Exact steps to reproduce the behavior: try to run any BERT-pruning example with the following recipe
```yaml
pruning_modifiers:
  - !GlobalMagnitudePruningModifier
    params:
      - re:bert.encoder.layer.*.attention.self.query.weight
      - re:bert.encoder.layer.*.attention.self.key.weight
      - re:bert.encoder.layer.*.attention.self.value.weight
      - re:bert.encoder.layer.*.attention.output.dense.weight
      - re:bert.encoder.layer.*.intermediate.dense.weight
      - re:bert.encoder.layer.*.output.dense.weight
    start_epoch: 0
    end_epoch: 2
    init_sparsity: 0.7
    final_sparsity: 0.9
    inter_func: cubic
    update_frequency: 0.5
    leave_enabled: True
    mask_type: unstructured
```

**Errors**
The pruner will be called two times at `epoch=0`, and it will prune the model two times with the same `target_sparsity`.

PR #1091, opened 2022-10-19
Add oneshot checks

eldarkurtic:
Fixes https://github.com/neuralmagic/sparseml/issues/1090  by explicitly adding one-shot checks. Also probably useful for future maintenance as it makes it clear what one-shot setup/config is.


rahul-tuli (approved):
Thanks for the quick fix!

src/sparseml/pytorch/sparsification/pruning/modifier_pruning_base.py
```diff
@@ -319,7 +325,8 @@ def initialize(
 
         self.initialize_extras(module)
 
-        self.check_mask_update(module, epoch, steps_per_epoch=1, **kwargs)
+        if self.is_oneshot(epoch, steps_per_epoch=1):
```
corey-nm, line 328: so this just checks if epoch is inf then? out of curiosity what's the point of steps_per_epoch here?

eldarkurtic: To the best of my knowledge, `steps_per_epoch=1` is convention used in the codebase for these "one-shot" use-cases. That's why the `check_mask_update` in the line below is also called with `steps_per_epoch=1` (when called like this from the `initialize` method).

`is_oneshot` method in this specific use-case will check if epoch is `inf`, but in general use-case it makes it explicit what we consider as one-shot throughout the codebase. For example, gradient-based pruners need to check somehow if they are called in one-shot or gradual setup, and (hopefully) this will be a unified way to do that by calling: `if self.is_oneshot(...)`.

corey-nm: Gotcha, thanks for the details!

corey-nm (approved):

PR #1092, opened 2022-10-20
[Transformers] Drop creation of HF model cards

KSGulin:
The sparseml transformers integration is broken as of `huggingface_hub` release v0.10. The new release added validators to model card creation which breaks our flow. Choosing to drop model card creation as it's not utilized in the sparseml flow.


bfineran (approved):

corey-nm (approved):
ez

rahul-tuli (approved):

PR #1093, opened 2022-10-20
Export with one shot recipe [transformers]

rahul-tuli:
This PR adds support to apply one shot recipe before exporting models from our transformers integration similar to sparseml diff #1084


KSGulin (approved):
Great feature to have

bfineran (approved):

PR #1095, opened 2022-10-20
Drop object detection datasets tests

KSGulin:
Dropping test coverage for object detection dataset classes as they're currently unused and not actively supported and the underlying data hosts may change over time. Also dropping empty testing directories for other uncovered dataset directories


corey-nm (approved):
Great diff 🚀

bfineran (approved):

PR #1098, opened 2022-10-24
[WIP] Forkless transformers

KSGulin:
This PR includes hot patching that allows transformers to work with sparseml without the need for an integration fork. In its current state it is purely for demonstration and discussion purposes


src/sparseml/transformers/hot_patching/models/utils.py
```diff
@@ -0,0 +1,77 @@
...
+    Drop in replacement for torch function with custom function, only during the scope
+    of forward_func execution. e.g. replaces torch.matmul with
+    QATOperation(torch.matmul) during execution of
+    transformers.models.bert.modeling_bert.BertSelfAttention.forward
+    """
+
+    def forward_wrapped(self, *args, **kwargs):
+        original_torch_func = _rgetattr(torch, original_torch_func_path)
+        patched_torch_func = _torch_operation_patch(original_torch_func)
+        _rsetattr(torch, original_torch_func_path, patched_torch_func)
```
corey-nm, line 60: So this is essentially doing
```python
original_mm = getattr(torch, "matmul")
setattr(torch, "matmul", QATOperation(original_mm))
```
when `original_torch_func_path` is "matmul"?

And the key is that this patches the function right before forward is called, so there can be multiple different versions of matmul?

PR #1101, opened 2022-10-24
Setting opset based on torch version for `sparseml/pytorch`

corey-nm:
2nd take at #1089

### DIFF
--- a/src/sparseml/pytorch/__init__.py
+++ b/src/sparseml/pytorch/__init__.py
@@ -22,10 +22,21 @@ import warnings
 from packaging import version
 
 
+def _default_opset(torch_version: version.Version) -> int:
+    if torch_version < version.parse("1.3"):
+        return 9
+    if torch_version < version.parse("1.10.0"):
+        return 11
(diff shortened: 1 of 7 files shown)


src/sparseml/pytorch/__init__.py
```diff
@@ -22,10 +22,21 @@
...
+        return 11
+    return 13
+
+
 try:
     import torch
 
     _PARSED_TORCH_VERSION = version.parse(torch.__version__)
+
+    TORCH_ONNX_OPSET = _default_opset(_PARSED_TORCH_VERSION)
```
bfineran, line 38: let's move this to a helpers file, in general we don't perform much logic/set variables in our init files so it's not a place that is usually looked for

bfineran, line 38: also would prefer to name to `TORCH_DEFAULT...`

bfineran (commented):
lgtm pending comments

bfineran (approved):

PR #1108, opened 2022-10-25
Adding --max-train-steps to torchvision train

corey-nm:


src/sparseml/pytorch/torchvision/train.py
```diff
@@ -51,8 +51,9 @@ def train_one_epoch(
     optimizer.zero_grad()
 
     header = f"Epoch: [{epoch}]"
-    for i, (image, target) in enumerate(
-        metric_logger.log_every(data_loader, args.print_freq, header)
+    for i, (image, target) in zip(
+        range(args.max_train_steps),
```
bfineran, line 55: would prefer to have an explicit if/break clause for readability, especially since it's already using `enumerate`

bfineran: this also let's us avoid needing to overwrite a default for max train steps

bfineran (changes requested):
let's do the same for `max_eval_steps`

src/sparseml/pytorch/torchvision/train.py
corey-nm: Check new logic

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -54,6 +55,8 @@ def train_one_epoch(
     for i, (image, target) in enumerate(
         metric_logger.log_every(data_loader, args.print_freq, header)
     ):
+        if i > steps_per_epoch:
```
corey-nm, line 58:
```suggestion
        if i >= steps_per_epoch:
```

KSGulin (approved):

bfineran (approved):

PR #1109, opened 2022-10-25
Adding --max-eval-steps to torchvision

corey-nm:


src/sparseml/pytorch/torchvision/train.py
```diff
@@ -95,14 +95,19 @@ def train_one_epoch(
...
     model.eval()
     metric_logger = utils.MetricLogger(delimiter="  ")
     header = f"Test: {log_suffix}"
 
     num_processed_samples = 0
     with torch.inference_mode():
-        for image, target in metric_logger.log_every(data_loader, print_freq, header):
+        for i, (image, target) in zip(
+            range(args.max_eval_steps),
+            metric_logger.log_every(data_loader, print_freq, header),
```
dbogunowicz, line 109: why not just `enumerate`?

corey-nm: the `zip(range(), ...)` makes the loop stop when the range runs out (ie when max_eval_steps is reached). its kinda like doing a slice on an interable (`data_loader[:max_eval_steps]`). maybe its too subtle?

corey-nm: Updated logic

KSGulin (approved):
LGTM

src/sparseml/pytorch/torchvision/train.py
rahul-tuli: args.max_eval_steps is always set and will be an int, I feel this idiom is very well accepted in python community, and it also works if the second iterator is smaller than the first.

A nit I have is to rename `i` to `_` because it's a throw away variable

rahul-tuli (approved):

PR #1110, opened 2022-10-25
Adding --one-shot to torchvision

corey-nm:


src/sparseml/pytorch/torchvision/train.py
```diff
@@ -444,6 +444,10 @@ def collate_fn(batch):
         return
 
     manager = ScheduledModifierManager.from_yaml(args.recipe_path)
+    if args.one_shot:
+        manager.apply(model)
+        return
```
rahul-tuli, line 449: What about saving the one shot checkpoint as in code block at line 476 - line 492

corey-nm: Added

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -466,6 +466,25 @@ def collate_fn(batch):
         return
 
     manager = ScheduledModifierManager.from_yaml(args.recipe_path)
+    if args.one_shot:
+        manager.apply(model)
+        checkpoint = {
```
bfineran, line 471: probably worth it to move the checkpoint building to a helper function since there will be repeated logic around what to save as well as recipe composition

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -466,6 +466,25 @@ def collate_fn(batch):
...
+            "model": model_without_ddp.state_dict(),
```
bfineran, line 472: need to save the (composed) recipe for model state reproduction on re-load

bfineran (changes requested):

src/sparseml/pytorch/torchvision/train.py
corey-nm: Good call. Will wait for checkpoint to be merged in. Closing this one for now to de-clutter

PR #1114, opened 2022-10-26
Add --save-best-after to torchvision

corey-nm:


src/sparseml/pytorch/torchvision/train.py
```diff
@@ -760,6 +765,13 @@ def get_args_parser(add_help=True):
         type=int,
         help="gradient accumulation steps",
     )
+    parser.add_argument(
+        "--save-best-after",
+        default=1,
```
rahul-tuli, line 770: Would `0` make for a better default?

rahul-tuli (approved):

bfineran (approved):

PR #1122, opened 2022-10-31
Adding pass to flatten zero_point/scale params

corey-nm:
As of torch 1.12, the zero_point/scale params are all arrays with shape `(1,)`. This causes some crashes for yolov5 integration tests (but for some reason not other models).

This PR flattens all the zero_point/scale inits to shape `()` if they have shape `(1,)`

Testing plan: manual export of yolov5. I can probably add a simple unit test for this too

### DIFF
--- a/src/sparseml/pytorch/utils/exporter.py
+++ b/src/sparseml/pytorch/utils/exporter.py
@@ -556,6 +556,7 @@ def export_onnx(
     # onnx file fixes
     onnx_model = onnx.load(file_path)
     _fold_identity_initializers(onnx_model)
+    _flatten_qparams(onnx_model)
     if batch_norms_wrapped:
         # fix changed batch norm names
         _unwrap_batchnorms(onnx_model)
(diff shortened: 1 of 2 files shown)


src/sparseml/pytorch/utils/exporter.py
```diff
@@ -662,6 +663,36 @@ def _save_label_to_class_mapping(
...
+                # zero_point is optional AND shape must match
+                # scale. so if scale is (1,), then so will zero point
+                if len(node.input) == 3:
+                    inits_to_flatten.add(node.input[2])
+
+    for i, init in enumerate(model.graph.initializer):
+        if init.name not in inits_to_flatten:
+            continue
+        a = numpy_helper.to_array(init)
+        assert a.shape == (1,)
```
bfineran, line 689: why are we asserting here? if the shape were say (8,) for a channel quantized model this may crash

corey-nm: the only things that get added to inits_to_flatten should have this shape (per the `list(init.dims) == [1]` above)

bfineran (approved):

KSGulin (approved):

KSGulin:
Glad to have this fixed 🥳

PR #1124, opened 2022-11-01
Adding checkpoint_manager to torchvision

corey-nm:
Did i oversimplify this?

### DIFF
--- a/src/sparseml/pytorch/torchvision/train.py
+++ b/src/sparseml/pytorch/torchvision/train.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import datetime
+import math
 import os
 import time
 import warnings
@@ -430,6 +431,13 @@ def main(args):
         if scaler:
             scaler.load_state_dict(checkpoint["scaler"])
 
+        checkpoint_manager = ScheduledModifierManager.from_yaml(
+            checkpoint["checkpoint_recipe"]
+        )
+        checkpoint_manager.apply_structure(model_without_ddp, math.inf)
+    else:
+        checkpoint_manager = None
+
     if args.test_only:
         # We disable the cudnn benchmarking because it can
         # noticeably affect the accuracy
@@ -480,6 +488,13 @@ def main(args):
                 checkpoint["model_ema"] = model_ema.state_dict()
             if scaler:
                 checkpoint["scaler"] = scaler.state_dict()
+
+            checkpoint["checkpoint_recipe"] = str(
+                ScheduledModifierManager.compose_staged(checkpoint_manager, manager)
+                if checkpoint_manager is not None
+                else manager
+            )
+
             utils.save_on_master(
                 checkpoint, os.path.join(args.output_dir, f"model_{epoch}.pth")
             )
--- a/src/sparseml/pytorch/torchvision/train.py
+++ b/src/sparseml/pytorch/torchvision/train.py
@@ -455,15 +455,6 @@ def main(args):
             checkpoint["checkpoint_recipe"]
         )
         checkpoint_manager.apply_structure(model, epoch=checkpoint["epoch"])
-
-        # load values
-        model.load_state_dict(checkpoint["model"])
-        optimizer.load_state_dict(checkpoint["optimizer"])
-        lr_scheduler.load_state_dict(checkpoint["lr_scheduler"])
-        if model_ema:
-            model_ema.load_state_dict(checkpoint["model_ema"])
-        if scaler:
-            scaler.load_state_dict(checkpoint["scaler"])
     elif args.resume:
         checkpoint = _load_checkpoint(args.resume)
 
@@ -472,20 +463,23 @@ def main(args):
         checkpoint_manager = None
         manager.initialize(model, epoch=checkpoint["epoch"])
 
-        # load params
-        model.load_state_dict(checkpoint["model"])
-        optimizer.load_state_dict(checkpoint["optimizer"])
-        lr_scheduler.load_state_dict(checkpoint["lr_scheduler"])
-        if model_ema:
-            model_ema.load_state_dict(checkpoint["model_ema"])
-        if scaler:
-            scaler.load_state_dict(checkpoint["scaler"])
-
         # NOTE: override start epoch
         args.start_epoch = checkpoint["epoch"] + 1
+
+        lr_scheduler.load_state_dict(checkpoint["lr_scheduler"])
     else:
+        checkpoint = None
         manager = ScheduledModifierManager.from_yaml(args.recipe_path)
         checkpoint_manager = None
+
+    # load params
+    if checkpoint is not None:
+        model.load_state_dict(checkpoint["model"])
+        optimizer.load_state_dict(checkpoint["optimizer"])
+        if model_ema and "model_ema" in checkpoint:
+            model_ema.load_state_dict(checkpoint["model_ema"])
+        if scaler and "scaler" in checkpoint:
+            scaler.load_state_dict(checkpoint["scaler"])
 
     optimizer = manager.modify(model, optimizer, len(data_loader))
 
--- a/src/sparseml/pytorch/torchvision/train.py
+++ b/src/sparseml/pytorch/torchvision/train.py
@@ -237,6 +237,11 @@ def load_data(traindir, valdir, args):
 
 
 def main(args):
+    if args.resume is not None and args.checkpoint_path is not None:
+        raise ValueError(
+            "Only one of --resume or --checkpoint-path can be specified, not both."
+        )
+
     if args.output_dir:
         utils.mkdir(args.output_dir)
 
@@ -419,8 +424,10 @@ def main(args):
             model_without_ddp, device=device, decay=1.0 - alpha
         )
 
-    if args.resume:
-        checkpoint = torch.load(args.resume, map_location="cpu")
+    manager = ScheduledModifierManager.from_yaml(args.recipe_path)
+
+    if args.checkpoint_path:
+        checkpoint = torch.load(args.checkpoint_path, map_location="cpu")
         model_without_ddp.load_state_dict(checkpoint["model"])
         if not args.test_only:
             optimizer.load_state_dict(checkpoint["optimizer"])
@@ -434,7 +441,25 @@ def main(args):
         checkpoint_manager = ScheduledModifierManager.from_yaml(
             checkpoint["checkpoint_recipe"]
         )
-        checkpoint_manager.apply_structure(model_without_ddp, math.inf)
+        checkpoint_manager.apply_structure(model_without_ddp, epoch=checkpoint["epoch"])
+    elif args.resume:
+        checkpoint = torch.load(args.resume, map_location="cpu")
+        model_without_ddp.load_state_dict(checkpoint["model"])
+        if not args.test_only:
+            optimizer.load_state_dict(checkpoint["optimizer"])
+            lr_scheduler.load_state_dict(checkpoint["lr_scheduler"])
+        args.start_epoch = checkpoint["epoch"] + 1
+        if model_ema:
+            model_ema.load_state_dict(checkpoint["model_ema"])
+        if scaler:
+            scaler.load_state_dict(checkpoint["scaler"])
+
+        # NOTE: override manager with the checkpoint's manager
+        manager = ScheduledModifierManager.from_yaml(checkpoint["checkpoint_recipe"])
+        checkpoint_manager = None
+
+        # TODO do we still do this?
+        manager.apply_structure(model_without_ddp, epoch=checkpoint["epoch"])
     else:
         checkpoint_manager = None
 
@@ -451,7 +476,7 @@ def main(args):
             evaluate(model, criterion, data_loader_test, device=device)
         return
 
-    manager = ScheduledModifierManager.from_yaml(args.recipe_path)
+    # TODO do we still need this in case of resume or checkpoint_path?
     optimizer = manager.modify(model, optimizer, len(data_loader))
 
     print("Start training")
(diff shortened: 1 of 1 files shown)


KSGulin:
In the current IC implementation, we require recipes for all training runs (in Sparseml, this is unique to IC). Are we keeping that rule in place here? For now I'll assume that we are to make this simpler.

There are two use-cases where a user may "resume" a run, but they might not be both covered by same CLI arg(s). 

1. A user wants to apply a new recipe on top of an existing checkpoint. This use case is well covered in PR.
2. A user has a run interrupt for any number of reasons and would like to continue it. This use case is uncovered in the PR.

The easiest way to add coverage of case #2 is to just apply the recipe up to the epoch that was saved with the checkpoint.

But there is an extra complication: when sparse transfer learning from a quantized model, we want to start from an unquantized baseline. So we save the last model before quantization as part of the model card and upon loading the model for continued training, we load this unquantized checkpoint. At this point we have to be careful not to apply the checkpoint recipe to the last epoch trained, because we'll re-quantized the model that way. (@bfineran is this still the approach we're planning on taking to sparse transfer learning from quantized models?)

Nit: IMO `--resume` is an indicator of case #2 and not case #1. Case #1 would be covered by a `--checkpoint-path` arg instead. But this might be a conversation to have with Product

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -430,6 +431,13 @@ def collate_fn(batch):
         if scaler:
             scaler.load_state_dict(checkpoint["scaler"])
 
+        checkpoint_manager = ScheduledModifierManager.from_yaml(
+            checkpoint["checkpoint_recipe"]
+        )
+        checkpoint_manager.apply_structure(model_without_ddp, math.inf)
```
corey-nm, line 437:
```suggestion
        checkpoint_manager.apply_structure(model_without_ddp, checkpoint["epoch"])
```

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -430,6 +431,13 @@ def collate_fn(batch):
         if scaler:
             scaler.load_state_dict(checkpoint["scaler"])
 
+        checkpoint_manager = ScheduledModifierManager.from_yaml(
```
corey-nm, line 434: If case 2 from @KSGulin then this would overwrite `manager`

corey-nm: and also the range for training should be `range(checkpoint["epoch"] + 1, ...)`

corey-nm: Noting that start_epoch is overwriting in this block already

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -480,6 +488,13 @@ def collate_fn(batch):
                 checkpoint["model_ema"] = model_ema.state_dict()
             if scaler:
                 checkpoint["scaler"] = scaler.state_dict()
+
+            checkpoint["checkpoint_recipe"] = str(
```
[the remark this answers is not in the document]
corey-nm: Saving as -1 now

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -418,7 +424,25 @@ def collate_fn(batch):
             model_without_ddp, device=device, decay=1.0 - alpha
         )
 
-    if args.resume:
+    manager = ScheduledModifierManager.from_yaml(args.recipe_path)
+
+    if args.checkpoint_path:
```
corey-nm, line 429: Do all these blocks need to check for -1 epoch in checkpoint?

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -430,6 +454,15 @@ def collate_fn(batch):
         if scaler:
             scaler.load_state_dict(checkpoint["scaler"])
 
+        # NOTE: override manager with the checkpoint's manager
+        manager = ScheduledModifierManager.from_yaml(checkpoint["checkpoint_recipe"])
+        checkpoint_manager = None
+
+        # TODO do we still do this?
+        manager.apply_structure(model_without_ddp, epoch=checkpoint["epoch"])
```
corey-nm, line 462: @KSGulin if resuming, do we still do this apply_structure since the model is loaded above? And also how do checkpoint_path/resume affect the manager.modify below - does that only happen if we are NOT doing either resume/checkpoint_path?

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -418,7 +424,25 @@ def collate_fn(batch):
...
+        args.start_epoch = checkpoint["epoch"] + 1
+        if model_ema:
+            model_ema.load_state_dict(checkpoint["model_ema"])
+        if scaler:
+            scaler.load_state_dict(checkpoint["scaler"])
+
+        checkpoint_manager = ScheduledModifierManager.from_yaml(
+            checkpoint["checkpoint_recipe"]
+        )
+        checkpoint_manager.apply_structure(model_without_ddp, epoch=checkpoint["epoch"])
```
bfineran, line 444: all model loading should go through here for consistency: https://github.com/neuralmagic/sparseml/blob/21b48cc4fa8e6a1ba183b6f1ad0fb221b9fc3ff8/src/sparseml/pytorch/utils/model.py#L58

src/sparseml/pytorch/torchvision/train.py
bfineran, line 462: instead should `initialize` the manager instead of `apply_structure` so that we get to the exact model and manager state at the given epoch. (apply structure involves calling into both `initialize` and `finalize`)

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -443,7 +476,7 @@ def collate_fn(batch):
             evaluate(model, criterion, data_loader_test, device=device)
         return
 
-    manager = ScheduledModifierManager.from_yaml(args.recipe_path)
+    # TODO do we still need this in case of resume or checkpoint_path?
```
bfineran, line 479: in resume, prefer to use initialize (modify ok if it's just initializing, check code)

src/sparseml/pytorch/torchvision/train.py
bfineran: after the manager is initialized and the model object structure is recovered, it's safe to call into `model.load_state_dict`, not safe before on resume

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -473,13 +506,20 @@ def collate_fn(batch):
                 "model": model_without_ddp.state_dict(),
                 "optimizer": optimizer.state_dict(),
                 "lr_scheduler": lr_scheduler.state_dict(),
-                "epoch": epoch,
+                "epoch": -1 if epoch == manager.max_epochs - 1 else epoch,
```
bfineran, line 509: epoch here should factor in composed staged (ie add the max epoch for the checkpoint recipe)

bfineran (changes requested):

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -415,21 +417,51 @@ def collate_fn(batch):
...
+        manager = ScheduledModifierManager.from_yaml(checkpoint["checkpoint_recipe"])
+        checkpoint_manager = None
+
+        manager.initialize(model, epoch=checkpoint["epoch"])
+    else:
+        manager = ScheduledModifierManager.from_yaml(args.recipe_path)
+        checkpoint_manager = None
+
+    # TODO do we still need this in case of resume or checkpoint_path?
+    optimizer = manager.modify(model, optimizer, len(data_loader))
```
corey-nm, line 463: @bfineran still have a question here on if this should be done for all three cases (resume, checkpoint_path, start from scratch)

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -415,21 +417,51 @@ def collate_fn(batch):
...
+        checkpoint = _load_checkpoint(args.checkpoint_path)
+        model.load_state_dict(checkpoint["model"])
+        optimizer.load_state_dict(checkpoint["optimizer"])
+        lr_scheduler.load_state_dict(checkpoint["lr_scheduler"])
+        if model_ema:
+            model_ema.load_state_dict(checkpoint["model_ema"])
+        if scaler:
+            scaler.load_state_dict(checkpoint["scaler"])
+
         args.start_epoch = checkpoint["epoch"] + 1
```
corey-nm, line 433: does this need to do something if epoch is -1? is that possible? same question for resume block

bfineran: yeah, we should only override start epoch on reload, in reload mode epoch should not be -1

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -415,21 +417,51 @@ def collate_fn(batch):
...
         if scaler:
             scaler.load_state_dict(checkpoint["scaler"])
 
+        args.start_epoch = checkpoint["epoch"] + 1
+
+        # NOTE: override manager with the checkpoint's manager
+        manager = ScheduledModifierManager.from_yaml(checkpoint["checkpoint_recipe"])
+        checkpoint_manager = None
+
+        manager.initialize(model, epoch=checkpoint["epoch"])
```
bfineran, line 457: to clarify - model state dict load in resume should come after the checkpoint recipe is initialized (ie after graph state is reproduced)

src/sparseml/pytorch/torchvision/train.py
bfineran: bumping this since marked outdated

src/sparseml/pytorch/torchvision/train.py
bfineran: in resume, we should be using the initialized checkpoint manager (can call modify up there with the optimizer)

bfineran (changes requested):
few comments, also let's move the logic for loading/saving with the composed recipes into helper functions (readability/reuse/testing)

src/sparseml/pytorch/torchvision/train.py
corey-nm: Updated - check if new logic is correct

src/sparseml/pytorch/torchvision/train.py
corey-nm: I kept this in - otherwise the optimizer isn't changed without this call?

src/sparseml/pytorch/torchvision/train.py
corey-nm: Yep in resume the manager gets set to the checkpoint's recipe (i.e. the checkpoint manager)

src/sparseml/pytorch/torchvision/train.py
corey-nm: I think I did this right - can you verify?

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -432,21 +434,52 @@ def collate_fn(batch):
...
+        manager = ScheduledModifierManager.from_yaml(args.recipe_path)
+        checkpoint_manager = ScheduledModifierManager.from_yaml(
+            checkpoint["checkpoint_recipe"]
+        )
+        checkpoint_manager.apply_structure(model, epoch=checkpoint["epoch"])
+
+        # load values
+        model.load_state_dict(checkpoint["model"])
+        optimizer.load_state_dict(checkpoint["optimizer"])
+        lr_scheduler.load_state_dict(checkpoint["lr_scheduler"])
```
bfineran, line 453: any reason these can't be shared between the two cases? Also thoughts on not resuming lr_scheduler state for a fresh run? I think optimizer/ema/scaler are ok

bfineran (changes requested):
refactor + logic look good, see single coment

additionally after looking at this code + the changes in the other PRs, we really should be pulling out all of these relevant sparseml logic changes/controls into a seaparate helper file/class to call in to. This will give us a pathway to test as well as increase readability of the base training script

corey-nm:
> additionally after looking at this code + the changes in the other PRs, we really should be pulling out all of these relevant sparseml logic changes/controls into a seaparate helper file/class to call in to. This will give us a pathway to test as well as increase readability of the base training script

Agree, will do that in a separate PR after functionality is merged in - that way we can keep the individual PRs as minimal as possible to model future integrations after. And then after the fact have a single PR which shows how to simplify them

src/sparseml/pytorch/torchvision/train.py
corey-nm: Pulled out into separate block

src/sparseml/pytorch/torchvision/train.py
bfineran: looks good

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -441,21 +443,46 @@ def collate_fn(batch):
...
+            model, device=device, decay=1.0 - alpha
         )
 
-    if args.resume:
-        checkpoint = torch.load(args.resume, map_location="cpu")
-        model_without_ddp.load_state_dict(checkpoint["model"])
-        if not args.test_only:
-            optimizer.load_state_dict(checkpoint["optimizer"])
-            lr_scheduler.load_state_dict(checkpoint["lr_scheduler"])
+    if args.checkpoint_path:
```
bfineran, line 449: looks functionally correct for now - but would be nice to break out into a helper function that takes in the checkpoint, resume, and model and returns the loaded model and manager(s)

right now, the overall flow is a bit dense (line 449-481), especially since it happens in two stages managing different requirements. might be good to have just one stage for each (checkpoint, resume, base)

src/sparseml/pytorch/torchvision/train.py
bfineran: looks good - see comment about potentially refactoring

bfineran (approved):
Approving as logic flow looks correct (definitely need to manually test these as a first priority, and then add in unit tests post refactor). as mentioned in comment I think this will look really good after a refactor, noted your comment that this will happen in a follow-up

PR #1128, opened 2022-11-03
Passing quant_min/quant_max to observer in torch 1.9

corey-nm:
See comment in code to what is different.

Testing:

This test was ripped from QAs test, I am unsure what went into this test.
```python
from transformers.models.bert.modeling_bert import QATMatMul
from sparseml.pytorch.sparsification import QuantizationModifier
import torch

dummy_sequential = torch.nn.Sequential(
    QATMatMul()
)  # give QATMatMul a layer to be wrapped
QuantizationModifier().apply(dummy_sequential)
qat_matmul = dummy_sequential[0]
_ = qat_matmul(torch.randn(10, 10), torch.randn(10, 10))

# affected zero point, 127 before fix, 128 after
zp = qat_matmul.input_quant_stubs[1].activation_post_process.zero_point
assert zp[0] == 128, zp
```

Status of this test:
- torch 1.9 on main currently passes this
- torch 1.12 on main currently fails this, the zero point has a value of 127
- **AFTER THIS PR, 1.9 also fails this test with the same error of 1.12**

I am unsure of the correct/expected behavior. Does 1.9 have a bug or does 1.12?

### DIFF
--- a/src/sparseml/pytorch/sparsification/quantization/helpers.py
+++ b/src/sparseml/pytorch/sparsification/quantization/helpers.py
@@ -18,6 +18,7 @@ Helper functions for performing quantization aware training with PyTorch
 
 from copy import deepcopy
 from dataclasses import dataclass, field
+from functools import partial
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import torch
@@ -31,7 +32,7 @@ from sparseml.pytorch.utils import get_layer
 
 
 _PARSED_TORCH_VERSION = version.parse(torch.__version__)
-
+_TORCH_PRE_112 = _PARSED_TORCH_VERSION < version.parse("1.12.0")
 
 __all__ = [
     "QATWrapper",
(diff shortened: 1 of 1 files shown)


corey-nm:
This seems to be exactly the problem found in https://github.com/neuralmagic/sparseml/pull/604. Notably that PR has the test in question and discusses the 127 vs 128 problem.

corey-nm:
@bfineran @KSGulin Based on PR I linked to above, I propose:
1. remove this function `def fix_observer_quant_range(module: Module):` in favor of the functools partial approach in this PR
2. Use the 127 as zero point for uint8
3. Add unit tests for different dtype ranges to ensure the correct zero point is found

KSGulin:
As per [this commit in pytorch](https://github.com/pytorch/pytorch/commit/7c1f3cc89e5ab335c638a4879748043698c5eff4), it seems they've adopted the philosophy of no longer having a default quant range for the observer as it doesn't correspond to all dtypes. In the same commit they fixed the issue with FakeQuantize min/max not getting propagated to the observer. 

So we have a situation where the default behavior in 1.12 for symmetric FakeQuant gets us a zero point of 127, but we expect 128. Our patched behavior in >1.12 is to keep the default of 128 for symmetric quant. There is an open question of whether we want to maintain a dynamically determined zero point of 127 or an enforced default of 128. 

This PR in its current state addresses the issue if we assume the former (127), but I believe a better solution in that case would be to just revert #604

corey-nm:
Closing as we want to maintain 128 zero point, will re-open with different approach

PR #1129, opened 2022-11-03
Ensuring proper has_custimzed_qrange for torch <=1.9.1 and >=1.12.0

corey-nm:
See comments in code for more details.

Related to #604 

Test plan: added unit tests and ran on torch 1.9 and 1.12.0. The GHW will also run this test on both versions.


src/sparseml/pytorch/sparsification/quantization/helpers.py
```diff
@@ -572,14 +572,45 @@ def get_observer(
...
+    # 1. customized qrange has zero point of 127
+    # 2. non-customized has zero point of 128.
+    # source:
+    # https://github.com/pytorch/pytorch/blob/v1.12.1/torch/ao/quantization/observer.py#L293
+    #
+    # **we want to ensure that the zero point is 128**
+    # see https://github.com/neuralmagic/sparseml/pull/604
+    #
+    # NOTE: This assumes we *never* want a customized qrange, and that the
+    #       compute_range above doesn't return customized qranges
```
corey-nm, line 602: Is this a safe assumption?

src/sparseml/pytorch/sparsification/quantization/helpers.py
```diff
@@ -588,44 +619,6 @@ def get_observer(
     return observer
 
 
-def fix_observer_quant_range(module: Module):
-    """
-    As of torch 1.10.2 there is a bug in FakeQuantize initialization where
-    quant_min and quant_max of FakeQuantize are not propagated to its
-    activation_post_process observer. This function propagates FakeQuantize quant
-    ranges to their Observer objects
```
corey-nm, line 596: It's only safe to remove this function if we always use non-customized ranges. Otherwise I can do a functools.partial above to properly pass the values

src/sparseml/pytorch/sparsification/quantization/helpers.py
```diff
@@ -588,44 +631,6 @@ def get_observer(
     return observer
 
 
-def fix_observer_quant_range(module: Module):
```
corey-nm, line 591: we don't need this anymore because the functools.partial above handles custom qranges

bfineran (dismissed):
tests look good

src/sparseml/pytorch/sparsification/quantization/helpers.py
```diff
@@ -571,15 +572,57 @@ def get_observer(
...
     observer_kwargs = dict(
-        observer=torch_quantization.MovingAverageMinMaxObserver,
-        quant_min=quant_min,
-        quant_max=quant_max,
         dtype=dtype,
         qscheme=qscheme,
         reduce_range=reduce_range,
     )
+
+    # in torch 1.9.1, quant_min and quant_max are not passed to observer:
```
dbogunowicz, line 584: For readability, why not put it between triple quotation marks? Will it be picked up by the doc parsing? If not, I'd like that, instead of commenting separate lines out. for readability.

dbogunowicz (dismissed):
one nit in comments

src/sparseml/pytorch/sparsification/quantization/helpers.py
dbogunowicz: I am almost sure your first sentence is correct (triple quotes only get picked up if they appear in the docstring context)

corey-nm: Updated

KSGulin (approved):
Really nice solution 🙂

PR #1130, opened 2022-11-04
fix registration of torchvision models for torchvsion>=0.13

bfineran:
torchvision changes the keyword for specifying pretrained models after their 0.13 release from `pretrained` to `weights`. This caused an issue in the registration of those models to our model registry, breaking our tests against torchvision>=0.13.

this change allows searches for the "weights" keyword as well to register the models

**test_plan:**
previously broken tests `tests/sparseml/pytorch/models/external/test_torchvision.py` fixed after resolution. tested locally against torchvision 0.13 and 0.10


bfineran:
duplicate of #1126, closing

PR #1132, opened 2022-11-07
[Refactor] export dataloader for token-classification

rahul-tuli:
propagate changes for token classification

### DIFF
--- a/src/sparseml/transformers/token_classification.py
+++ b/src/sparseml/transformers/token_classification.py
@@ -26,13 +26,18 @@ Fine-tuning the library models for token classification
 import logging
 import os
 import sys
+from contextlib import nullcontext
 from dataclasses import dataclass, field
-from typing import Optional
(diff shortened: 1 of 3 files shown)


src/sparseml/transformers/token_classification.py
```diff
@@ -730,6 +575,296 @@ def compute_metrics(p):
...
+
+def get_tokenized_token_classification_dataset(
+    data_args: DataTrainingArguments,
+    tokenizer: transformers.PreTrainedTokenizerBase,
+    model: Module,
+    cache_dir: Optional[str] = None,
+    do_train: bool = False,
+    do_eval: bool = True,
+    do_predict: bool = False,
+    main_process_func=None,
```
bfineran, line 630: why include these if never used?

bfineran (approved):
LGTM overall, must kick off an example training run and document in comment before merging

PR #1136, opened 2022-11-08
Using ModelRegistry for torchvision

corey-nm:


src/sparseml/pytorch/torchvision/train.py
```diff
@@ -217,15 +222,11 @@ def load_data(traindir, valdir, args):
         print(f"Loading dataset_test from {cache_path}")
         dataset_test, _ = torch.load(cache_path)
     else:
-        if args.weights and args.test_only:
-            weights = torchvision.models.get_weight(args.weights)
-            preprocessing = weights.transforms()
```
corey-nm, line 222: Is this okay to delete? This is something that isn't in existing image_classification training scripts, but seems like a good feature in torchvision (though this requires 0.13.0+ I believe)

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -316,8 +317,13 @@ def collate_fn(batch):
         steps_per_eval = min(steps_per_eval, args.max_eval_steps)
 
     print("Creating model")
-    model = torchvision.models.get_model(
-        args.model, weights=args.weights, num_classes=num_classes
+    model = _create_model(
+        key=args.arch_key,
+        pretrained=args.pretrained,
+        num_classes=num_classes,
+        checkpoint_path=None,  # TODO fill this in later once checkpoints are merged
```
corey-nm, line 324: This depends on #1124 being merged

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -25,9 +25,14 @@
 from torch.utils.data.dataloader import default_collate
 from torchvision.transforms.functional import InterpolationMode
 
+from sparseml.pytorch.image_classification.utils.helpers import (
+    _download_model_from_zoo_using_recipe,
```
rahul-tuli, line 29: could we make this imported function non private?

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -537,6 +536,38 @@ def collate_fn(batch):
     print(f"Training time {total_time_str}")
 
 
+def _create_model(args, num_classes):
```
rahul-tuli, line 539: Would it make more sense to have individual params? or maybe a class here? right now this function needs to know the contents of args namespace, which could be difficult to debug later

src/sparseml/pytorch/utils/helpers.py
```diff
@@ -981,13 +981,13 @@ def set_deterministic_seeds(seed: int = 0):
...
 @contextmanager
-def torch_distributed_zero_first(local_rank: int):
+def torch_distributed_zero_first(local_rank: Optional[int]):
     """
     Decorator to make all processes in distributed training wait for each
     local 0 ranked process to do something.
     :param local_rank: the local rank of this process
     """
-    if local_rank not in [-1, 0]:
+    if local_rank is not None and local_rank not in [-1, 0]:
```
rahul-tuli, line 990: nice

src/sparseml/pytorch/torchvision/train.py
corey-nm: yeah i can separate it out

src/sparseml/pytorch/torchvision/train.py
corey-nm: Oh wait we should be using a non image_classification helper for this, good catch

src/sparseml/pytorch/torchvision/train.py
corey-nm: Updated logic - moved rank management outside this function which should make it easier to debug/test

src/sparseml/pytorch/torchvision/train.py
corey-nm: Updated to use non-private, thanks!

src/sparseml/pytorch/torchvision/train.py
bfineran: fine for now - why in this PR?

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -537,6 +547,43 @@ def collate_fn(batch):
...
 
+def _create_model(
+    arch_key: str,
+    num_classes: int,
+    checkpoint_path: Optional[str],
+    recipe_path: Optional[str],
+    pretrained: bool,
+    pretrained_dataset: Optional[str],
+):
+    if checkpoint_path and checkpoint_path.startswith("zoo"):
```
bfineran, line 558: the zoo logic already happens with `ModelRegistry.create`, no? why duplicate here?

src/sparseml/pytorch/torchvision/train.py
corey-nm: Just noticed it - haven't hit that branch yet during my simple test runs so far

src/sparseml/pytorch/torchvision/train.py
corey-nm: Oh I didn't realize that - this was copied from the image_classification stuff, I can just remove this then

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -323,9 +324,20 @@ def collate_fn(batch):
         steps_per_eval = min(steps_per_eval, args.max_eval_steps)
 
     print("Creating model")
-    model = torchvision.models.get_model(
-        args.model, weights=args.weights, num_classes=num_classes
-    )
+    if args.arch_key in ModelRegistry.available_keys():
+        with torch_distributed_zero_first(args.rank if args.distributed else None):
```
dbogunowicz, line 328: Let's explain why: the first process is downloading the model from sparsezoo (other processes need to wait until that's done)

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -596,11 +608,44 @@ def get_args_parser(add_help=True):
...
+        "--arch-key",
+        default=None,
+        type=str,
+        help=(
+            "The architecture key for image classification model; "
+            "example: `resnet50`, `mobilenet`. "
+            "Note: Will be read from the checkpoint if not specified"
+        ),
+    )
+    parser.add_argument(
```
dbogunowicz, line 626: Maybe not in the context of this PR, but why do we have this? Possiby delete?

src/sparseml/pytorch/utils/helpers.py
dbogunowicz, line 990: If you feel curious and adventurous: why -1?

src/sparseml/pytorch/torchvision/train.py
dbogunowicz, line 328: it sounds to me that by passing rank = None you run in non-parallel (synchronous?) mode. it would be probably more readable to not use the context manager here if the synchronous mode is chosen. otherwise confusing.

dbogunowicz (commented):
Few nits, but close to approval.

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -323,9 +324,20 @@ def collate_fn(batch):
...
+                pretrained=args.pretrained,
+                pretrained_path=args.checkpoint_path,
+                pretrained_dataset=args.pretrained_dataset,
+                num_classes=num_classes,
+            )
+    else:
+        # fall back to torchvision
+        model = torchvision.models.__dict__[args.arch_key](
+            pretrained=args.pretrained, num_classes=num_classes
+        )
```
bfineran, line 340: We should have a check and error message that lists available NM model repo keys if torchvision fallback also fails

bfineran (approved):

PR #1138, opened 2022-11-09
Use correct URL for ONNX export

mathemusician:
When I click on the link, it leads to a 404. Now, it points to the right place


markurtz (approved):

mgoin (approved):

markurtz:
Thanks @mathemusician for fixing!

PR #1142, opened 2022-11-10
transformers onnx export - fix bug in model name override

bfineran:
key issue here was the previous code was attempting to add a string to a set with `update` instead of `add` which will lead to issues where each character of the onnx model string will be treated as expected files leading to an error.

additionally fixes issues around using an object as a default, user interface of requiring a set

**test_plan:**
bug reproduced locally, confirmed fix locally

### DIFF
--- a/src/sparseml/transformers/export.py
+++ b/src/sparseml/transformers/export.py
@@ -63,12 +63,13 @@ sparseml.transformers.export_onnx \
 
 import argparse
 import collections
+import copy
 import inspect
 import logging
 import math
 import os
 import shutil
-from typing import Any, Optional, Set
+from typing import Any, List, Optional
 
 from torch.nn import Module
 from transformers import AutoConfig, AutoTokenizer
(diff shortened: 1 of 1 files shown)


src/sparseml/transformers/export.py
```diff
@@ -270,21 +271,17 @@ def create_deployment_folder(
...
+        default file names with.
     :return: path to the valid deployment directory
     """
 
-    if onnx_file_name != MODEL_ONNX_NAME:
-        # replace the default onnx name with the custom one
-        deployment_files.remove(MODEL_ONNX_NAME)
-        deployment_files.update(onnx_file_name)
-
-    if training_directory.split("/")[-1] != "training":
```
bfineran, line 283: removing this warning - 1) should be using os or pathlib for working with paths since "/" for directories are os specific 2) default transformers flow right now does not have a directory called "training"

natuan (approved):
I verified this fixes the issue I reported

rahul-tuli (approved):

PR #1149, opened 2022-11-14
[WIP] Add convenience functions to recipe template

rahul-tuli:
The following PR adds 4 convenience functions to recipe_template module, the goal being that they can be used directly and have good defaults, the function could in turn be also used as an alternative to  passing a recipe to `ScheduleModifierManager.from_yaml(...)` and instead do something like `ScheduledModifierManager.from_yaml(create_recipe())`


src/sparseml/pytorch/recipe_template/utils.py
```diff
@@ -0,0 +1,50 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
```
bfineran, line 1: let's move this to `src/sparseml.recipe_template`

PR #1151, opened 2022-11-14
Don't overwrite vars in modifier with recipe_args

KSGulin:
Currently, modifier vars in recipes can be overwritten by recipe_args if they match the provided arg name. This is unintended behavior this PR removes.

Example affected case:

**recipe_args arg**
`{\"end_epoch\":10}`

**Recipe**
```
end_epoch: 5
training_modifiers:
  - !EpochRangeModifier
    start_epoch: 0
    end_epoch: eval(end_epoch)

pruning_modifiers:
  - !MagnitudePruningModifier
    params: __ALL_PRUNABLE__
    init_sparsity: 0.5
    final_sparsity: 0.9
    start_epoch: 0
    end_epoch: 3
```

Resulting recipe after processing (Currently):
```
end_epoch: 10
training_modifiers:
  - !EpochRangeModifier
    start_epoch: 0
    end_epoch: 10

pruning_modifiers:
  - !MagnitudePruningModifier
    params: __ALL_PRUNABLE__
    init_sparsity: 0.5
    final_sparsity: 0.9
    start_epoch: 0
    end_epoch: 10
```

Resulting recipe after processing (after PR):
```
end_epoch: 10
training_modifiers:
  - !EpochRangeModifier
    start_epoch: 0
    end_epoch: 10

pruning_modifiers:
  - !MagnitudePruningModifier
    params: __ALL_PRUNABLE__
    init_sparsity: 0.5
    final_sparsity: 0.9
    start_epoch: 0
    end_epoch: 3
```


bfineran (dismissed):

rahul-tuli (dismissed):

natuan:
@KSGulin Can we also add one unit test (e.g. using the one in the description) the ensure the modifier arg won't changed by the recipe arg of the same name?

eldarkurtic:
@KSGulin just to verify if I understand the diff and the expected pathway: after this PR, the `recipe_args` will be able to update only non-modifier variables, and the modifier variables with `eval(...)` will get be assigned the eval-value later during the ` _maybe_evaluate_recipe_equation` call?

KSGulin:
@natuan good call. I modified an existing test to cover this case

KSGulin:
@eldarkurtic yes exactly. The intention has always been to only allow modification of the global variables (although a sensible format for referencing specific modifiers with recipe-args might be worth developing in the future). So this PR brings us back to that state. As you said, the eval()'s are assigned later by `_maybe_evaluate_recipe_equation` and will reflect the values updated by the recipe-args

rahul-tuli (approved):

dbogunowicz (approved):

PR #1104, opened 2022-10-25
torchvision image classification

corey-nm:
This is the base feature branch for torchvision image classification rewrite

TODOS:
- [x] Add in managers/finalizers #1103 
- [x] Override lr scheduler with manager #1141 
- [x] Handle EMA #1113 
- [x] Add `--gradient-accum-steps` #1106 
- [x] ~Add --max-train-steps~ REVERTED #1108
- [x] ~Add --max-eval-steps~ REVERTED #1109
- [x] Add `--save-best-after` #1114
- [x] Recipe staging and model re-loading #1124
- [x] Override AMP after certain epoch #1107 
- [x] Use ModelRegistry and fall back to torchvision models if not found in model registry #1136 
- [x] export script (requires sample inputs & outputs) #1154
    - [ ] (Lower priority) add `--one-shot` arg

Verification
- [ ] ensure you can plug in torchvision models in deepsparse pipelines
- [x] verify reproducibility for current recipes
- [x] ensure export_onnx works on output of training (See #1154)

Later:
- [ ] add helper functions
- [ ] override weight decay with manager?
- [ ] Add `--log-dir` and add tensorboard support
- [ ] Swap to using click for cli

# Testing plan

### Base training

Ran the following and verified that they both achieve similar top1 accuracy scores of > 90%
```bash
sparseml.image_classification.train \
    --recipe-path zoo:cv/classification/resnet_v1-50/pytorch/sparseml/imagenette/pruned-conservative?recipe_type=original \
    --pretrained True \
    --arch-key resnet50 \
    --dataset imagenette \
    --dataset-path ~/.cache/nm_datasets/imagenette/imagenette-320/  \
    --train-batch-size 128 --test-batch-size 128 \
    --optim SGD \
    --optim-args '{}' \
    --model-tag resnet50-imagenette-pruned-conservative \
    --save-dir ./training-runs/image_classification-pretrained \
    --logs-dir ./training-runs/image_classification-pretrained
```

```bash
python src/sparseml/pytorch/torchvision/train.py \
    --recipe-path zoo:cv/classification/resnet_v1-50/pytorch/sparseml/imagenette/pruned-conservative?recipe_type=original \
    --pretrained True \
    --arch-key resnet50 \
    --dataset imagenette \
    --dataset-path ~/.cache/nm_datasets/imagenette/imagenette-320/  \
    --batch-size 128 \
    --opt sgd-nesterov \
    --momentum 0.9 \
[...]

PR #1156, opened 2022-11-16
Adding torchvision training scripts with sparseml additions

corey-nm:
This contains all the torchvision code from https://github.com/pytorch/vision/tree/v0.12.0/references/classification, in addition to the code merged in from #1104.

See #1104 for testing plan


corey-nm:
@bfineran do i need to do anything about pytorch vision licenses in the files from their repo?

PR #1178, opened 2022-11-22
Adding export script for openpifpaf

corey-nm:
The main differences from the builtin openpifpaf export script are:
1. Using our ModuleExporter
2. Adding dataset arguments to cli so the correct model structure is exported (builtin one assumes cocokp)

Test plan:
1. Ran export command with a checkpoint from training
2. Ran deepsparse.benchmark on the resulting .onnx file

### DIFF
--- a/src/sparseml/openpifpaf/export_onnx.py
+++ b/src/sparseml/openpifpaf/export_onnx.py
@@ -22,6 +22,7 @@ import onnx
 import torch
 
 import openpifpaf
+from sparseml.pytorch.optim.manager import ScheduledModifierManager
 from sparseml.pytorch.utils import ModuleExporter
 
 
@@ -66,7 +67,9 @@ def main():
     )
     parser.add_argument("--input-width", type=int, default=129)
     parser.add_argument("--input-height", type=int, default=97)
-
+    parser.add_argument(
+        "--one-shot", type=str, help="Path to recipe to apply in a zero shot fashion."
+    )
     openpifpaf.datasets.cli(parser)
 
     args = parser.parse_args()
@@ -76,6 +79,10 @@ def main():
     datamodule = openpifpaf.datasets.factory(args.dataset)
 
     model, _ = openpifpaf.network.Factory().factory(head_metas=datamodule.head_metas)
+
+    if args.one_shot:
+        manager = ScheduledModifierManager.from_yaml(args.one_shot)
+        manager.apply(model)
 
     image_size_warning(model.base_net.stride, args.input_width, args.input_height)
 
--- a/src/sparseml/openpifpaf/export_onnx.py
+++ b/src/sparseml/openpifpaf/export_onnx.py
@@ -17,6 +17,7 @@
 import argparse
 import logging
 import os
+from typing import Optional
 
 import onnx
 import torch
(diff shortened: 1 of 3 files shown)


src/sparseml/openpifpaf/export_onnx.py
```diff
@@ -0,0 +1,98 @@
...
+
+    exporter = ModuleExporter(model, args.save_dir)
+    exporter.export_onnx(
+        torch.randn(1, 3, args.input_height, args.input_width),
+        name=args.name,
+        input_names=["input_batch"],
+        output_names=[meta.name for meta in datamodule.head_metas],
+    )
+    onnx.checker.check_model(os.path.join(args.save_dir, args.name))
+    exporter.create_deployment_folder()
```
dbogunowicz, line 94: nice!

src/sparseml/openpifpaf/export_onnx.py
```diff
@@ -0,0 +1,98 @@
...
+    parser.add_argument("--input-width", type=int, default=129)
+    parser.add_argument("--input-height", type=int, default=97)
+
+    openpifpaf.datasets.cli(parser)
+
+    args = parser.parse_args()
+
+    openpifpaf.network.Factory.configure(args)
+
+    datamodule = openpifpaf.datasets.factory(args.dataset)
```
dbogunowicz, line 76: This fails for me locally.
args.dataset in current setting is `None`.
factory method of `openpifpaf.datasets` calls:
```python
def factory(dataset):
    if '-' in dataset:
        datamodules = [factory(ds) for ds in dataset.split('-')]
        return MultiDataModule(datamodules)

    if dataset not in DATAMODULES:
        raise Exception('dataset {} unknown'.format(dataset))
    return DATAMODULES[dataset]()
```
and fails if dataset is `None`.

This happens on `openpifpaf==0.13.6`

dbogunowicz: Wondering how to solve it. Either requiring the dataset name (one of those that are supported by openpifpaf) or setting if by default to CIFCAF heads.

corey-nm: Are you passing in `--dataset cocokp` on cli?

dbogunowicz: Fair point. Works with the manually specified argument, all good.

src/sparseml/openpifpaf/export_onnx.py
corey-nm: 🥂

src/sparseml/openpifpaf/export_onnx.py
```diff
@@ -77,6 +80,10 @@ def main():
 
     model, _ = openpifpaf.network.Factory().factory(head_metas=datamodule.head_metas)
 
+    if args.one_shot:
+        manager = ScheduledModifierManager.from_yaml(args.one_shot)
+        manager.apply(model)
```
dbogunowicz, line 85: nice and clean 8)

src/sparseml/openpifpaf/export_onnx.py
```diff
@@ -0,0 +1,105 @@
...
+    openpifpaf.datasets.cli(parser)
+
+    args = parser.parse_args()
+
+    openpifpaf.network.Factory.configure(args)
+
+    datamodule = openpifpaf.datasets.factory(args.dataset)
+
+    model, _ = openpifpaf.network.Factory().factory(head_metas=datamodule.head_metas)
+
```
bfineran, line 82: Let's separate main logic from arg parsing into a separate function

rahul-tuli: +1

bfineran (approved):

PR #1181, opened 2022-11-23
[bug-fix] enforce all converted quantized ONNX initializers are same dtype

bfineran:
torch 1.12 exports some Q/DQ inputs as initializers now instead of constants - this causes a pass in the quantized graph conversion step to miss a conversion from INT8 to UINT8. this PR adds an explicit pass to convert any missed INT8 to UINT8 initializers. without this fix, models may fail on ORT and deepsparse compile due to a mismatch between INT8 and UINT8 inputs

**test_plan:**
@KSGulin has reproduced the error to verify fix and commandeer
need to verify compilation + accuracy

### DIFF
--- a/src/sparseml/pytorch/sparsification/quantization/quantize_qat_export.py
+++ b/src/sparseml/pytorch/sparsification/quantization/quantize_qat_export.py
@@ -243,6 +243,25 @@ def _convert_single_constants_to_initializers(model: ModelProto):
     model.graph.node.extend(non_single_constant_nodes)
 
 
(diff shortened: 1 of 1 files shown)


KSGulin (approved):
Tested the scenarios:
- Torch==1.12.1 + `quantize_conv_activations: False` in quantization modifier
- Torch==1.12.1 + NO `quantize_conv_activations: False` in quantization modifier
- Torch==1.9.1 + `quantize_conv_activations: False` in quantization modifier

All passed. Great fix! 🥳

corey-nm (approved):
lgtm!

PR #1188, opened 2022-11-29
Adding checkpoint manager functionality to openpifpaf

corey-nm:
Adds ability to load recipes from checkpoints similar to torchvision

### DIFF
--- a/src/sparseml/openpifpaf/train.py
+++ b/src/sparseml/openpifpaf/train.py
@@ -19,6 +19,7 @@ import copy
 import logging
 import os
 import socket
+from typing import Optional, Tuple
 
 import torch
 
@@ -26,6 +27,7 @@ import openpifpaf
 from openpifpaf import __version__
 from openpifpaf.train import default_output_file
 from sparseml.openpifpaf.trainer import SparseMLTrainer
+from sparseml.pytorch.optim.manager import ScheduledModifierManager
 from sparseml.pytorch.utils.helpers import download_framework_model_by_recipe_type
 from sparsezoo import Model
 
@@ -44,6 +46,9 @@ def cli():
         "--version",
         action="version",
         version="OpenPifPaf {version}".format(version=__version__),
+    )
+    parser.add_argument(
+        "--recipe", default=None, required=True, help="Path to sparseml recipe"
     )
     parser.add_argument("-o", "--output", default=None, help="output file")
     parser.add_argument("--disable-cuda", action="store_true", help="disable CUDA")
@@ -163,11 +168,13 @@ def main():
     zoo_stub = None
     if args.checkpoint and args.checkpoint.startswith("zoo:"):
         checkpoint_path = download_framework_model_by_recipe_type(Model(zoo_stub))
-        checkpoint = torch.load(checkpoint_path, map_location="cpu")
-        args.basenet = checkpoint["meta"]["args"]["basenet"]
+        zoo_checkpoint = torch.load(checkpoint_path, map_location="cpu")
+        args.basenet = zoo_checkpoint["meta"]["args"]["basenet"]
         LOG.info(f"Overriding --basenet with value from checkpoint: {args.basenet}")
         zoo_stub = args.checkpoint
         args.checkpoint = None
+    else:
+        zoo_checkpoint = None
 
     # NOTE: configure this after removing zoo stub checkpoint
     openpifpaf.network.Factory.configure(args)
@@ -177,9 +184,9 @@ def main():
     )
     loss = openpifpaf.network.losses.Factory().factory(datamodule.head_metas)
 
-    if zoo_stub is not None:
-        # just load state dict from zoo stub
-        net_cpu.load_state_dict(checkpoint["state_dict"])
+    manager, checkpoint_manager = _load_managers_from_checkpoint(
+        args.recipe, net_cpu, zoo_checkpoint
+    )
 
     checkpoint_shell = None
     if not args.disable_cuda and torch.cuda.device_count() > 1 and not args.ddp:
@@ -234,13 +241,16 @@ def main():
     lr_scheduler = openpifpaf.optimize.factory_lrscheduler(
         args, optimizer, len(train_loader), last_epoch=start_epoch
     )
+
+    optimizer = manager.modify(net, optimizer, len(train_loader))
+
     trainer = SparseMLTrainer(
         net,
         loss,
         optimizer,
         args.output,
-        checkpoint_path=args.checkpoint or zoo_stub,
-        num_batches_per_epoch=len(train_loader),
+        manager=manager,
+        checkpoint_manager=checkpoint_manager,
         checkpoint_shell=checkpoint_shell,
         lr_scheduler=lr_scheduler,
         device=args.device,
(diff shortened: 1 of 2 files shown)


src/sparseml/openpifpaf/train.py
```diff
@@ -151,11 +152,37 @@ def main():
...
     net_cpu, start_epoch = openpifpaf.network.Factory().factory(
         head_metas=datamodule.head_metas
     )
     loss = openpifpaf.network.losses.Factory().factory(datamodule.head_metas)
 
+    if zoo_stub is not None:
+        # just load state dict from zoo stub
+        checkpoint_path = download_framework_model_by_recipe_type(Model(zoo_stub))
+        checkpoint = torch.load(checkpoint_path, map_location="cpu")
+        net_cpu.load_state_dict(checkpoint["state_dict"])
```
dbogunowicz, line 184: Curious, I remember that in image classification checkpoints we save the name of the model architecture inside the checkpoint dict. Are we doing the same thing here? If yes, we could raise an error if the `args.basenet` deviates from the network loaded from the checkpoint. I imagine this could be quite good for the overall UX.

src/sparseml/openpifpaf/trainer.py
```diff
@@ -75,6 +82,43 @@ def __init__(
             model_meta_data=model_meta_data,
         )
 
+    def _load_checkpoint(self, checkpoint_path, model):
```
dbogunowicz, line 85: nit: typing for arguments and outputs

src/sparseml/openpifpaf/trainer.py
```diff
@@ -116,16 +160,25 @@ def write_model(self, epoch, final=True):
...
-            {
-                "model": model_to_save,
-                "state_dict": model_to_save.state_dict(),
-                "checkpoint_recipe": str(self.manager),
-                "epoch": epoch,
-                "meta": self.model_meta_data,
-            },
-            filename,
-        )
+
```
dbogunowicz, line 163: 🔥

dbogunowicz (approved):

src/sparseml/openpifpaf/train.py
corey-nm: Oh yeah great idea, I'll make that change

src/sparseml/openpifpaf/train.py
```diff
@@ -151,11 +152,35 @@ def main():
 
     datamodule = openpifpaf.datasets.factory(args.dataset)
 
+    """
+    Cases for checkpointing:
+    1. No checkpoint - no changes to base flow
+    2. Checkpoint from local file - no changes to base flow
+    3. Checkpoint from zoo stub:
+        a. set args.checkpoint to None to get a base net
```
bfineran, line 160: how is the base net specified? does the user need to include that as CLI args? If so, let's look into including the specification in the state dict to remove that potential point of user failure

src/sparseml/openpifpaf/trainer.py
```diff
@@ -75,6 +82,45 @@ def __init__(
...
+        if checkpoint["epoch"] == -1:
+            # restore state from finished recipe
+            LOG.info(
+                "Checkpoint was from epoch -1, "
+                "checkpoint recipe is NOT overriding configured recipe"
+            )
+            checkpoint_manager.apply_structure(model, epoch=checkpoint["epoch"])
+            self.checkpoint_manager = checkpoint_manager
+        else:
+            # resume
```
bfineran, line 115: why are we choosing to override the given training recipe which was specified by the user? should at least log a warning in this case, however not convinced that this should be the default/only behavior.
What is the current "resume" rules for openpifpaf?

src/sparseml/openpifpaf/trainer.py
```diff
@@ -75,6 +82,45 @@ def __init__(
...
+        checkpoint_manager = ScheduledModifierManager.from_yaml(
+            checkpoint["checkpoint_recipe"]
+        )
+        if checkpoint["epoch"] == -1:
+            # restore state from finished recipe
+            LOG.info(
+                "Checkpoint was from epoch -1, "
+                "checkpoint recipe is NOT overriding configured recipe"
+            )
+            checkpoint_manager.apply_structure(model, epoch=checkpoint["epoch"])
```
bfineran, line 112: we should load weights after applying the structural changes

src/sparseml/openpifpaf/train.py
```diff
@@ -151,11 +152,35 @@ def main():
...
+    openpifpaf.network.Factory.configure(args)
+
     net_cpu, start_epoch = openpifpaf.network.Factory().factory(
         head_metas=datamodule.head_metas
     )
     loss = openpifpaf.network.losses.Factory().factory(datamodule.head_metas)
 
+    if zoo_stub is not None:
+        # just load state dict from zoo stub
+        net_cpu.load_state_dict(checkpoint["state_dict"])
```
bfineran, line 182: calling into `load_state_dict` immediately after model initialization isn't safe since if the model was structurally changed in a previous run, the module state will not state dict keys. we should connect this with the logic in `_load_managers_from_checkpoint` so the flow includes this

bfineran (changes requested):

src/sparseml/openpifpaf/trainer.py
corey-nm: That's what we did in torchvision too - was just copying that. Is that not what we want?

src/sparseml/openpifpaf/train.py
corey-nm: --basenet from cli. it is also saved into the checkpoint under checkpoint["meta"]["args"]["basenet"]. For loading from a checkpoint we read basenet from the checkpoint. For normal training users have to specify on CLI

src/sparseml/openpifpaf/train.py
```diff
@@ -151,11 +157,37 @@ def main():
...
+        b. manually load in state dict in _load_managers_from_checkpoint
+    """
+    if args.checkpoint:
+        if args.checkpoint.startswith("zoo:"):
+            args.checkpoint = download_framework_model_by_recipe_type(
+                Model(args.checkpoint)
+            )
+        checkpoint = torch.load(args.checkpoint, map_location="cpu")
+        # NOTE: set basenet instead of checkpoint so we get a randomized network
+        args.basenet = checkpoint["meta"]["args"]["basenet"]
```
bfineran, line 174: Is this always set by openpifpaf?

src/sparseml/openpifpaf/train.py
```diff
@@ -228,5 +264,46 @@ def main():
     trainer.loop(train_loader, val_loader, start_epoch=start_epoch)
 
 
+def _load_managers_from_checkpoint(
```
bfineran, line 267: Let's update function name since it also includes weight loading

bfineran (approved):
LGTM pending comments

src/sparseml/openpifpaf/train.py
corey-nm: It's from the Network cli. the basenet and checkpoint arguments are mutually exclusive, so if checkpoint is set there are checks that basenet is not set and vice versa.

The network factory code is essentially:
```
if args.basenet:
    return net, epoch
else:
    load_from_checkpoint
```

So the code here is basically making sure we never use their checkpoint loading logic, since we always want to be applying structure and loading weights ourselves

src/sparseml/openpifpaf/train.py
corey-nm: Updated

PR #1190, opened 2022-11-30
Fixing bugs with resume & checkpointing in torchvision

corey-nm:
These two lines fix two separate issues:
1. A value Error that resume and checkpoint-path cannot both be specified is thrown if checkpoint-path is specified, because the default of resume is empty string
2. The model params are in "state_dict" instead of "model"

### DIFF
--- a/src/sparseml/pytorch/torchvision/train.py
+++ b/src/sparseml/pytorch/torchvision/train.py
@@ -440,7 +440,7 @@ def main(args):
 
     # load params
     if checkpoint is not None:
-        model.load_state_dict(checkpoint["model"])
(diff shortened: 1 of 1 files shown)


rahul-tuli (approved):

dbogunowicz (approved):

bfineran (approved):
LGTM, pending we verify click does not choke on setting a str to None

corey-nm:
> LGTM, pending we verify click does not choke on setting a str to None

Yeah it doesn't - ran training without specifying resume/checkpoint-path and specifying both and it all worked!

PR #1192, opened 2022-12-01
Generic model export pipeline + torch to ONNX and ONNX to deepsparse refactor

dbogunowicz:
# Summary

This PR reworks all of the onnx optimization passes contained in [src/sparseml/pytorch/sparsification/quantization/quantize_qat_export.py](https://github.com/neuralmagic/sparseml/blob/v1.2.0/src/sparseml/pytorch/sparsification/quantization/quantize_qat_export.py).

The new classes/functions are introduced:
1. Transforms: `class BaseTransform` and `class OnnxTransform(BaseTransform)`, which represent abstract transforms that act on onnx graphs
2. Exporters: `class BaseExporter(BaseTransform)`, `class TorchToONNX(BaseExporter)`, and `class ONNXToDeepsparse(BaseExporter)`, which represent transforms that can also save things to disk

In addition, every transform from quantize_qat_export (and other files), has been added as a `OnnxTransform`. This includes:
1. The actual transform implementation
2. docstring of exactly what the transform does
3. a unit test for the transform

The `get_structural_matches` utility function was also added to aid in the implementation of these transforms. Notably it allows callees to match sub-graphs in any onnx graph, which makes implementation details more clear/explicit.

# Test Plan

As mentioned in the summary, every transform has a unit test associated with it. Additionally, regression tests were added against the old module exporter logic for the following models:
1. Resnet50
2. Yolov5
3. Bert for QA
So both the graph structure and numerical output match *exactly* for both the new exporters and the old exporters.


bfineran (approved):
🚀

KSGulin (approved):
Super clean implementation!

dbogunowicz:
@corey-nm great job! I like the PR message.
Can we override the tests and land it now?

corey-nm:
I think @bfineran would have to force merge it

PR #1194, opened 2022-12-01
[Exporter Refactor] Fold Identity Initializers Transform

dbogunowicz:
Adds:
- `fold_identity_initializers` transform


src/sparseml/exporters/transforms/base_transform.py
```diff
@@ -0,0 +1,58 @@
...
+    A transform is an object that, when applied to a model,
+    does a single modification to it
+    and returns a modified model.
+    """
+
+    def __init__(self):
+        pass
+
+    @abstractmethod
+    def __call__(self, model: Any) -> Any:
```
dbogunowicz, line 31: It felt appropriate to apply the transform through the `__call__` method, feel free to challenge me.

src/sparseml/exporters/transforms/base_transform.py
```diff
@@ -0,0 +1,58 @@
...
+        """
+        1. Validate the input model
+        2. Apply the transform to the model
+        3. Validate the resulting model and return it
+
+        :param model: The input model to be transformed
+        :return: The transformed model
+        """
+        self._validate_input(model)
```
dbogunowicz, line 40: Maybe we can add a private function that maps from any input (model object, path string) to the actual model object e.g.`_read_the_model`? This way we might specify all the functionalities of the `__call__` method in the base class and would rarely have a need to modify it in the children classes.

corey-nm: I like this idea but let's put off refactors like it until we actually have a good set of transforms. Don't want to make assumptions like this without actually having many of the transforms built

src/sparseml/exporters/transforms/base_transform.py
```diff
@@ -0,0 +1,58 @@
...
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from abc import ABC, abstractmethod
+from typing import Any
+
+
+class BaseTransform(ABC):
```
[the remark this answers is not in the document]
corey-nm: Yep totally agree - that is what the OnnxTransform in the PRD is supposed to capture 💪

src/sparseml/exporters/transforms/fold_identity_initializers.py
```diff
@@ -0,0 +1,74 @@
...
+
+from typing import Union
+
+from onnx import ModelProto, NodeProto
+
+from sparseml.exporters.transforms import BaseTransform
+from sparseml.onnx.utils import ONNXGraph, check_load_model, validate_onnx_file
+
+
+def is_identity(node: NodeProto, graph: "ONNXGraph") -> bool:
```
corey-nm, line 23: One of my goals for OnnxTransform is to have it specified like:

```python
class OnnxTransform(BaseTransform):
    def iter_matches(self, graph: ONNXGraph) -> Iterable[<something>]:
        ...
    def transform_match(self, graph: ONNXGraph, match: <something>):
        ...
    def _transform(self, model):
        for match in self.iter_matches(graph):
             self.transform_match(graph, match)
```

And then in that way we could have FoldIdentityInitializers look like:

```python
class FoldIdentityInitializers(ONNXTransform):
     def iter_matches(self, ...):
          <pull is_identity into this function>
     def transform_match(self, ...):
         < pull inner loop into this function>
```

corey-nm (approved):
lgtm. i expect the base classes may go through some more iterations as we get the more complex transforms added in

tests/sparseml/exporters/transforms/test_fold_identity_initializers.py
```diff
@@ -0,0 +1,55 @@
...
+
+def test_fold_identity_initializers():
+    model_input = onnx.helper.make_tensor_value_info(
+        "input", onnx.TensorProto.FLOAT, (1,)
+    )
+    model_output = onnx.helper.make_tensor_value_info(
+        "output", onnx.TensorProto.FLOAT, (1,)
+    )
+
+    init1 = onnx.helper.make_tensor(
```
bfineran, line 28: let's maybe move graph construction out into a helper function with a comment on the graph structure

src/sparseml/exporters/transforms/base_transform.py
bfineran: prefer to have an explicitly named method like `apply()` as well, can have `__call__` just got into `apply`

akin to `__call__` and `forward()` in pytorch

src/sparseml/exporters/transforms/fold_identity_initializers.py
dbogunowicz: Let's sync on this on Monday, this looks like a good idea.

src/sparseml/exporters/transforms/base_transform.py
```diff
@@ -0,0 +1,58 @@
...
+        # The transform algorithm that will be applied to the model
+        raise NotImplementedError
+
+    @abstractmethod
+    def _validate_input(self, model: Any):
+        # Validate the input model before applying the transform
+        raise NotImplementedError
+
+    @abstractmethod
+    def _validate_output(self, model: Any):
```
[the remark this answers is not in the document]
dbogunowicz: it's just that they are not meant to be called by the user, by rather as a part of the logic in `__call__`/`apply` method.
But having a docstring sounds good to me.

src/sparseml/exporters/transforms/fold_identity_initializers.py
```diff
@@ -0,0 +1,74 @@
...
+        return model
+
+    def _validate_input(self, model: ModelProto):
+        validate_onnx_file(model)
+
+    def _validate_output(self, model: ModelProto):
+        validate_onnx_file(model)
+
+    def apply(self, model: Union[ModelProto, str]) -> ModelProto:
+        onnx_model = check_load_model(model)
```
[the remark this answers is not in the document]
dbogunowicz: Yeah, let us figure it out today/tomorrow, when we get `ONNXTransform` in place.

src/sparseml/exporters/transforms/fold_identity_initializers.py
```diff
@@ -0,0 +1,69 @@
...
+class FoldIdentityInitializers(OnnxTransform):
+    """
+    Folds any `Identity` initializer node into the graph.
+    TODO: Add graph
+    """
+
+    def transform(self, model: ModelProto) -> ModelProto:
+        count_converted_nodes = 0
+        graph = ONNXGraph(model)
+        for match in get_structural_matches(
```
dbogunowicz, line 59: I am wondering whether this match check is strong enough. Originally (in the pre-refactor code), this was the set of criteria needed for the check to pass:

```python
  def is_match(node: onnx.NodeProto) -> bool:
        return (
            node.op_type == "Identity"
            and len(node.input) == 1
            and len(node.output) == 1
            and node.input[0] in graph._name_to_initializer
        ]
```
But from my quick study of identity ONNX OP (http://www.xavierdupre.fr/app/mlprodict/helpsphinx/onnxops/onnx__Identity.html) it seems that all those need to be also fulfilled to create a valid identity op in the onnx framework, so we may not need to explicitly encode them here

corey-nm: 🔥 nice!!

src/sparseml/exporters/transforms/fold_identity_initializers.py
```diff
@@ -0,0 +1,82 @@
...
+    |      OUTPUT
+    """
+
+    def transform(self, model: ModelProto) -> ModelProto:
+        count_converted_nodes = 0
+        graph = ONNXGraph(model)
+        for match in get_structural_matches(
+            graph,
+            op_type="Identity",
+        ):
```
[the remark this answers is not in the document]
dbogunowicz: I think that our linter will force this style due to the length of the line

src/sparseml/exporters/transforms/fold_identity_initializers.py
corey-nm: Ehh I guess not, it's pretty clear from the onnx docs that this is the case https://github.com/onnx/onnx/blob/main/docs/Operators.md#Identity

src/sparseml/exporters/transforms/__init__.py
```diff
@@ -11,6 +11,14 @@
...
+Required to avoid running into
+circular dependencies when importing:
+
+isort:skip_file
+"""
+
 # flake8: noqa
 from .base_transform import *
 from .onnx_transform import *
+from .fold_identity_initializers import *
```
bfineran, line 24: tiny nit, can add later, but let's add a blank line to separate the base modules (base, onnx) vs the implementations

src/sparseml/exporters/transforms/fold_identity_initializers.py
```diff
@@ -0,0 +1,82 @@
...
+    :param match: Match node to be folded into the graph
+    :param model: ONNX model to be transformed
+    :return: ONNX model with the match node folded into the graph
+    """
+    graph = ONNXGraph(model)
+    for child_node in graph.get_node_children(match.node):
+        for i, child_node_input in enumerate(child_node.input):
+            if child_node_input == match.node.output[0]:
+                child_node.input[i] = match.node.input[0]
+    model.graph.node.remove(match.node)
```
bfineran, line 46: fyi - can save on time by just using this function over the matched nodes (outside of this function) https://github.com/neuralmagic/sparseml/blob/6f8bc9336f5f9b8370dced9e7017045236bd531f/src/sparseml/onnx/utils/graph_editor.py#L193
instead of calling `model.graph.node.remove` for each node individually.

this isn't so important for this pass since it's O(n^2) anyways but keep in mind for future as this has actually led to significant speedups in the past

tests/sparseml/exporters/transforms/test_fold_identity_initializer.py
```diff
@@ -0,0 +1,71 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import onnx
+
+from sparseml.exporters.transforms import FoldIdentityInitializers
+
```
bfineran, line 18: lgtm - nit for future is to make checks more generic so tests can be parameterized for future updates

bfineran (approved):
LGTM - see comments for food for thought

src/sparseml/exporters/transforms/fold_identity_initializers.py
dbogunowicz: This is an excellent comment.

PR #1195, opened 2022-12-01
Falling back to model.pth if model.ckpt.pth is missing

corey-nm:
If you run the transfer learning command from here https://sparsezoo.neuralmagic.com/models/cv%2Fclassification%2Fresnet_v1-50%2Fpytorch%2Fsparseml%2Fimagenet%2Fpruned95-none, the model.ckpt.pth file is missing, so you will get a value error raised even though the model.pth file exists.

This PR just falls back to model.pth if model.ckpt.pth is missing.

Testing plan:
1. ran training command above to verify crash
2. ran same training command after fix to verify training started


KSGulin (approved):

bfineran (approved):
nice

Chibukach (approved):
LGTM

PR #1198, opened 2022-12-01
[QuantizationModifier] improved quantization flow - control of propagation with schemes and stronger testing

bfineran:
this PR adds the following functionality and improvements:

core:
  * quantization modifier now fully controls propagation of schemes and qconfigs by having its own function for selecting which submodules get a quantization_scheme, then only adding qconfigs to submodules with schemes
  * modifier now by default targets all (minus a hardcoded list of 'non' quantizable modules) leaf submodules for quantization

improvements:
  * testing fixes
  * testing includes check that observers are actually created
  * quantization flow in modifier improved to just set schemes -> apply qat from schemes

**test_plan:**
unit tests included

### DIFF
--- a/src/sparseml/pytorch/sparsification/quantization/__init__.py
+++ b/src/sparseml/pytorch/sparsification/quantization/__init__.py
@@ -14,6 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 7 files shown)


KSGulin (approved):
LGTM!

PR #1201, opened 2022-12-02
[QuantizationModifier] submodule_schemes property impl - target specific submodules by scheme

bfineran:
`submodule_schemes` replaces the `submodules` functionality of the existing quantization modifier while enabling the user to specify different schemes for different submodules. through this functionality, users may build configs that control exactly what to quantize through the entire model

if `submodule_schemes` is specified, then the default behavior of quantizing "everything" will be scoped down to just the targeted submodules

the `_QuantizationSchemesDict` was necessary to add for yaml serialization - open to suggestions for cleaner ways of getting the property properly serialized

**test_plan:**
unit tests updated for:
* yaml serialization
* testing that only submodule schemes are quantized if specified
* cases added for each


bfineran:
note: a few small nits will be fixed in follow up PR adding support for `module_type_schemes`

bfineran:
NOTE - the UX here changes as of #1239

KSGulin (approved):
LGTM!

PR #1202, opened 2022-12-02
[QuantizationModifier] module_type_schemes - override quantization scheme by layer type

bfineran:
`module_type_schemes` allows a user to override quantization schemes by matching layer class names (ie `Linear`, `Conv2d`)

this replaces arguments in the existing quantization modifier that handle specific layer types (ie conv/linear activations, embeddings, etc)

**test_plan:**:
unit tests updated for:
* yaml serialization
* check the correct scheme is set when there is a module_type_schemes match


bfineran:
NOTE - the UX here changes as of #1239

KSGulin (approved):

PR #1207, opened 2022-12-05
Adding match_structure and iter_structural_matches

corey-nm:


src/sparseml/exporters/transforms/utils/add_quantized_conv_matmul_add_ops.py
```diff
@@ -19,7 +19,7 @@
 import torch
 from onnx import AttributeProto, ModelProto, NodeProto, TensorProto, numpy_helper
 
-from sparseml.exporters.transforms.utils import QuantizationParams
+from .helpers import QuantizationParams
```
dbogunowicz, line 22: By convention, we are refraining from using relative imports in our codebase.

src/sparseml/exporters/transforms/utils/helpers.py
```diff
@@ -161,3 +161,115 @@ def assert_node_type(node: NodeProto, op: Union[List[str], Set[str], str]) -> bo
         return node.op_type == op
     else:
         return node.op_type in op
+
+
+NodeOrInit = Union[NodeProto, TensorProto]
```
dbogunowicz, line 166: maybe a short comment here? it is not common knowledge that `TensorProto` in the context of exporters will usually describe an initializer

src/sparseml/exporters/transforms/utils/helpers.py
```diff
@@ -161,3 +161,115 @@ def assert_node_type(node: NodeProto, op: Union[List[str], Set[str], str]) -> bo
...
+
+
+def _is_optional_node(tag: str) -> Tuple[str, bool]:
+    if tag.startswith("Optional-"):
+        return tag.split("-")[1], True
+    else:
+        return tag, False
+
+
+def iter_structural_matches(
```
dbogunowicz, line 190: I'd add docstrings to the public functions in this script, they would be potentially:
- hard for us to decipher if we return to working on them in the future
- hard for somebody to expand on our work

I am very grateful to Tuan for leaving bits and pieces of comments in his code, that we are now refactoring. Let's employ the same mindset.

src/sparseml/exporters/transforms/utils/helpers.py
```diff
@@ -161,3 +161,115 @@ def assert_node_type(node: NodeProto, op: Union[List[str], Set[str], str]) -> bo
...
         return node.op_type in op
+
+
+NodeOrInit = Union[NodeProto, TensorProto]
+
+
+class MatchResult:
+    def __init__(self, node: NodeOrInit) -> None:
+        self.node: Optional[NodeOrInit] = node
+        self.parents: List[Optional[NodeOrInit]] = []
```
dbogunowicz, line 172: not List of Lists?

tests/sparseml/exporters/transforms/test_convert_quantizable_conv_integer.py
```diff
@@ -105,6 +105,6 @@ def test():
         "quantize_linear_node_0",
         "conv_node_quant",
         "conv_node_bias_add_quant",
-        "conv_nose_bias_add_quant_cast",
+        "conv_node_bias_add_quant_cast",
```
dbogunowicz, line 108: 🐽

src/sparseml/exporters/transforms/convert_quantizable_conv_integer.py
```diff
@@ -180,31 +125,32 @@ class ConvertQuantizableConvInteger(BaseTransform):
...
-            model = convert_conv_to_quantized(
-                model,
-                conv_node,
-                weight_quantize_node,
-                weight_dequantize_node,
-                input_quantize_node,
+        graph = ONNXGraph(model)
+        for match in iter_structural_matches(
+            graph,
+            parent_ops=[
```
dbogunowicz, line 131: Ok, this is seriously awesome. I like how this one function exhaustively describes the condition for the match.

src/sparseml/exporters/transforms/convert_quantizable_conv_integer.py
```diff
@@ -113,10 +54,13 @@ def convert_conv_to_quantized(
     :param input_quantize_node: The quantize node for the input of the conv node
     :return: The converted ONNX model
     """
-    graph = ONNXGraph(model)
+    input_quantize_node, input_dequantize_node = match.parents[0]
```
dbogunowicz, line 57: I am just worried whether the numbering of the nodes is commutative.
I.e what such a situation be possible
A = match.parents[0]
B = match.parents[1]

as well  as

A = match.parents[1]
B = match.parents[0]

src/sparseml/exporters/transforms/convert_quantizable_conv_integer.py
```diff
@@ -125,22 +69,23 @@ def convert_conv_to_quantized(
...
-        target_output=conv_node.output[0],
+        bias_initializer=bias_init,
+        bias_add_name="{}_bias_add".format(match.node.name),
+        target_output=match.node.output[0],
         transpose_weight=False,
     )
 
     delete_quant_node(model, weight_dequantize_node)
     delete_quant_node(model, weight_quantize_node)
+    delete_quant_node(model, input_dequantize_node)
```
dbogunowicz, line 80: Watch out, this duplicates line 85

src/sparseml/exporters/transforms/convert_quantizable_conv_integer.py
dbogunowicz: But yeah, having a list of list called `parent_ops` needs a good documentation and comments :)

src/sparseml/exporters/transforms/utils/helpers.py
```diff
@@ -161,3 +161,115 @@ def assert_node_type(node: NodeProto, op: Union[List[str], Set[str], str]) -> bo
...
+            node,
+            op_type=op_type,
+            parent_ops=parent_ops,
+            children_ops=children_ops,
+        )
+        if match is not None:
+            yield match
+
+
+def match_structure(
```
dbogunowicz, line 208: This is an awesome function, and my love of it is unjustifiably diminished by its complexity. Complexity is an easy thing to get rid of tho.

I know you probably now aimed for making the functionality work and will work on the cleanup now. Anyway, I'd like to propose:
- moving all the matching functions (incl. `iter_structural_matches` etc) to a separate script `.../exporters/transforms/utils/matching.py`
- breaking up `match_structure` into three subfunctions: `assert_node_validity` (L218-233), `assert_parent_ops` (L233-254) and `assert_children_ops (L254-275)
- adding comments

dbogunowicz (commented):
Kudos on the analysis of the larger pattern for matching and converting. This going to make our lives so much easier! 🚀

src/sparseml/exporters/transforms/convert_quantizable_conv_integer.py
corey-nm: the parents that are specified in the match definition are matched in the same order as the `node.input` values in the NodeProto. since we know this is a conv node, we know what each parent is.

i agree this is incredibly subtle. any idea how we could make it less subtle/more clear?

src/sparseml/exporters/transforms/convert_quantizable_conv_integer.py
corey-nm: That is deleting the quantize node it looks like? whereas this is the dequant?

I've seen some quantize nodes have multiple children, which would mean we can't delete it without breaking the graph

corey-nm: Perhaps the delete_quant_node should check that there are no more usages of the node before deleting?

src/sparseml/exporters/transforms/convert_quantizable_conv_integer.py
corey-nm: Agree, i find it way more clear! And yeah, any suggestions on naming of anything? A good name will probably save us a lot of headache. Will add docstring now

src/sparseml/exporters/transforms/convert_quantizable_conv_integer.py
dbogunowicz: this is subtle, but eventually, this is good news.
i think good docstrings will suffice, I would make them overly verbose if necessary

src/sparseml/exporters/transforms/utils/helpers.py
```diff
@@ -161,3 +161,116 @@ def assert_node_type(node: NodeProto, op: Union[List[str], Set[str], str]) -> bo
...
+            match.node = None
+        else:
+            return None
+
+    if parent_ops:
+        if not (isinstance(node, NodeProto) and len(parent_ops) <= len(node.input)):
+            return None
+
+        parents = graph.get_node_parents(node)
+        assert len(parents) == len(node.input)
```
dbogunowicz, line 231: raise error?

src/sparseml/exporters/transforms/utils/helpers.py
corey-nm: Good call out!

src/sparseml/exporters/transforms/utils/helpers.py
corey-nm: Separated it out as you suggested and put it into matching.py

PR #1210, opened 2022-12-07
[Exporter Refactor] `BaseTransform` and `OnnxTransform`

dbogunowicz:
This PR adds two interfaces: `BaseTransform` and `OnnxTransform` (that inherits from `BaseTransform`)


src/sparseml/exporters/transforms/base_transform.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @abstractmethod
+    def transform(self, model: Any) -> Any:
+        """
+        Logic for applying the transformation to the model.
+
+        :param model: The input model to be transformed
+        :return: The transformed model
+        """
+        # The transform algorithm that will be applied to the model
```
dbogunowicz, line 53:
```suggestion
        # The transform algorithm that will be applied to the model
```

src/sparseml/exporters/transforms/base_transform.py
```diff
@@ -0,0 +1,71 @@
...
+        should follow the following steps:
+        1. Validate the input model
+        2. Apply the transform to the model
+        3. Validate the resulting model and return it
+
+        :param model: The input model to be
+            validated, transformed and once again, validated
+        :return: The transformed model
+        """
+        raise NotImplementedError
```
corey-nm, line 43: optional nit: default implementation
```suggestion
        self.pre_validate(model)
        self.transform(model)
        self.post_validate(model)
```

src/sparseml/exporters/transforms/base_transform.py
```diff
@@ -0,0 +1,71 @@
...
+    Interface that all transforms must implement.
+    A transform is an object that, when applied to a model,
+    does a single modification to it
+    and returns a modified model.
+    """
+
+    def __call__(self, model: Any) -> Any:
+        return self.apply(model)
+
+    @abstractmethod
```
corey-nm, line 30: if default implementation comment below, then remove this

src/sparseml/exporters/transforms/onnx_transform.py
```diff
@@ -0,0 +1,80 @@
...
+            It can be a path to the model or the model itself.
+        :return: The transformed ONNX model
+        """
+        if not ((isinstance(model, str) or isinstance(model, ModelProto))):
+            raise ValueError(
+                f"Invalid model type: {type(model)}. "
+                "Must be a string (path to the .onnx file) or ONNX ModelProto"
+            )
+
+        model = check_load_model(model)
```
corey-nm, line 50: move these to pre_validate maybe?

Also maybe make pre_validate/post_validate return models as well?

If both of these then i think we don't need to override the default implementation of apply above

corey-nm (approved):
looks good to me! left some small suggestions, but this is great 🔥

src/sparseml/exporters/transforms/base_transform.py
bfineran: +1

bfineran: additionally, what's the expected flow here? do the validation functions raise errors as they encounter them, or do we return an error string that the apply step will raise?

bfineran (approved):
LGTM pending comment

PR #1200, opened 2022-12-02
[Exporter Refactor] Dev Branch

dbogunowicz:
Adds:
- `convert_quantizable_matmul` transform

PR #1211, opened 2022-12-07
[Exporter Refactor] Onnx sub graph matching

corey-nm:
This adds a function (`get_structural_matches`) that allows us to find all sub-graphs in an onnx graph that match a specific structure. Much of the code in the existing passes exists just to match sub-graphs, and it all follows a similar pattern. This function allows us to compactly and explicitly specify what we want to match against. It also retrieves the actual nodes for us and makes using them simple.

See docstrings for more information.

# Testing plan

unit tests & we converted two complex passes to use this function in this PR (https://github.com/neuralmagic/sparseml/pull/1200), which is tested by unit tests


src/sparseml/exporters/transforms/utils/matching.py
```diff
@@ -0,0 +1,329 @@
...
+2. an initializer (TensorProto)
+"""
+
+
+class MatchResult:
+    """
+    The output of :func:`get_structural_matches`
+    """
+
+    def __init__(self, node: NodeOrInit) -> None:
```
dbogunowicz, line 157: not necessary according to pep
```suggestion
    def __init__(self, node: NodeOrInit):
```

src/sparseml/exporters/transforms/utils/matching.py
```diff
@@ -0,0 +1,329 @@
...
+        op_type="Identity",
+        parent_ops=[[INITIALIZER_MATCH]]
+    )
+    for match in matches:
+        id_node = match.node
+        (init, ) = match.parents[0]
+        assert isinstance(init, onnx.TensorProto)
+    ```
+
+    Here is a really complicated example with optional nodes and multiple parents. Here
```
dbogunowicz, line 97: Can we also add the example with `parent_ops = [[], ["a", "b", ...]]`?

corey-nm: Good idea yep

dbogunowicz (approved):
Sweet

src/sparseml/exporters/transforms/utils/matching.py
```diff
@@ -0,0 +1,329 @@
...
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import List, Optional, Tuple, Union
+
+from onnx import NodeProto, TensorProto
+
+from sparseml.onnx.utils import ONNXGraph
```
dbogunowicz, line 19: nit: 
`__all__` missing

corey-nm: Added

src/sparseml/exporters/transforms/utils/matching.py
```diff
@@ -0,0 +1,356 @@
...
+    node: Union[NodeProto, TensorProto],
+    parent_ops: List[List[str]],
+) -> bool:
+    if not (isinstance(node, NodeProto) and len(parent_ops) <= len(node.input)):
+        return False
+
+    parents = graph.get_node_parents(node)
+    assert len(parents) == len(node.input)
+    for parent, expected_op_sequence in zip(parents, parent_ops):
+        if expected_op_sequence == []:
```
bfineran, line 294: let's add a comment for what this case represents

tests/sparseml/exporters/transforms/utils/test_matching.py
```diff
@@ -0,0 +1,136 @@
...
+import onnx
+import pytest
+
+from sparseml.exporters.transforms.utils.matching import (
+    INITIALIZER_MATCH,
+    get_structural_matches,
+    optional_node,
+)
+from sparseml.onnx.utils.graph_editor import ONNXGraph
+
```
bfineran, line 24: definitely need to add some comments around the fixture and maybe a little in the functions

tests/sparseml/exporters/transforms/utils/test_matching.py
```diff
@@ -0,0 +1,136 @@
...
+
```
bfineran, line 25: for a test of this scope, probably good to pull down a zoo model so we can do more intricate testing. looks like we don't test too deeply into the recursive call

tests/sparseml/exporters/transforms/utils/test_matching.py
```diff
@@ -0,0 +1,136 @@
...
+    assert len(matches[0].children[0]) == 1
+    assert matches[0].children[0][0].name == "id2"
+
+
+def test_only_op_type(onnx_graph: ONNXGraph):
+    matches = get_structural_matches(onnx_graph, op_type="Conv")
+    assert len(matches) == 0
+
+    matches = get_structural_matches(onnx_graph, op_type="Identity")
+    assert len(matches) == 2
```
bfineran, line 106: any reason not to assert an `==` against the actual expected lists or even expected match objects?
that way we could pull everything into a parameterized test function

src/sparseml/exporters/transforms/utils/matching.py
```diff
@@ -0,0 +1,356 @@
...
+
+
+def _is_optional_node(tag: str) -> Tuple[str, bool]:
+    if tag.startswith("Optional-"):
+        return tag.split("-")[1], True
+    else:
+        return tag, False
+
+
+def _match_structure(
```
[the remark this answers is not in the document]
corey-nm: Mainly just easier to make private things public than other way around. If its useful to expose then we can pretty easily

src/sparseml/exporters/transforms/utils/matching.py
```diff
@@ -0,0 +1,356 @@
...
+    graph: ONNXGraph,
+    node: Union[NodeProto, TensorProto],
+    parent_ops: List[List[str]],
+) -> bool:
+    if not (isinstance(node, NodeProto) and len(parent_ops) <= len(node.input)):
+        return False
+
+    parents = graph.get_node_parents(node)
+    assert len(parents) == len(node.input)
+    for parent, expected_op_sequence in zip(parents, parent_ops):
```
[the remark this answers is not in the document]
corey-nm: Its actually necessary to not match exactly against length. e.g. "QuantizeLinear" has 3 inputs but with our recursion we only ever match against 1 input.

So the only checks right now are just to ensure that then length is at least the length of the actual parents.

tests/sparseml/exporters/transforms/utils/test_matching.py
corey-nm: Since the elements of parents/children are nodes i wasn't sure how to match against them without transforming it in some way

tests/sparseml/exporters/transforms/utils/test_matching.py
corey-nm: +1, as discussed in slack will be doing this at a later point

src/sparseml/exporters/transforms/utils/matching.py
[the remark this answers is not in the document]
bfineran: ah I see, so the only issue would be if we want to skip over any of the first `k` inputs to match against any later ones, then they'd need to add placeholder `[]`.

like if we want to match to a QuantizeLinear with an initializer zero point for example it would be `[[], [], [INITIALIZER]]`?

bfineran (approved):
LGTM pending ACK of comments

src/sparseml/exporters/transforms/utils/matching.py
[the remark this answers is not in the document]
corey-nm: Yep exactly!

tests/sparseml/exporters/transforms/utils/test_matching.py
corey-nm: Yeah I like that idea, will do when more rigorous testing is added for  this function

PR #1216, opened 2022-12-07
Stop integration tests on first failure

KSGulin:
Our set of integration tests take ~10 minutes to run on GHA. This can add up to a significant amount of waiting time when bug hunting. This PR updates the integration test command to stop after the first error is encountered to be able to cycle on fixes faster

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,7 @@ test:
 # run integration tests
 testinteg:
 	@echo "Running integration tests";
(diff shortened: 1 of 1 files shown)


corey-nm (approved):
🚀

bfineran:
think this is ok just for integration tests given they take longer, but how much information do we lose if there are multiple bugs?

KSGulin:
@bfineran there's a bit of information loss, but the integration tests can be re-run locally to get the full scope of errors and the tradeoff for the time saved on cycling on a fix for integration tests seems worthwhile

bfineran (approved):

PR #1221, opened 2022-12-08
[Export Refactor] Adding ConvToQLinearConv transform

corey-nm:
Original pass is here https://github.com/neuralmagic/sparseml/blob/9778cec8684d3fcb683504eef88655d9a4607fd6/src/sparseml/pytorch/sparsification/quantization/quantize_qat_export.py#L357

Which is called from _convert_quantizable_ops

Test plan:
unit tests included

### DIFF
--- a/src/sparseml/exporters/transforms/conv_to_qlinearconv.py
+++ b/src/sparseml/exporters/transforms/conv_to_qlinearconv.py
@@ -27,11 +27,15 @@ from sparseml.exporters.transforms.utils.matching import (
     MatchResult,
     get_structural_matches,
 )
-from sparseml.onnx.utils.graph_editor import (
+from sparseml.onnx.utils import (
     ONNXGraph,
+    get_init_by_name,
+    get_node_output_nodes,
     remove_node_and_params_from_graph,
 )
-from sparseml.onnx.utils.helpers import get_init_by_name, get_node_output_nodes
+
+
+__all__ = ["ConvToQLinearConv"]
 
 
 class ConvToQLinearConv(OnnxTransform):
@@ -43,11 +47,11 @@ class ConvToQLinearConv(OnnxTransform):
                         |
     input               QuantizeLinear
     |                   |
-    DequantizeLinear    DequantizeLinear
-    |                   |
-            Conv
-            |
-            QuantizeLinear
+    DequantizeLinear    DequantizeLinear    bias (optional)
+    |                   |                   |
+                        Conv
+                        |
+                        QuantizeLinear
     ```
 
     into
@@ -68,13 +72,15 @@ class ConvToQLinearConv(OnnxTransform):
         )
         for match in matches:
             self._do_transform(model, match)
-        ONNXGraph(model).sort_nodes_topologically()
+        graph = ONNXGraph(model)
+        graph.sort_nodes_topologically()
+        graph.delete_unused_initializers()
         return model
 
     def _do_transform(self, model: ModelProto, match: MatchResult):
         conv_node = match.node
         (input_dequant,) = match.parents[0]
-        weight_init, weight_quant, weight_dequant = match.parents[1]
+        _, weight_quant, weight_dequant = match.parents[1]
         (output_quant,) = match.children[0]
 
         weight_quantize_params = get_quantization_params(
@@ -97,9 +103,8 @@ class ConvToQLinearConv(OnnxTransform):
         model.graph.initializer.append(quantized_weight_initializer)
 
         # get qconv inputs and outputs
-        qconv_input = input_dequant.input[0]
         qconv_inputs = [
-            qconv_input,  # x
+            input_dequant.input[0],  # x
             input_dequant.input[1],  # x_scale
             input_dequant.input[2],  # x_zero_point
             quantized_weight_name,  # w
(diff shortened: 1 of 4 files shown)


src/sparseml/exporters/transforms/conv_to_qlinearconv.py
```diff
@@ -0,0 +1,149 @@
...
+    delete_quant_node,
+    get_quantization_params,
+    quantize_array,
+)
+from sparseml.exporters.transforms.utils.matching import (
+    INITIALIZER_MATCH,
+    MatchResult,
+    get_structural_matches,
+)
+from sparseml.onnx.utils.graph_editor import (
```
dbogunowicz, line 30:
```suggestion
from sparseml.onnx.utils import (
```
suffices

src/sparseml/exporters/transforms/utils/helpers.py
```diff
@@ -0,0 +1,221 @@
...
+        node.op_type in _QUANTIZE_OP_NAMES
+    ), "Op Type must be either QuantizeLinear or DequantizeLinear, found {} ".format(
+        node.op_type
+    )
+    if keep_weight:
+        del node.input[0]
+    remove_node_and_params_from_graph(model, node)
+
+
+def check_for_sequence_of_children_nodes(
```
dbogunowicz, line 105: this and the next function (checking for sequence of parent nodes) seem to be redundant given your matching functionality.

tests/sparseml/exporters/transforms/test_conv_to_qlinearconv.py
```diff
@@ -0,0 +1,106 @@
...
+        "weight_quant",
+        "weight_dequant",
+        "conv",
+        "output_quant",
+    ]
+    return model
+
+
+def test_vanilla(onnx_model: onnx.ModelProto):
+    onnx_model = ConvToQLinearConv().apply(onnx_model)
```
dbogunowicz, line 95:
```suggestion
    onnx_model = ConvToQLinearConv(onnx_model)
```

its nicer :)

tests/sparseml/exporters/transforms/test_conv_to_qlinearconv.py
```diff
@@ -0,0 +1,106 @@
...
+    onnx.checker.check_model(onnx_model)
+    assert [i.name for i in onnx_model.graph.initializer] == [
```
dbogunowicz, line 97: nice, need to employ this check as well

src/sparseml/exporters/transforms/conv_to_qlinearconv.py
```diff
@@ -0,0 +1,149 @@
...
+                [INITIALIZER_MATCH, "QuantizeLinear", "DequantizeLinear"],
+            ],
+            children_ops=[["QuantizeLinear"]],
+        )
+        for match in matches:
+            self._do_transform(model, match)
+        ONNXGraph(model).sort_nodes_topologically()
+        return model
+
+    def _do_transform(self, model: ModelProto, match: MatchResult):
```
dbogunowicz, line 74: Want to stick to my convention and have those functions outside the class? No reason for it, other than readability (however, this is subjective) and consistency across all the passes -> for the reader's comfort.

dbogunowicz: Also, this function should be public and have some docstring with explanation

src/sparseml/exporters/transforms/conv_to_qlinearconv.py
```diff
@@ -0,0 +1,149 @@
...
+            qconv_kwargs.update(attribute_to_kwarg(attribute))
+
+        # create qconv node and add it to graph
+        qconv_node = helper.make_node(
+            "QLinearConv", qconv_inputs, [qconv_output], qconv_name, **qconv_kwargs
+        )
+        model.graph.node.append(qconv_node)
+
+        # delete original conv and folded quantization ops
+        remove_node_and_params_from_graph(model, conv_node)
```
dbogunowicz, line 142: per Ben: let's do the removal of the `match.nodes` in the `transformation` logic (outside of this function)

dbogunowicz (commented):
Few nits, otherwise LGTM

src/sparseml/exporters/transforms/conv_to_qlinearconv.py
```diff
@@ -0,0 +1,149 @@
...
+)
+from sparseml.onnx.utils.helpers import get_init_by_name, get_node_output_nodes
+
+
+class ConvToQLinearConv(OnnxTransform):
+    """
+    Transforms
+
+    ```
+                        weight (initializer)
```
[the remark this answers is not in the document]
corey-nm: Hah wait what don't you like about this format?!

src/sparseml/exporters/transforms/conv_to_qlinearconv.py
corey-nm: I think having interior private function makes more sense - we are exporting the Transform object, not the singular function for doing one transform operation. Since the match object that is passed in is tightly coupled to the match result, you shouldn't really use it separately from that IMO. Unless we move the transform to actually take in the nodes like it was done before, then having a public function exterior would be fine.

bfineran (commented):
LGTM overall, let's be sure to sync up on where specific helpers are going to end up, think I'm seeing a few things across PRs

src/sparseml/exporters/transforms/conv_to_qlinearconv.py
bfineran: agree with your logic @corey-nm. to the point of standardizing to moving things out - it could help with variable scoping, so things don't leak

src/sparseml/exporters/transforms/conv_to_qlinearconv.py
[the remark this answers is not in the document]
dbogunowicz: I tried to mimick the curving arrows in my docstrings in the same fashion as the original codebase did for example.

src/sparseml/exporters/transforms/conv_to_qlinearconv.py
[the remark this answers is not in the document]
corey-nm: Okay I'll change them all to _do_conversion during standardization

bfineran (approved):

PR #1225, opened 2022-12-08
[Export Refactor]Adding  GemmToQLinearMatMul transform

corey-nm:
Original pass: https://github.com/neuralmagic/sparseml/blob/9778cec8684d3fcb683504eef88655d9a4607fd6/src/sparseml/pytorch/sparsification/quantization/quantize_qat_export.py#L447

Testing plan: unit tests

### DIFF
--- a/src/sparseml/exporters/transforms/gemm_to_qlinearmatmul.py
+++ b/src/sparseml/exporters/transforms/gemm_to_qlinearmatmul.py
@@ -1,0 +1,252 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from onnx import ModelProto, helper, numpy_helper
+
+from sparseml.exporters.transforms.onnx_transform import OnnxTransform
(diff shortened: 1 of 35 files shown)


src/sparseml/exporters/transforms/utils/matching.py
```diff
@@ -333,6 +333,15 @@ def _match_children(
         return False
 
     children = graph.get_node_children(node)
+    if children == [] and all(
+        all(op.startswith("Optional-") for op in ops) for ops in children_ops
```
bfineran, line 337: we should make `"Optional-"` a constant for consistency

src/sparseml/exporters/transforms/gemm_to_qlinearmatmul.py
```diff
@@ -0,0 +1,252 @@
...
+                        |
+    input               QuantizeLinear
+    |                   |
+    DequantizeLinear    DequantizeLinear
+    |                   |
+            Gemm ( with no attributes)
+            |
+            QuantizeLinear
+            |
+            NOT (DequantizeLinear -> Gemm)
```
bfineran, line 54: what does NOT mean here?

corey-nm: Yeah can probably add more documentation on that. Basically if there is a Q->DQ->Gemm after the first gemm, this transform does not apply. Unsure how to capture that.

i.e. you can have:
1. Gemm -> Q -> W/e
2.  Gemm -> Q -> DQ -> Not GEMM

src/sparseml/exporters/transforms/utils/matching.py
corey-nm: Good call, will do

bfineran: bump

src/sparseml/exporters/transforms/gemm_to_qlinearmatmul.py
bfineran: makes sense

bfineran (commented):
lgtm pending comment

PR #1229, opened 2022-12-08
[Export Refactor] Adding FlattenQParams transform

corey-nm:
Pass from https://github.com/neuralmagic/sparseml/blob/9778cec8684d3fcb683504eef88655d9a4607fd6/src/sparseml/pytorch/utils/exporter.py#L666

### DIFF
--- a/src/sparseml/exporters/transforms/flatten_qparams.py
+++ b/src/sparseml/exporters/transforms/flatten_qparams.py
@@ -1,0 +1,60 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numpy
+from onnx import ModelProto, numpy_helper
+
+from sparseml.exporters.transforms.onnx_transform import OnnxTransform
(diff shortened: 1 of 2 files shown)


src/sparseml/exporters/transforms/flatten_qparams.py
```diff
@@ -0,0 +1,60 @@
...
+                scale_init = graph.get_init_by_name(node.input[1])
+                if scale_init is not None and list(scale_init.dims) == [1]:
+                    inits_to_flatten.add(node.input[1])
+
+                    # zero_point is optional AND shape must match
+                    # scale. so if scale is (1,), then so will zero point
+                    if len(node.input) == 3:
+                        inits_to_flatten.add(node.input[2])
+
+        for i, init in enumerate(model.graph.initializer):
```
bfineran, line 48: nit: avoid single character variable names, prefer `idx`

src/sparseml/exporters/transforms/flatten_qparams.py
```diff
@@ -0,0 +1,60 @@
...
+            if init.name not in inits_to_flatten:
+                continue
+            a = numpy_helper.to_array(init)
+            assert a.shape == (1,)
```
bfineran, line 52: is this a safe assertion?

src/sparseml/exporters/transforms/flatten_qparams.py
```diff
@@ -0,0 +1,60 @@
...
+            b = numpy.array(a[0])
+            assert b.shape == ()
+            assert b.dtype == a.dtype
+            model.graph.initializer[i].CopyFrom(
```
bfineran, line 56: cool use of `CopyFrom`. is it in place though? if so, then why do we need to access via initializer list index instead of just using the `init` reference?

corey-nm: Great question, lemme see if it makes a difference

corey-nm: Yep that works, thanks!

bfineran (approved):

PR #1233, opened 2022-12-09
[Export Refactor] Delete Trivial Onnx Adds

dbogunowicz:
Original implementation:
https://github.com/neuralmagic/sparseml/blob/bbc61bec8fd250b0e94a89cbbb129b52da0eb542/src/sparseml/pytorch/utils/exporter.py#L780

### DIFF
--- a/src/sparseml/exporters/transforms/delete_trivial_onnx_adds.py
+++ b/src/sparseml/exporters/transforms/delete_trivial_onnx_adds.py
@@ -1,57 +1,77 @@
-from sparseml.exporters.transforms import OnnxTransform
-from onnx import ModelProto
-import logger
-from onnx import numpy_helper
-from sparseml.onnx.utils import ONNXGraph
-from sparseml.exporters.utils.matching import get_structural_matches
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


corey-nm (changes requested):
Having done just the pass that converts constants to initializers, I think the ordering of these two could cause issues - if they are ordered backwards then this one won't work (e.g. ConstantsToInitializers() -> DeleteTrivialOnnxAdds()).

Options:
1. Change this to look for initializers instead of constants
2. Change this to look for both initializers and constants

Thoughts?

Other than that looks good

src/sparseml/exporters/transforms/delete_trivial_onnx_adds.py
```diff
@@ -0,0 +1,77 @@
...
+        graph = ONNXGraph(model)
+        for match in get_structural_matches(
+            graph, op_type="Add", parent_ops=[[], ["Constant"]]
+        ):
+
+            if should_delete_trivial_onnx_add(match):
+                _LOGGER.debug(f"Matched Identity node: {match.node.name}")
+                constant_node = match.parents[1][0]
+                remove_node_and_params_from_graph(model, match.node)
+                remove_node_and_params_from_graph(model, constant_node)
```
corey-nm, line 72: I think this also needs to reroute the output of parent of add into the children of add (from this line in original impl: https://github.com/neuralmagic/sparseml/blob/bbc61bec8fd250b0e94a89cbbb129b52da0eb542/src/sparseml/pytorch/utils/exporter.py#L798)

PR #1234, opened 2022-12-09
[Export Refactor] Quantize QAT Embedding

dbogunowicz:
Original implementation:
https://github.com/neuralmagic/sparseml/blob/9778cec8684d3fcb683504eef88655d9a4607fd6/src/sparseml/pytorch/sparsification/quantization/quantize_qat_export.py#L1366


src/sparseml/exporters/transforms/utils/matching.py
```diff
@@ -337,7 +337,13 @@ def _match_children(
     #       is the graph output.
     # this is a difference in behavior to get_node_parents, which replaces input
     # with None, instead of removing it.
-    if not (len(children_ops) <= len(children)):
+    non_optional_children_ops = [
```
dbogunowicz, line 340: @corey-nm opinion?

src/sparseml/exporters/transforms/quantize_qat_embedding.py
```diff
@@ -0,0 +1,160 @@
...
+
+    # detect QDQ block on output
+    children = match.children
+    qdq_output = False
+    if children:
+        if len(children[0]) == 2:
+            if assert_node_type(children[0][0], "QuantizeLinear") and assert_node_type(
+                children[0][1], "DequantizeLinear"
+            ):
+                qdq_output = True
```
corey-nm, line 66: a couple notes here:
1. children will always be a list so you can remove the if children
2. our matching code should guarantee that `len(children[i]) == len(children_ops[i])`, so you can get rid of the second if branch
3. if the values in the list are not None, then they should be guaranteed to be the op types that you specify in children ops, so i think you can just check if both are not none

thoughts?
```suggestion
    opt_quant, opt_dequant = match.children
    qdq_output = opt_quant is not None and opt_dequant is not None
```

src/sparseml/exporters/transforms/utils/matching.py
corey-nm: Is this to handle cases where the children are all optional? I added something similar in another PR cause I ran into this case when testing as well 😁

corey-nm (approved):
lgtm!

src/sparseml/exporters/transforms/utils/matching.py
dbogunowicz: as discussed -> your PR solves exactly this problem. will remove this hack before landing

bfineran (approved):
LGTM leaving up to @corey-nm if changes are necessary before land

PR #1235, opened 2022-12-09
[Exporter Refactor] Adding FoldConvDivBn

corey-nm:
Add this pass https://github.com/neuralmagic/sparseml/blob/9778cec8684d3fcb683504eef88655d9a4607fd6/src/sparseml/pytorch/sparsification/quantization/quantize_qat_export.py#L176

### DIFF
--- a/src/sparseml/exporters/transforms/fold_conv_div_bn.py
+++ b/src/sparseml/exporters/transforms/fold_conv_div_bn.py
@@ -1,0 +1,100 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


src/sparseml/exporters/transforms/fold_conv_div_bn.py
```diff
@@ -0,0 +1,100 @@
...
+    remove_node_and_params_from_graph,
+)
+
+
+__all__ = ["FoldConvDivBn"]
+
+
+class FoldConvDivBn(OnnxTransform):
+    """
+    Transforms
```
bfineran, line 36: need to update docstring here to state purpose of transform (ie that this specifically gets rid of the undo step of the fake batch norm fold that happens in pytorch qat)... confusing even for the initialized...

PR #1238, opened 2022-12-09
[Exporter Refactor] Adding MatMulToMatMulIntegerAddCastMul transform

corey-nm:
This pass https://github.com/neuralmagic/sparseml/blob/9778cec8684d3fcb683504eef88655d9a4607fd6/src/sparseml/pytorch/sparsification/quantization/quantize_qat_export.py#L1007


src/sparseml/exporters/transforms/matmul_to_matmulinteger_add_cast_mul.py
```diff
@@ -0,0 +1,167 @@
...
+
+    ```
+    | Starting with:
+    |                        weight (initializer)
+    |                            |
+    |                      QuantizeLinear
+    |                            |
+    |          INPUT     DequantizeLinear
+    |            |               |
+    |     DequantizeLinear   Transpose
```
bfineran, line 54: noting this transpose should definitely be optional (looking at onnx modifier support)

bfineran (approved):

PR #1241, opened 2022-12-09
[QuantizationModifier] strict mode - raise if unmatched submodules or types

bfineran:
strict by default adds a check that the all overrides and ignores given by the user are actually found in the module. meant to catch any config errors earlier rather than later but gives user the option to turn off

**test_plan:**
* yaml test updated
* existing tests now passing under strict (found and fixed one issue)
* test added to check for expected errors

### DIFF
--- a/src/sparseml/pytorch/sparsification/quantization/quantize.py
+++ b/src/sparseml/pytorch/sparsification/quantization/quantize.py
@@ -111,6 +111,7 @@ def set_quantization_schemes(
     scheme: QuantizationScheme,
     scheme_overrides: Optional[Dict[str, QuantizationScheme]] = None,
     ignore: Optional[List[str]] = None,
+    strict: bool = True,
 ):
     """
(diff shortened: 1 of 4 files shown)


src/sparseml/pytorch/sparsification/quantization/quantize.py
```diff
@@ -308,3 +314,40 @@ def _get_qat_module_mappings() -> Dict[Module, Module]:
...
+    def _get_unmatched_types_or_names(types_or_names):
+        unmatched = []
+        for type_or_name in types_or_names:
+            matched = False
+            for submodule_name, submodule in model.named_modules():
+                if submodule_name.startswith(type_or_name) or (
+                    submodule.__class__.__name__ == type_or_name
+                ):
+                    matched = True
+                    break
```
corey-nm, line 333: optional nit: use `not any(...)` here to make it more clear, unsure if it would help.

also think in this case shorter variable names would reduce noise of this snippet

corey-nm (approved):

PR #1242, opened 2022-12-09
[Exporter Refactor] Adding PropagateEmbeddingQuantization transform

corey-nm:
This pass https://github.com/neuralmagic/sparseml/blob/9778cec8684d3fcb683504eef88655d9a4607fd6/src/sparseml/pytorch/sparsification/quantization/quantize_qat_export.py#L1719


bfineran (approved):
this could probably just be combined with the main embedding quantization pass given that we're just skipping a bunch of non-data changing ops. additionally if this pass causes issues we can talk to runtime, suspect that they already optimize for this

PR #1244, opened 2022-12-09
[Exporter Refactor] Adding RemoveDuplicateQConvWeights transform

corey-nm:
This pass 
https://github.com/neuralmagic/sparseml/blob/9778cec8684d3fcb683504eef88655d9a4607fd6/src/sparseml/pytorch/sparsification/quantization/quantize_qat_export.py#L1273

### DIFF
--- a/src/sparseml/exporters/transforms/remove_duplicate_qconv_weights.py
+++ b/src/sparseml/exporters/transforms/remove_duplicate_qconv_weights.py
@@ -26,7 +26,8 @@ __all__ = ["RemoveDuplicateQConvWeights"]
 
 class RemoveDuplicateQConvWeights(OnnxTransform):
     """
-    Removes weights of QLinearConv and ConvInteger that are the same.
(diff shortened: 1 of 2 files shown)


KSGulin (changes requested):
Looks good to me pending comment

src/sparseml/exporters/transforms/remove_duplicate_qconv_weights.py
```diff
@@ -0,0 +1,77 @@
...
+from sparseml.exporters.transforms.onnx_transform import OnnxTransform
+from sparseml.onnx.utils import ONNXGraph
+
+
+__all__ = ["RemoveDuplicateQConvWeights"]
+
+
+class RemoveDuplicateQConvWeights(OnnxTransform):
+    """
+    Removes weights of QLinearConv and ConvInteger that are the same.
```
bfineran, line 29: nit: "removes" -> "deduplicates" or something that doesn't imply that we're deleting all

bfineran (approved):

PR #1246, opened 2022-12-12
GPU Synchronisation bug

Chibukach:
There is a GPU synchronisation error that occurs after an image classification training runs to completion in Kubeflow DDP 

"""
ProcessGroupNCCL.cpp:325] Some NCCL operations have failed or timed out. Due to the asynchronous nature of CUDA kernels, subsequent GPU operations might run on corrupted/incomplete data. To avoid this inconsistency, we are taking the entire process down.
terminate called after throwing an instance of 'std::runtime_error'
  what():  NCCL error: unhandled system error, NCCL version 2.7.8
ncclSystemError: System call (socket, malloc, munmap, etc) failed.
Fatal Python error: Aborted
"""

This PR fixes the GPU synchronisation error by making sure that the image classification validation (post training) that runs only on the rank 0 GPU uses the local (rank 0) copy of the model rather than the distributed version. This avoids the NCCL communication during validation reaching out to other (rank > 0) processes that have already completed their training execution. 

Reproduction:
I ran the training command listed under transfer-classification.md in this link in kubeflow:
https://sparsezoo.neuralmagic.com/models/cv%2Fclassification%2Fresnet_v1-50%2Fpytorch%2Fsparseml%2Fimagenet%2Fpruned95-none

### DIFF
--- a/src/sparseml/pytorch/image_classification/utils/trainer.py
+++ b/src/sparseml/pytorch/image_classification/utils/trainer.py
@@ -233,14 +233,11 @@ class ImageClassificationTrainer(Trainer):
         _LOGGER.info(f"Applied {self.recipe_path} to manager")
 
     def _initialize_module_tester(self):
-        """
-        This uses self.model.module unlike the ModuleTrainer which uses self.model
-        so that the validation runs on the local copy of the model on the rank 0 gpu
-        rather than on the distributed model
-        """
-        
+        # This uses self.model.module unlike the ModuleTrainer which uses self.model
+        # so that the validation runs on the local copy of the model on the rank 0 gpu
+        # rather than on the distributed model
         tester = ModuleTester(
-            module=self.model.module,
+            module=self.model.module if is_parallel_model(self.model) else self.model,
             device=self.device,
             loss=self.val_loss,
             loggers=self.loggers,
--- a/src/sparseml/pytorch/image_classification/utils/trainer.py
+++ b/src/sparseml/pytorch/image_classification/utils/trainer.py
@@ -233,8 +233,14 @@ class ImageClassificationTrainer(Trainer):
         _LOGGER.info(f"Applied {self.recipe_path} to manager")
 
     def _initialize_module_tester(self):
+        """
+        This uses self.model.module unlike the ModuleTrainer which uses self.model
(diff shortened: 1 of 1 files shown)


src/sparseml/pytorch/image_classification/utils/trainer.py
```diff
@@ -233,8 +233,14 @@ def _apply_one_shot(self):
...
     def _initialize_module_tester(self):
+        """
+        This uses self.model.module unlike the ModuleTrainer which uses self.model
+        so that the validation runs on the local copy of the model on the rank 0 gpu
+        rather than on the distributed model
+        """
+        
         tester = ModuleTester(
-            module=self.model,
+            module=self.model.module,
```
bfineran, line 243: this would fail if the model is not DP/DDP right?
let's add a check for `is_parallel_module` (https://github.com/neuralmagic/sparseml/blob/bbc61bec8fd250b0e94a89cbbb129b52da0eb542/src/sparseml/pytorch/utils/model.py#L355) before unwrapping

src/sparseml/pytorch/image_classification/utils/trainer.py
```diff
@@ -233,8 +233,14 @@ def _apply_one_shot(self):
         _LOGGER.info(f"Applied {self.recipe_path} to manager")
 
     def _initialize_module_tester(self):
+        """
+        This uses self.model.module unlike the ModuleTrainer which uses self.model
+        so that the validation runs on the local copy of the model on the rank 0 gpu
+        rather than on the distributed model
```
bfineran, line 239: nit: for private methods, use in-line comments (`# `) instead of block. we generally standardize to reserving block comments for top level file stuff and pydocs

bfineran (changes requested):

src/sparseml/pytorch/image_classification/utils/trainer.py
Chibukach: fair point

bfineran (approved):

tdg5 (approved):

markurtz (approved):

tdg5 (approved):

PR #1251, opened 2022-12-13
[Exporter Refactor] Fixing matching logic of qlinear transforms

corey-nm:
Old logic allowed QuantizeLinear OR DequantizeLinear as parent[0]. This moves logic to match that.

### DIFF
--- a/src/sparseml/exporters/transforms/conv_to_qlinearconv.py
+++ b/src/sparseml/exporters/transforms/conv_to_qlinearconv.py
@@ -100,6 +100,10 @@ class ConvToQLinearConv(OnnxTransform):
         # sanity check - matching should ensure this
         assert weight_quantize_params.target is not None
 
+        # can fold the input/output quant ops if they are trivial
+        fold_input_quant = input_quant.op_type == "DequantizeLinear"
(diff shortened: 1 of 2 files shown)


src/sparseml/exporters/transforms/conv_to_qlinearconv.py
```diff
@@ -79,7 +90,7 @@ def transform(self, model: ModelProto) -> ModelProto:
 
     def _do_transform(self, model: ModelProto, match: MatchResult):
         conv_node = match.node
-        (input_dequant,) = match.parents[0]
+        (input_quant,) = match.parents[0]
         _, weight_quant, weight_dequant = match.parents[1]
         (output_quant,) = match.children[0]
```
corey-nm, line 95: potentially need to add the fold_input_quant/fold_output_quant back in?

corey-nm: Added

bfineran (approved):
LGTM, was this causing an issue?

dbogunowicz (approved):

corey-nm:
Yes, the matching QuantizeLinear or DequantizeLinear was required for resnet50 actually. That then means the folding logic is also required

PR #1252, opened 2022-12-14
[Exporter Refactor] Fixing conv integer node deletion & sorting in core ops

corey-nm:
Title - matches old behavior now

### DIFF
--- a/src/sparseml/exporters/transforms/convert_quantizable_matmul.py
+++ b/src/sparseml/exporters/transforms/convert_quantizable_matmul.py
@@ -149,7 +149,6 @@ class ConvertQuantizableMatmul(OnnxTransform):
             # Convert
             model = convert_matmul_to_quantized(match, model)
             remove_node_and_params_from_graph(model, match.node)
-            ONNXGraph(model).sort_nodes_topologically()
 
             count_converted_nodes += 1
 
@@ -158,4 +157,5 @@ class ConvertQuantizableMatmul(OnnxTransform):
                 f"Converted {count_converted_nodes} quantizable MatMul ops "
                 "to QLinearMatMul"
             )
+        ONNXGraph(model).sort_nodes_topologically()
         return model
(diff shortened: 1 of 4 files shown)


src/sparseml/exporters/transforms/gemm_to_matmulinteger_add_cast_mul.py
```diff
@@ -104,6 +104,8 @@ def transform(self, model: ModelProto) -> ModelProto:
 
             self._do_transform(model, match, attr)
 
+        graph = ONNXGraph(model)
+        graph.sort_nodes_topologically()
```
dbogunowicz, line 108: Why is this necessary here?

src/sparseml/exporters/transforms/convert_quantizable_matmul.py
```diff
@@ -158,4 +157,5 @@ def transform(self, model: ModelProto) -> ModelProto:
                 f"Converted {count_converted_nodes} quantizable MatMul ops "
                 "to QLinearMatMul"
             )
+        ONNXGraph(model).sort_nodes_topologically()
```
dbogunowicz, line 160: Why is this necessary here?

dbogunowicz (approved):
thanks for catching that!

src/sparseml/exporters/transforms/convert_quantizable_matmul.py
corey-nm: Just moving it outside the loop so we are only doing it once instead of every time we match. not a bug fix, but an easy optimization

src/sparseml/exporters/transforms/gemm_to_matmulinteger_add_cast_mul.py
corey-nm: Any transform that adds nodes will require a call to this

bfineran (approved):

PR #1253, opened 2022-12-14
[Exporter Refactor] Don't print onnx models on failure

corey-nm:
This was causing long debug/test failures because it was printing the whole onnx model to stdout.

### DIFF
--- a/src/sparseml/onnx/utils/helpers.py
+++ b/src/sparseml/onnx/utils/helpers.py
@@ -90,7 +90,7 @@ def validate_onnx_file(path: str):
         if not onnx_model.opset_import:
             raise ValueError("could not parse opset_import")
     except Exception as err:
-        raise ValueError("file at {} is not a valid onnx model: {}".format(path, err))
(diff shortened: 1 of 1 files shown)


dbogunowicz (commented):
Can you paste in here the behaviour pre/post this PR?

src/sparseml/onnx/utils/helpers.py
```diff
@@ -90,7 +90,7 @@ def validate_onnx_file(path: str):
         if not onnx_model.opset_import:
             raise ValueError("could not parse opset_import")
     except Exception as err:
-        raise ValueError("file at {} is not a valid onnx model: {}".format(path, err))
```
corey-nm, line 93: @dbogunowicz `path` here can be an onnx.ModelProto, which means it was printing out the entire serialized resnet50 model including parameters. same thing for the value error below

dbogunowicz (approved):

bfineran (approved):

PR #1255, opened 2022-12-14
[Exporter Refactor] Fix duplicate qconv

corey-nm:
The grouping logic may compare weights of different sizes, so there was a warning before about this


dbogunowicz (approved):
lgtm

bfineran (approved):
nice catch. first time I can think of not wanting broadcast to work out of the box 😄

PR #1257, opened 2022-12-14
[Exporter Refactor] Adding DeleteRepeatedQdq transform

corey-nm:
Adds pass from https://github.com/neuralmagic/sparseml/blob/main/src/sparseml/pytorch/sparsification/quantization/quantize_qat_export.py#L265

### DIFF
--- a/src/sparseml/exporters/transforms/delete_repeated_qdq.py
+++ b/src/sparseml/exporters/transforms/delete_repeated_qdq.py
@@ -15,7 +15,8 @@
 from onnx import ModelProto
 
 from sparseml.exporters.transforms.onnx_transform import OnnxTransform
-from sparseml.onnx.utils.graph_editor import ONNXGraph
+from sparseml.exporters.transforms.utils import assert_node_type
+from sparseml.onnx.utils import ONNXGraph
 
 
 __all__ = ["DeleteRepeatedQdq"]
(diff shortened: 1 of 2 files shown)


src/sparseml/exporters/transforms/delete_repeated_qdq.py
```diff
@@ -0,0 +1,70 @@
...
+    ```
+    """
+
+    def transform(self, model: ModelProto) -> ModelProto:
+        graph = ONNXGraph(model)
+        nodes_to_delete = []
+        quant_nodes = [n for n in model.graph.node if n.op_type == "QuantizeLinear"]
+        for quant_node_1 in quant_nodes:
+            dequant_node_1 = graph.get_node_single_child(quant_node_1)
+            if not dequant_node_1 or dequant_node_1.op_type != "DequantizeLinear":
```
dbogunowicz, line 50: We have a helper function for that:
```suggestion
            if not assert_op_type(dequant_node_1, "DequantizeLinear"):
```

dbogunowicz (approved):

bfineran (approved):
let's add a small note to the docstring or inline that this pass may affect graph outputs and should be avoided by having more careful quantized graph construction

src/sparseml/exporters/transforms/delete_repeated_qdq.py
```diff
@@ -0,0 +1,71 @@
...
+from sparseml.exporters.transforms.onnx_transform import OnnxTransform
+from sparseml.exporters.transforms.utils import assert_node_type
+from sparseml.onnx.utils import ONNXGraph
+
+
+__all__ = ["DeleteRepeatedQdq"]
+
+
+class DeleteRepeatedQdq(OnnxTransform):
+    """
```
corey-nm, line 26:
```suggestion
    """
    Removes a QDQ that immediately follows another QDQ.
    NOTE: this will modify graph outputs and does not
    guarantee identical graph behavior. This should be avoided
    and handled with better graph construction.  
```

PR #1260, opened 2022-12-15
[Exporters Refactor] Fix failing base tests

dbogunowicz:
Remove torch from base tests dependency.


src/sparseml/exporters/transforms/utils/helpers.py
```diff
@@ -125,22 +126,17 @@ def assert_node_type(node: NodeProto, op: Union[List[str], Set[str], str]) -> bo
...
-        zero_point = zero_point.item()
-
-    quant_tensor = torch.quantize_per_tensor(tensor, scale, zero_point, tensor_dtype)
-    return quant_tensor.int_repr().numpy()
+        return quantize_array_torch(array, scale, zero_point, dtype)
+    except ModuleNotFoundError as err:
+        _LOGGER.debug(f"Error: {err}. Defaulting to numpy implementation.")
+        dmin = numpy.iinfo(dtype).min
+        dmax = numpy.iinfo(dtype).max
+        return ((array / scale).round() + zero_point).clip(dmin, dmax).astype(dtype)
```
corey-nm, line 139: Could we just always use this numpy version? I assume this is the same thing torch is doing?

dbogunowicz: @bfineran: "I think it was some rounding error reason relative to the torch quantization execution for why we switched"

dbogunowicz: I ran a few tests to check whether the two implementations match. They did, but I was probably not too thorough to spot the difference.

corey-nm: "Fun" challege: try to find the pytorch implementation of this

corey-nm: This is the best I could find about what they do. https://github.com/pytorch/pytorch/wiki/torch_quantization_design_proposal#quantized-representation

I couldn't find the C++ code where this actually happens 😩 

We could try running all our regression testing with this version?

corey-nm: Okay I was able to find some of the cuda kernels at least https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/quantized/cuda/FakeQuantizeCore.cu

Seems like they may do some weird rounding stuff. I suspect your implementation may different for values in a certain range?

src/sparseml/exporters/transforms/utils/helpers.py
```diff
@@ -125,22 +126,17 @@ def assert_node_type(node: NodeProto, op: Union[List[str], Set[str], str]) -> bo
...
-
-    tensor = torch.Tensor(array.copy()).to(torch.float32)
-    if isinstance(scale, numpy.ndarray):
-        scale = scale.item()
-    if isinstance(zero_point, numpy.ndarray):
-        zero_point = zero_point.item()
-
-    quant_tensor = torch.quantize_per_tensor(tensor, scale, zero_point, tensor_dtype)
-    return quant_tensor.int_repr().numpy()
+        return quantize_array_torch(array, scale, zero_point, dtype)
```
corey-nm, line 134: I think we should keep the logic for torch quantization in this file. It's not used anywhere else and it helps to keep number of files & levels of indirection down if we have it here.

corey-nm (approved):
💪

PR #1261, opened 2022-12-15
[QuantizationModifier] take ownership of add_observers_, unit test fixes

bfineran:
this PR addresses three issues found in testing of the new `QuantizationModifier`:

1. addresses an break in recipe template tests
2. uses `FakeQuantizeBase` for inheritance comparisons where possible (introduced as default in later versions of torch)
3. torch 1.9 tests revealed that torch quantization will inject its own activation post process hooks into certain activations without any way to disable. this PR adds its own implementation of add observers that does not include this override. (fixes issues with quantization of `Sigmoid` and other activations)
4. previously, the new modifier still deferred to torch for quantization of output activations - this was a known hole as we try to quantize "everything" while torch mostly keeps to a white-list. as part of the activation fix in (3), we update our add observers implementation to quantize activations of *all* target modules as determined by `QuantizationScheme` propagation
5. fix for compatibility with the `FloatFunctional` module used for resnet quantization

**test_plan:**
* tests updated to ensure that output activation quantization is tested for existence and adherence to the scheme

### DIFF
--- a/src/sparseml/pytorch/sparsification/quantization/quantize.py
+++ b/src/sparseml/pytorch/sparsification/quantization/quantize.py
@@ -18,6 +18,8 @@ structured configurations
 """
 from typing import Dict, List, Optional
 
+import torch
+from packaging import version
 from torch.nn import Identity, Module
 
 from sparseml.pytorch.sparsification.quantization.constants import (
@@ -50,6 +52,7 @@ __all__ = [
     "set_quantization_schemes",
     "set_qconfigs_from_quantization_schemes",
     "add_input_activation_quant_wrappers",
+    "add_output_activation_observers",
     "raise_if_torch_quantization_not_available",
 ]
 
@@ -59,11 +62,15 @@ def is_qat_helper_module(module: Module) -> bool:
     :param module: module to check
     :return: True if module is an instance of a torch QAT helper class
     """
+    # prefer FakeQuantizeBase which was introduced around torch 1.9
+    fake_quantize_class = getattr(
+        torch_quantization, "FakeQuantizeBase", torch_quantization.FakeQuantize
+    )
     return isinstance(
         module,
         (
+            fake_quantize_class,
             torch_quantization.ObserverBase,
-            torch_quantization.FakeQuantize,
             torch_quantization.DeQuantStub,
             torch_quantization.QuantStub,
             Identity,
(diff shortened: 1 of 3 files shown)


src/sparseml/pytorch/sparsification/quantization/quantize.py
```diff
@@ -232,16 +305,26 @@ def convert_module_qat_from_schemes(module: Module):
...
+    )
     torch_quantization.convert(
-        module, mapping=mapping, inplace=True, remove_qconfig=False
+        module,
+        mapping=_get_qat_module_mappings(),
+        inplace=True,
+        remove_qconfig=False,
+        **convert_kwargs,
     )
-    torch_quantization.add_observer_(module, non_leaf_module_list=set(mapping.values()))
```
corey-nm, line 239: So the new add_output_activation_observers replaces this call?

bfineran: yup

src/sparseml/pytorch/sparsification/quantization/quantize.py
```diff
@@ -214,6 +221,72 @@ def add_input_activation_quant_wrappers(module: Module) -> Module:
...
+def add_output_activation_observers(module: Module):
+    """
+    implementation of torch.quantization add_observers_ that only adds observers
+    according to attached quantization_scheme properties. the existing implementation
+    (1.9+) includes its own logic for propagating including overriding set qconfigs
+    for certain activations without the ability to disable this behavior
+
+    :param module: module to add output activation observers to
+    """
+    # adapted from torch/ao/quantization/quantize.py::_add_observer_
```
corey-nm, line 233:
```suggestion
    # adapted from torch/ao/quantization/quantize.py::_add_observer_
    # source: https://github.com/pytorch/pytorch/blob/v1.13.0/torch/ao/quantization/quantize.py#L135
```

corey-nm: Or i guess the 1.9 version since it seems like it was that version https://github.com/pytorch/pytorch/blob/880c811aeb6937a44d5cc700ab3d802f95538aa9/torch/quantization/quantize.py#L90

bfineran: thanks, added with a flake8 override for line length

bfineran: it's adapted across multiple versions for compatibility

src/sparseml/pytorch/sparsification/quantization/quantize.py
```diff
@@ -214,6 +221,72 @@ def add_input_activation_quant_wrappers(module: Module) -> Module:
...
+    device = next(module.parameters()).device
+
+    def _needs_observer(target_module: Module):
```
corey-nm, line 237:
```suggestion
    def _needs_observer(target_module: Module):
        # NOTE: this is a difference from the original implementation, which just checked for qconfig
```

bfineran: they also check for the descendent stuff, just in a different way (ie through leaf module checks and recursion), will throw in a similar comment

corey-nm: Yeah just noticed all that

src/sparseml/pytorch/sparsification/quantization/quantize.py
```diff
@@ -214,6 +221,73 @@ def add_input_activation_quant_wrappers(module: Module) -> Module:
...
+
+        for descendent_module in target_module.modules():
+            if descendent_module is target_module:
+                continue  # skip itself
+            descendent_scheme = getattr(descendent_module, "quantization_scheme", None)
+            if descendent_scheme is not None and (
+                descendent_scheme.output_activations is not None
+            ):
+                # a descendent of this module targets output activations, return False
+                return False
```
corey-nm, line 258: It seems like the 1.9 version has the potential to recurse down more than 1 level, whereas this seems to not recurse in this loop. Is the recursion deeper somewhere else or do we not actually need that?

bfineran: they use `children()` which goes 1 layer deep so need to recurse (because they do module swapping through the overrides dict which we don't need). `modules()` goes through all descendents

corey-nm: Ohhhhh I did not realize modules was everything! nice simplification

src/sparseml/pytorch/sparsification/quantization/quantize.py
```diff
@@ -232,16 +309,26 @@ def convert_module_qat_from_schemes(module: Module):
...
-        module, mapping=mapping, inplace=True, remove_qconfig=False
+        module,
+        mapping=_get_qat_module_mappings(),
+        inplace=True,
+        remove_qconfig=False,
+        **convert_kwargs,
     )
-    torch_quantization.add_observer_(module, non_leaf_module_list=set(mapping.values()))
+    # re-attach any quantization schemes lost during conversion
+    _reattach_quantization_schemes(module)
```
corey-nm, line 325: Is it okay that this was moved to before the addition of observers intead of after like before?

corey-nm (approved):
lgtm, your version is way easier to understand than pytorch lol 🚀 💪

src/sparseml/pytorch/sparsification/quantization/quantize.py
bfineran: yup, doesn't affect anything, had kept convert and add observer together previously since they were both called under the hood under `torch.quantization.prepare_qat`, but now we need the schemes re-attached for the updated add observer

corey-nm: Gotcha, nice!

PR #1263, opened 2022-12-16
[Exporter Refactor] Adding add/delete methods to OnnxTransform

corey-nm:
Also adds
- delete unused_initializers
- graph sorting

to base class so subclasses don't have to call

Testing plan: existing unit tests

### DIFF
--- a/src/sparseml/exporters/transforms/onnx_transform.py
+++ b/src/sparseml/exporters/transforms/onnx_transform.py
@@ -12,13 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
 from abc import abstractmethod
 from typing import Union
 
-from onnx import ModelProto
+from onnx import ModelProto, NodeProto
 
 from sparseml.exporters.transforms import BaseTransform
-from sparseml.onnx.utils import check_load_model, validate_onnx_file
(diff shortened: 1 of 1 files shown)


src/sparseml/exporters/transforms/onnx_transform.py
```diff
@@ -61,5 +82,11 @@ def post_validate(self, model: ModelProto) -> ModelProto:
         :param model: The input ONNX model to be validated
         :return The validated ONNX model
         """
+        model.graph.node.extend(self._nodes_to_add)
+        for node in self._nodes_to_delete:
+            model.graph.node.remove(node)
```
bfineran, line 87: see: https://github.com/neuralmagic/sparseml/blob/6cf2b086441c02647915357ef6ba46b2395de8b2/src/sparseml/onnx/utils/graph_editor.py#L193

bfineran (approved):
LGTM pending suggestion

src/sparseml/exporters/transforms/onnx_transform.py
corey-nm: I originally used this function, but it was doing some weird things in the tests like removing nodes it shouldn't

dbogunowicz: Oh wow, weird

dbogunowicz (approved):

PR #1264, opened 2022-12-16
[Exporter Refactor] Standardize trivial transforms

corey-nm:
First pass at standardization. These passes don't delete nodes or do anything complicated.

### DIFF
--- a/src/sparseml/exporters/transforms/flatten_qparams.py
+++ b/src/sparseml/exporters/transforms/flatten_qparams.py
@@ -52,12 +52,12 @@ class FlattenQParams(OnnxTransform):
         for init in model.graph.initializer:
             if init.name not in inits_to_flatten:
                 continue
-            _LOGGER.debug(f"Flattening initializer {init.name}")
+            self.log_match(init)
             a = numpy_helper.to_array(init)
             assert a.shape == (1,)
             b = numpy.array(a[0])
             assert b.shape == ()
             assert b.dtype == a.dtype
             init.CopyFrom(numpy_helper.from_array(b, name=init.name))
-        _LOGGER.info(f"Flattened {len(inits_to_flatten)} initializers")
+
         return model
--- a/src/sparseml/exporters/transforms/flatten_qparams.py
+++ b/src/sparseml/exporters/transforms/flatten_qparams.py
@@ -12,8 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import logging
-
 import numpy
 from onnx import ModelProto, numpy_helper
 
@@ -22,8 +20,6 @@ from sparseml.onnx.utils.graph_editor import ONNXGraph
 
 
 __all__ = ["FlattenQParams"]
-
-_LOGGER = logging.getLogger(__name__)
 
 
 class FlattenQParams(OnnxTransform):
--- a/src/sparseml/exporters/transforms/flatten_qparams.py
+++ b/src/sparseml/exporters/transforms/flatten_qparams.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
+
 import numpy
 from onnx import ModelProto, numpy_helper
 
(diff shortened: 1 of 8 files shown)


src/sparseml/exporters/transforms/flatten_qparams.py
```diff
@@ -21,6 +23,8 @@
 
 __all__ = ["FlattenQParams"]
 
+_LOGGER = logging.getLogger(__name__)
```
bfineran, line 26: could add a shared logger in the base `Transform` class that prepends standard information about the transform like `self.__class__.__name__` to log calls

src/sparseml/exporters/transforms/flatten_qparams.py
```diff
@@ -48,11 +52,12 @@ def transform(self, model: ModelProto) -> ModelProto:
         for init in model.graph.initializer:
             if init.name not in inits_to_flatten:
                 continue
+            _LOGGER.debug(f"Flattening initializer {init.name}")
```
bfineran, line 55: note `%` style formatting preferred for debug logs as string formatting is only evaluated on debug call time

src/sparseml/exporters/transforms/flatten_qparams.py
```diff
@@ -48,11 +52,12 @@ def transform(self, model: ModelProto) -> ModelProto:
...
             a = numpy_helper.to_array(init)
             assert a.shape == (1,)
             b = numpy.array(a[0])
             assert b.shape == ()
             assert b.dtype == a.dtype
             init.CopyFrom(numpy_helper.from_array(b, name=init.name))
-
+        _LOGGER.info(f"Flattened {len(inits_to_flatten)} initializers")
```
bfineran, line 62: whats the reasoning for info logging all the minor transforms as opposed to just the main QDQ conversions? not sure I would care about this outside of debug and might prefer to just parse through the main transforms' logs

src/sparseml/exporters/transforms/initializers_to_uint8.py
```diff
@@ -33,6 +37,7 @@ def transform(self, model: ModelProto) -> ModelProto:
             arr_int8 = numpy_helper.to_array(init)
             if arr_int8.dtype != numpy.int8:
                 continue
+            _LOGGER.debug(f"Matched {init.name}")
```
bfineran, line 40: yeah definitely thinking having a standard transform method to handle logging will be helpful

src/sparseml/exporters/transforms/skip_input_quantize.py
```diff
@@ -21,6 +21,8 @@
 from sparseml.onnx.utils import ONNXGraph
 
 
+__all__ = ["SkipInputQuantize"]
```
bfineran, line 24: nice catch

bfineran (commented):
added a few suggestions that will likely hold to the other PRs

src/sparseml/exporters/transforms/flatten_qparams.py
corey-nm: Consistency mainly, I'm fine with debug for the non-core ones

src/sparseml/exporters/transforms/flatten_qparams.py
corey-nm: Are you sure about this? `"%d" % 2` returns a string object, not a special formatting object which makes me think it doesn't actually do lazy evaluation?

bfineran: so you'd do `.debug("foo %s", my_string)`

honestly overhead here is minimal since we're running through once, the main thing that is delayed I believe is calling `str` on the injected variable(s). we take more care with this in say the deepsparse server where we could have lots of concurrent calls running in prod (not debug mode)

corey-nm: Ohhhhh didn't realize you could pass format args into debug, nice

bfineran (approved):

PR #1266, opened 2022-12-16
[Exporter Refactor] Standardizing MatMulToQLinearMatMul

corey-nm:
Depends on #1263 and #1262 

- Renames ConvertQuantizableMatMul to more explicitly `MatMulToQLinearMatMul`
- uses any_of
- uses deferred deletion of nodes
- Adds logging
- standardize docstring


src/sparseml/exporters/transforms/convert_quantizable_matmul.py
```diff
@@ -1,161 +0,0 @@
-# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
```
bfineran, line 1: did you `git add` this file after deleting? looks like git didn't pick up the rename/deletion

bfineran (approved):
LGTM pending comments in #1264 - see comment about git add as well

src/sparseml/exporters/transforms/convert_quantizable_matmul.py
corey-nm: It's because theres too many differences - I moved the function to a method of a different name and renamed the class

dbogunowicz (approved):

PR #1269, opened 2022-12-16
[ExporterRefactor] Standardizing XToMatMulIntegerAddCastMul transforms

corey-nm:
Mainly docstrings and logging, plus deferred node deletion


src/sparseml/exporters/transforms/gemm_to_matmulinteger_add_cast_mul.py
```diff
@@ -99,17 +92,18 @@ def transform(self, model: ModelProto) -> ModelProto:
             ):
                 # we do not currently handle Gemms with transposed A
                 # or scalar multiples
-                _LOGGER.debug(f"Skipping due to unsupported attributes {attr}")
                 continue
+            _LOGGER.debug(f"Matched {match}")
+            self._transform_match(graph, model, match, attr)
```
bfineran, line 97: the worry about not refreshing the graph on every transform is the internal dicts that track connections may become stale and lead to bugs - if you are confident this won't happen in this case should comment, otherwise let's revert in both cases here

building the `ONNXGraph` is relatively cheap (O(n)), what it speeds up the most is nested node/init name lookups (used to go 3-5 layers deep :|)

bfineran (approved):
LGTM pending comments in https://github.com/neuralmagic/sparseml/pull/1264 and taking a look at use of `ONNXGraph`

src/sparseml/exporters/transforms/gemm_to_matmulinteger_add_cast_mul.py
corey-nm: Yeah that's a fair point - should we also not deferred the node addition/removal for the same reason then?

corey-nm: Hmm i think the only concern i would have for these is that the helper function this calls (add_quantized_conv_matmul_add_ops) adds nodes/initializers to the graph. But the transforms shouldn't consider those new additions. Ideally they would do the deferred add on the nodes, but they don't have access to the transform (could make accept an ONNXTransform object?)

src/sparseml/exporters/transforms/matmul_to_matmulinteger_add_cast_mul.py
```diff
@@ -21,65 +21,61 @@
...
 
     def transform(self, model: ModelProto) -> ModelProto:
+        graph = ONNXGraph(model)
+        count = 0
         matches = get_structural_matches(
-            ONNXGraph(model),
+            graph,
             op_type="MatMul",
             parent_ops=[
                 ["DequantizeLinear"],
```
corey-nm, line 81: This should be `any_of("QuantizeLinear", "DequantizeLinear")`. also update docstring

src/sparseml/exporters/transforms/gemm_to_matmulinteger_add_cast_mul.py
```diff
@@ -44,36 +39,35 @@ class GemmToMatMulIntegerAddCastMul(OnnxTransform):
...
     """
 
     def transform(self, model: ModelProto) -> ModelProto:
+        graph = ONNXGraph(model)
         matches = get_structural_matches(
-            ONNXGraph(model),
+            graph,
             op_type="Gemm",
             parent_ops=[
                 ["DequantizeLinear"],
```
corey-nm, line 73: This should be `any_of("QuantizeLinear", "DequantizeLinear")` and update docstrings

dbogunowicz (approved):

PR #1272, opened 2022-12-19
Layer-Wise Distillation

rahul-tuli:
This PR represents the main branch for all layer-wise distillation work

### DIFF
--- a/src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
+++ b/src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
@@ -383,16 +383,12 @@ class PerLayerDistillationModifier(BaseDistillationModifier):
                 }
             )
 
-            # NOTE: since we added a param group, if someone tries
-            # to serialize the optimizer, the new param group will be
-            # included. if they try call load_state_dict() on
-            # a fresh optimizer (one that hasn't been called
-            # with compute_distillation_loss), it will
-            # crash.
             _old_state_dict_fn = optimizer.state_dict
             _old_load_state_dict_fn = optimizer.load_state_dict
 
             def state_dict_without_projection():
+                # Remove the param_group we added, as we save that in the modifier's
+                # state_dict.
                 state = _old_state_dict_fn()
                 state = deepcopy(state)
                 idx = _get_projection_param_group_idx(state["param_groups"])
@@ -512,6 +508,9 @@ def _find_layers_by_name(
 
 
 def _get_projection_param_group_idx(param_groups: List[Dict]) -> Optional[int]:
+    """
+    :return: Optional index where the param group is if it is found.
+    """
     for idx, group in enumerate(param_groups):
         if _DISTILL_PARAM_GROUP_KEY in group:
             return idx
--- a/src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
+++ b/src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
@@ -43,7 +43,6 @@ _LOGGER = logging.getLogger(__name__)
 _DISTILLATION_TYPES = [torch.nn.Conv2d, torch.nn.Linear]
 
 
-
 @PyTorchModifierYAML()
 class PerLayerDistillationModifier(BaseDistillationModifier):
     """
--- a/src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
+++ b/src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
@@ -245,10 +245,10 @@ class PerLayerDistillationModifier(BaseDistillationModifier):
             Defaults to 0 (start of the training process)
         :param loggers: Optional list of loggers to log the modification process to
         :param distillation_teacher: teacher module to perform knowledge distillation
-            with. If not provided, self distillation will be used with a teacher
-             from a copy of the given module at the start epoch. If given string
-             "disable" this modifier will not apply distillation of any kind,
-             even in the active epoch range
+            with. Pass `"self"` to use self distillation with a teacher
+            from a copy of the given module at the start epoch.
+            Pass `"disable"` to disable this modifier.
+            Defaults to `"disable"`
         :param kwargs: Optional kwargs to support specific arguments
             for individual modifiers.
         """
--- a/src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
+++ b/src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
@@ -47,9 +47,15 @@ class PerLayerDistillationModifier(BaseDistillationModifier):
     A distillation_teacher module may be provided as a kwarg to
     the Manager initialization and loss_update(loss) must be called before any
     backwards pass in the integrated training flow.
+
     If no teacher model is provided, then self-distillation will be used.
     The feature difference between teacher and student can be weighted spatially
     by a weighing function.
+
+    The connection between layers in the student model and layers in the teacher
+    model is controlled via the `teacher_layer_names` and `student_layer_names`
+    fields. **These fields must be the same length!** The i'th item in both
+    of these arrays are paired together.
 
     # Sample yaml:
 
@@ -57,26 +63,34 @@ class PerLayerDistillationModifier(BaseDistillationModifier):
     !PerLayerDistillationModifier
         gain: 2.0
         start_epoch: 0.0
+        end_epoch: 1.0
+        update_frequency: 0.2
+        normalize: True
+        student_layer_names: ["layer2.0.conv2"]
+        teacher_layer_names: ["layer2.0.conv2"]
         project_features: True
-        student_layer_names
+        epsilon: 1e-6
     ```
 
     # Parameters
 
+    :param gain: How much to weight the distillation loss.
     :param start_epoch: The epoch to start the modifier at
     :param end_epoch: The epoch to end the modifier at
-    :param distill_output_keys: List of keys for the module outputs to use for
-        distillation if multiple outputs are present. None or empty list defaults
-        to using all available outputs
-    :param teacher_input_keys: List of keys to filter the inputs by before
-        passing into the teacher. None or empty list defaults to using
-        all available inputs
-    :param update_frequency:
-    :param gain: How much to weight the distillation loss. Default is 1.5
+    :param update_frequency: The number of epochs or fraction of epochs to
+            update at between start and end
     :param normalize: Whether to normalize the output difference by the
-        the magnitude of the teacher's output
+        the magnitude of the teacher's output.
+        Default is `True`.
+    :param student_layer_names: List of layer names to distill.
+        *Must be same length as teacher_layer_names.*
+    :param teacher_layer_names: List of layer names to distill from.
+        *Must be same length as student_layer_names.*
+    :param project_features: Whether to project the output of student layers to the
+        same size as the output of the teacher layers.
+        Default is `True`
     :param epsilon: Small value used to avoid division by zero when normalization
-        is used. Default is 1.e-6
+        is used. Default is `1e-6`
     """
 
     def __init__(
@@ -88,8 +102,8 @@ class PerLayerDistillationModifier(BaseDistillationModifier):
         normalize: bool = True,
         student_layer_names: Optional[List[str]] = None,
         teacher_layer_names: Optional[List[str]] = None,
-        project_features: bool = False,
-        epsilon: float = 1.0e-6,
+        project_features: bool = True,
+        epsilon: float = 1e-6,
     ):
         if (
             student_layer_names is not None
@@ -113,6 +127,7 @@ class PerLayerDistillationModifier(BaseDistillationModifier):
         self._teacher_layer_names = teacher_layer_names
         self._project_features = project_features
         self._epsilon = epsilon
+
         self._cached_student_output: Dict[str, torch.Tensor] = {}
         self._cached_teacher_output: Dict[str, torch.Tensor] = {}
         self._student_handles: List[RemovableHandle] = []
--- a/src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
+++ b/src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
@@ -17,7 +17,7 @@ Modifier for performing knowledge distillation via feature imitation.
 """
 
 import logging
-from typing import Any, List, Optional, Union, Mapping
+from typing import List, Optional, Union
 
 import torch
 from torch.nn import Module
@@ -82,28 +82,26 @@ class PerLayerDistillationModifier(BaseDistillationModifier):
         gain: float,
         start_epoch: float = -1.0,
         end_epoch: float = -1.0,
-        distill_output_keys: Optional[List[Any]] = None,
-        teacher_input_keys: Optional[List[Any]] = None,
         update_frequency: float = -1.0,
         normalize: bool = True,
         student_layer_names: Optional[List[str]] = None,
         teacher_layer_names: Optional[List[str]] = None,
         project_features: bool = False,
-        epsilon: float = 1.e-6,
+        epsilon: float = 1.0e-6,
     ):
         super().__init__(
             start_epoch=start_epoch,
             end_epoch=end_epoch,
-            distill_output_keys=distill_output_keys,
-            teacher_input_keys=teacher_input_keys,
+            distill_output_keys=None,
+            teacher_input_keys=None,
             update_frequency=update_frequency,
         )
-        self.gain = gain
-        self.normalize = normalize
-        self.student_layer_names = student_layer_names
-        self.teacher_layer_names = teacher_layer_names
-        self.project_features = project_features
-        self.epsilon = epsilon
+        self._gain = gain
+        self._normalize = normalize
+        self._student_layer_names = student_layer_names
+        self._teacher_layer_names = teacher_layer_names
+        self._project_features = project_features
+        self._epsilon = epsilon
         self._cached_student_output = None
         self._cached_teacher_output = None
         self._student_handles = None
@@ -150,7 +148,7 @@ class PerLayerDistillationModifier(BaseDistillationModifier):
         """
         return self._epsilon
 
-    @normalize.setter
+    @epsilon.setter
     def epsilon(self, value: float):
         """
         :params value: small value to avoid division per zero when normalization is used
@@ -159,24 +157,19 @@ class PerLayerDistillationModifier(BaseDistillationModifier):
 
     @ModifierProp()
     def student_layer_names(self) -> List[str]:
-        return self._student_names
+        return self._student_layer_names
 
     @student_layer_names.setter
     def student_layer_names(self, value: List[str]):
-        self._student_names = value
+        self._student_layer_names = value
 
     @ModifierProp()
     def teacher_layer_names(self) -> List[str]:
-        if self._teacher_names is not None:
-            return self._teacher_names
-        elif self.student_layer_names is not None:
-            return self.student_layer_names
-        else:
-            return None
+        return self._teacher_layer_names
 
     @teacher_layer_names.setter
     def teacher_layer_names(self, value: List[str]):
-        self._teacher_names = value
+        self._teacher_layer_names = value
 
     @ModifierProp()
     def project_features(self) -> bool:
@@ -226,22 +219,36 @@ class PerLayerDistillationModifier(BaseDistillationModifier):
                 self._student_names = list(cached_student_layers.keys())
                 self._teacher_names = list(cached_teacher_layers.keys())
             else:
-                _find_layers_by_name(module, self.student_layer_names, cached_student_layers)
-                _find_layers_by_name(distillation_teacher, self.teacher_layer_names, cached_teacher_layers)
+                _find_layers_by_name(
+                    module, self.student_layer_names, cached_student_layers
+                )
+                _find_layers_by_name(
+                    distillation_teacher,
+                    self.teacher_layer_names,
+                    cached_teacher_layers,
+                )
 
             self._student_handles = []
             self._teacher_handles = []
             for layer_name in cached_student_layers:
                 self._student_handles.append(
                     cached_student_layers[layer_name].register_forward_hook(
-                        _create_cache_output_hook(layer_name, self._cached_student_output, self._student_output_shapes)
+                        _create_cache_output_hook(
+                            layer_name,
+                            self._cached_student_output,
+                            self._student_output_shapes,
+                        )
                     )
                 )
 
             for layer_name in cached_teacher_layers:
                 self._student_handles.append(
                     cached_teacher_layers[layer_name].register_forward_hook(
-                        _create_cache_output_hook(layer_name, self._cached_teacher_output, self._teacher_output_shapes)
+                        _create_cache_output_hook(
+                            layer_name,
+                            self._cached_teacher_output,
+                            self._teacher_output_shapes,
+                        )
                     )
                 )
             self._teacher = distillation_teacher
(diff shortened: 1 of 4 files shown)


src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -0,0 +1,368 @@
...
+    the Manager initialization and loss_update(loss) must be called before any
+    backwards pass in the integrated training flow.
+    If no teacher model is provided, then self-distillation will be used.
+    The feature difference between teacher and student can be weighted spatially
+    by a weighing function.
+    | Sample yaml:
+    |   !PerLayerDistillationModifier
+    |       start_epoch: 0.0
+    |       gain: 2.0
+    |       number_of_classes: 80
```
corey-nm, line 61:
```suggestion
```

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -0,0 +1,368 @@
...
+    |       student_features: [64, 128, 256]
+    |       teacher_features: [128, 256, 512]
```
corey-nm, line 63: Remove these and add layer names to this example?

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -0,0 +1,368 @@
...
+    :param start_epoch: The epoch to start the modifier at
+    :param end_epoch: The epoch to end the modifier at
+    :param distill_output_keys: List of keys for the module outputs to use for
+        distillation if multiple outputs are present. None or empty list defaults
+        to using all available outputs
+    :param teacher_input_keys: List of keys to filter the inputs by before
+        passing into the teacher. None or empty list defaults to using
+        all available inputs
+    :param update_frequency:
```
corey-nm, line 72: document this

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -0,0 +1,368 @@
...
+    :param gain: How much to weight the distillation loss. Default is 1.5
+    :param normalize: Whether to normalize the output difference by the
+        the magnitude of the teacher's output
+    :param epsilon: Small value used to avoid division by zero when normalization
+        is used. Default is 1.e-6
+    """
```
corey-nm, line 78: need to add:
- project_features
- student_layer_names
- teacher_layer_names

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -0,0 +1,368 @@
...
+        self._normalize = value
+
+    @ModifierProp()
+    def epsilon(self) -> float:
+        """
+        :return: small value to avoid division per zero when normalization is used
+        """
+        return self._epsilon
+
+    @normalize.setter
```
corey-nm, line 153:
```suggestion
    @epsilon.setter
```

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -0,0 +1,368 @@
...
+    def epsilon(self, value: float):
+        """
+        :params value: small value to avoid division per zero when normalization is used
+        """
+        self._epsilon = value
+
+    @ModifierProp()
+    def student_layer_names(self) -> List[str]:
+        return self._student_names
```
corey-nm, line 162:
```suggestion
        return self._student_layer_names
```

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -0,0 +1,368 @@
...
+
+    @student_layer_names.setter
+    def student_layer_names(self, value: List[str]):
+        self._student_names = value
```
corey-nm, line 166:
```suggestion
        self._student_layer_names = value
```

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -0,0 +1,368 @@
...
+
+    @ModifierProp()
+    def teacher_layer_names(self) -> List[str]:
+        if self._teacher_names is not None:
+            return self._teacher_names
+        elif self.student_layer_names is not None:
+            return self.student_layer_names
+        else:
+            return None
```
corey-nm, line 175: Should probably just have this be `return self._teacher_layer_names`

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -0,0 +1,368 @@
...
+        :param reset_loggers: True to remove any currently attached loggers (default),
+            False to keep the loggers attached.
+        :param kwargs: Optional kwargs to support specific arguments
+            for individual modifiers.
+        """
+        super().finalize(module, reset_loggers, **kwargs)
+        for handle in self._student_handles:
+            handle.remove()
+        for handle in self._teacher_handles:
+            handle.remove()
```
corey-nm, line 272: Should be removed after self.end_epoch. Unclear what callback this will be in

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -0,0 +1,368 @@
...
+__all__ = [
+    "PerLayerDistillationModifier",
+]
+
+_LOGGER = logging.getLogger(__name__)
+
+_DISTILLATION_TYPES = [
+    torch.nn.Conv1d,
+    torch.nn.Conv2d,
+    torch.nn.Conv3d,
```
corey-nm, line 42: we only actually support Conv2d i believe (the other two don't have 4d input sizes so the projections won't work).

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -0,0 +1,368 @@
...
+        self._student_handles = None
+        self._teacher_handles = None
+        self._cached_student_output = None
+        self._cached_teacher_output = None
+
+    def compute_distillation_loss(self, optimizer, **kwargs):
+        distillation_loss = 0.0
+
+        if self.project_features and self._projection is None:
+            self._initialize_projection()
```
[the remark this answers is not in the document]
corey-nm: (in addition to the 4 lines of code below this, which can be moved into self._initialize_projection)

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
corey-nm: @bfineran what callback should this be moved to?

tests/sparseml/pytorch/sparsification/distillation/test_per_layer_distillation.py
```diff
@@ -0,0 +1,202 @@
...
+
+from tests.sparseml.pytorch.helpers import (  # noqa isort:skip
+    test_epoch,
+    test_loss,
+    test_steps_per_epoch,
+)
+
+
+DISTILLATION_MODIFIERS = [
+    lambda: DistillationModifier(start_epoch=0.0),
```
corey-nm, line 41: Probably want:
1. `PerLayerDistillationModifier` with no projections
2. `PerLayerDistillationModifier` with mixed projections (some layers projected, some not)

tests/sparseml/pytorch/sparsification/distillation/test_per_layer_distillation.py
```diff
@@ -0,0 +1,202 @@
...
+    return torch.randn(batch_size, *input_shape)
+
+
+@pytest.mark.skipif(
+    os.getenv("NM_ML_SKIP_PYTORCH_TESTS", False),
+    reason="Skipping pytorch tests",
+)
+@pytest.mark.parametrize("modifier_lambda", DISTILLATION_MODIFIERS, scope="function")
+@pytest.mark.parametrize("model_lambda", [LinearNet], scope="function")
+@pytest.mark.parametrize("optim_lambda", [create_optim_sgd], scope="function")
```
corey-nm, line 57: we should add another parametrize for teacher_lambda, which will be different from model_lambda in some test cases

tests/sparseml/pytorch/sparsification/distillation/test_per_layer_distillation.py
```diff
@@ -0,0 +1,202 @@
...
+        modifier_lambda,
+        model_lambda,
+        optim_lambda,
+        test_steps_per_epoch,  # noqa: F811
+    ):
+        modifier = modifier_lambda()
+        model = model_lambda()
+        optimizer = optim_lambda(model)
+
+        self.initialize_helper(modifier, model, distillation_teacher=model_lambda())
```
corey-nm, line 104: disillation_teacher should be the newly added teacher_lambda()

tests/sparseml/pytorch/sparsification/distillation/test_per_layer_distillation.py
```diff
@@ -0,0 +1,202 @@
...
+        assert isinstance(updated_loss, torch.Tensor)
+        assert updated_loss.shape == fake_loss.shape
+        assert fake_loss.item() != updated_loss.item()
+
+
+@pytest.mark.skipif(
+    os.getenv("NM_ML_SKIP_PYTORCH_TESTS", False),
+    reason="Skipping pytorch tests",
+)
+def test_distillation_modifier_yaml():
```
corey-nm, line 159: I think for this test we just need to:
1. Change the name of the class being used to the per layer one
2. Add all the possible fields

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -0,0 +1,419 @@
...
+        optimizer: torch.optim.Optimizer,
+        epoch: float,
+        steps_per_epoch: int,
+    ):
+        super().update(module, optimizer, epoch, steps_per_epoch)
+        if epoch >= self.end_epoch:
+            for handle in self._student_handles:
+                handle.remove()
+            for handle in self._teacher_handles:
+                handle.remove()
```
corey-nm, line 309: @bfineran is this what you meant by removing them after end epoch? will this work?

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -0,0 +1,432 @@
...
+
+    def compute_total_loss(self, loss, distillation_loss):
+        return loss + self.gain * distillation_loss
+
+    def compute_distillation_loss(self, optimizer: torch.optim.Optimizer, **kwargs):
+        if self.project_features and len(self._projection) == 0:
+            # NOTE: have to call initialize here because we need the cached output
+            # from the module. i.e. we need forward to have been called already
+            projection_params = self._initialize_projection()
+            optimizer.add_param_group({"params": projection_params})
```
corey-nm, line 341: noting that params is expected by pytorch

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -0,0 +1,442 @@
...
+            # crash.
+            _old_state_dict_fn = optimizer.state_dict
+
+            def state_dict_without_latest_param_group():
+                full_state_dict = _old_state_dict_fn()
+                original_state_dict = deepcopy(full_state_dict)
+                original_state_dict["param_groups"].pop()
+                return original_state_dict
+
+            optimizer.state_dict = state_dict_without_latest_param_group
```
corey-nm, line 348: @bfineran this is the only way we could think of getting around the serialization issue. feels very hacky though and we aren't sure how safe this is

KSGulin (commented):
Looks great. Just a few minor comments

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -0,0 +1,442 @@
...
+        start_epoch: float = -1.0,
+        end_epoch: float = -1.0,
+        update_frequency: float = -1.0,
+        normalize: bool = True,
+        student_layer_names: Optional[List[str]] = None,
+        teacher_layer_names: Optional[List[str]] = None,
+        project_features: bool = True,
+        epsilon: float = 1e-6,
+    ):
+        if (
```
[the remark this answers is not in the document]
corey-nm: I think that is technically allowed since in that case both would just default to all layers (and then there's a value error raised in initialize if they have different number of layers)

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -0,0 +1,442 @@
...
+             even in the active epoch range
+        :param kwargs: Optional kwargs to support specific arguments
+            for individual modifiers.
+        """
+        super().initialize(module, epoch, loggers, distillation_teacher, **kwargs)
+
+        if not self._distillation_enabled:
+            return
+
+        if distillation_teacher == "self":
```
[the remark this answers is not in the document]
corey-nm: Oh yeah docstring is out of date. It defaults to disable in base class, so I think we should keep it at that

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -0,0 +1,442 @@
...
+        Store the teacher model for distillation if provided
+        :param module: the PyTorch model/module to modify
+        :param epoch: The epoch to initialize the modifier and module at.
+            Defaults to 0 (start of the training process)
+        :param loggers: Optional list of loggers to log the modification process to
+        :param distillation_teacher: teacher module to perform knowledge distillation
+            with. If not provided, self distillation will be used with a teacher
+             from a copy of the given module at the start epoch. If given string
+             "disable" this modifier will not apply distillation of any kind,
+             even in the active epoch range
```
corey-nm, line 251:
```suggestion
        :param distillation_teacher: teacher module to perform knowledge distillation
            with. Pass `"self"` to use self distillation with a teacher
            from a copy of the given module at the start epoch.
            Pass `"disable"` to disable this modifier.
```

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
bfineran: what do you mean by this? is this a reserved name? why not name the group `distill_projection_params`?

tests/sparseml/pytorch/sparsification/distillation/test_per_layer_distillation.py
```diff
@@ -0,0 +1,410 @@
...
+def test_hooks_survive_after_quantization_is_applied():
+    # this should test fused module hooks are properly kept around during
+    # quantization
+    modifier = PerLayerDistillationModifier()
+    student = resnet18(pretrained=True)
+    teacher = resnet18(pretrained=True)
+    opt = create_optim_sgd(student)
+
+    modifier.initialize(student, distillation_teacher=teacher)
+
```
bfineran, line 391: let's run a fake forward pass here before the quantization modifier is added as well for completeness since additional logic may happen here

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -0,0 +1,442 @@
...
+
+    @epsilon.setter
+    def epsilon(self, value: float):
+        """
+        :params value: small value to avoid division per zero when normalization is used
+        """
+        self._epsilon = value
+
+    @ModifierProp()
+    def student_layer_names(self) -> List[str]:
```
bfineran, line 210: nit: this fn and the ones below are missing docstrings

tests/sparseml/pytorch/sparsification/distillation/test_per_layer_distillation.py
```diff
@@ -0,0 +1,410 @@
...
+    modifier.loss_update(
+        loss=fake_loss,
+        module=student,
+        optimizer=optimizer,
+        epoch=modifier.start_epoch,
+        steps_per_epoch=10,
+        student_inputs=x,
+        student_outputs=fake_loss,
+    )
+
```
bfineran, line 367: let's add some sanity checks that the projection param group exists

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
bfineran: I think we're going to need to override the state dict function unfortunately, don't see any way around that given this workflow. this however doesn't address the issue of reloading see my comment on reloading for that.

bfineran: additionally, on first glance, I'm not sure this properly overrides the state dict function since the method is not bound to the optimizer instance. (ie when `optimizer.state_dict()` is called, `self` will be implicitly passed in, however the function we override with doesn't include this) the test seems to pass for some reason though so maybe I am wrong here.
if we do need to bind,  I can give a quick code pointer on that

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -0,0 +1,442 @@
...
+        return loss + self.gain * distillation_loss
+
+    def compute_distillation_loss(self, optimizer: torch.optim.Optimizer, **kwargs):
+        if self.project_features and len(self._projection) == 0:
+            # NOTE: have to call initialize here because we need the cached output
+            # from the module. i.e. we need forward to have been called already
+            projection_params = self._initialize_projection()
+            optimizer.add_param_group({"params": projection_params})
+
+            # NOTE: since we added a param group, if someone tries
```
bfineran, line 334: let's add support for re-serialization, so information is not lost.

a few notes

`state_dict()`
1. we'll serialize everything to the **modifier** state dict
2. we'll include all the projection weights (projection module name -> `projection.state_dict()`
3. we'll include the optimizer param group dict

`load_state_dict()`
1. if projections have not been created yet (this is the likely case) - we should just save the info as a private property of the modifier and then load in the weights for the projections/optimizer as soon as the projections are added
2. if the projections have been created, we can load directly in

@corey-nm @rahul-tuli  let me know if this is clear. can go over offline if needed

bfineran (changes requested):
Looks great @rahul-tuli @corey-nm - few small comments and need to add that small change for serialization then LGTM!

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
corey-nm: @bfineran params is the keyword that pytorch requires you to use (all param groups is a dict of keys like "params", "lr", etc)

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
rahul-tuli: @bfineran kindly check, the new implementation is up

tests/sparseml/pytorch/sparsification/distillation/test_per_layer_distillation.py
rahul-tuli: Done!

tests/sparseml/pytorch/sparsification/distillation/test_per_layer_distillation.py
rahul-tuli: Done!

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -208,28 +211,64 @@ def epsilon(self, value: float):
...
+        if self.project_features:
+            state[_DISTILL_PARAM_GROUP_KEY] = {
+                name: p.state_dict() for name, p in self._projection.items()
+            }
+        return state
+
+    def load_state_dict(self, state_dict: Dict[str, Dict], strict: bool = True):
+        if self.project_features:
+            # save until self._projection is actually initialized after forward
+            self._loaded_projection = state_dict.pop(_DISTILL_PARAM_GROUP_KEY)
```
bfineran, line 269: why are we popping? this would modify the passed state dict in-place, no?

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -208,28 +211,64 @@ def epsilon(self, value: float):
...
     def project_features(self, value: bool):
+        """
+        :param value: Whether to perform projection
+            between student and teacher layers
+        """
         self._project_features = value
 
+    def state_dict(self) -> Dict[str, Dict]:
+        state = super().state_dict()
+        if self.project_features:
```
bfineran, line 260: we also need to save the optimizer group state dict here as well so they may be reloaded, right now, they're just left out from serialization

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -40,6 +40,8 @@
 
 _DISTILLATION_TYPES = [torch.nn.Conv2d, torch.nn.Linear]
 
+_DISTILL_PARAM_GROUP_KEY = "distillation_projection_params"
```
bfineran, line 43: let's rename to `_DISTILL_PARAM_GROUP_KEY` and add to `__all__`

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -328,24 +368,51 @@ def compute_distillation_loss(self, optimizer: torch.optim.Optimizer, **kwargs):
...
+        if _get_projection_param_group_idx(optimizer.param_groups) is None:
+            optimizer.add_param_group(
+                {
+                    _DISTILL_PARAM_GROUP_KEY: True,
+                    "params": [p.weight for p in self._projection.values()],
+                }
+            )
 
-            optimizer.state_dict = state_dict_without_latest_param_group
+            _old_state_dict_fn = optimizer.state_dict
```
bfineran, line 386: let's move this block to a helper function

bfineran (changes requested):
new state dict logic looks much better - LGTM pending comments

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
corey-nm: that is what is saved in this block - the optimizer's extra group is made up of the self._projection state dict

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
corey-nm: to handle cases where the key is not present, we could use `.get` instead to avoid mutating

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
rahul-tuli: Don't wanna add a "protected" attribute to `__all__` will remove leaading underscore if this needs to be added

bfineran: yes that was the intention, i forgot to remove the `_` after copy paste 🤦

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
bfineran: let's do that

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
rahul-tuli: Done!

bfineran (dismissed):
great work @rahul-tuli  @corey-nm

corey-nm (approved):
woohoo!

bfineran (approved):

PR #1274, opened 2022-12-19
Properly pass update_frequency in modifier_distillation_base.py

corey-nm:
This parameter is unused currently


bfineran (approved):

bfineran (commented):
Actually let's check if this is intentional for distillation modifier since updates should happen on every step. If so should update docstring

corey-nm:
@bfineran you are correct, it seems like this is intentional

PR #1280, opened 2022-12-21
Saving all hooks during quantization block fusing

corey-nm:
Currently only the following hooks are transferred during quantization fusing process:
1. the pre forward hooks for the first layer
2. the post forward hooks for the last layer

Notably, the following hooks are lost:
1. The post forward hooks in the first layer
2. The pre forward hooks in the last layer
3. Any hooks from any layer between first and last layer

This PR changes it so all pre/post hooks from every layer in the fused block is transferred to the new fused module.

# Testing plan

Existing unit tests, and also this was tested for the #1272 PR

### DIFF
--- a/src/sparseml/pytorch/sparsification/quantization/helpers.py
+++ b/src/sparseml/pytorch/sparsification/quantization/helpers.py
@@ -704,24 +704,24 @@ def prepare_embeddings_qat(
 
 def _delete_get_block_hooks(
     module: Module,
-    fuse_blocks: List[str],
+    fuse_blocks: List[List[str]],
 ) -> List[Tuple[Any, Any]]:
     block_hooks = []
     for block in fuse_blocks:
         pre_hooks = []
         post_hooks = []
 
-        # get first and last Module objects in block by their names
         for name in block:
+            # get Module objects in block by their names
             m = get_layer(name, module)
 
-            for handle_id, pre_hook_fn in list(m._forward_pre_hooks.items()):
-                pre_hooks.append(pre_hook_fn)
-                del m._forward_pre_hooks[handle_id]
+            # extract the hooks
+            pre_hooks.extend(m._forward_pre_hooks.values())
+            post_hooks.extend(m._forward_hooks.values())
 
-            for handle_id, hook_fn in list(m._forward_hooks.items()):
-                post_hooks.append(hook_fn)
-                del m._forward_hooks[handle_id]
+            # de-register the hooks from this module
+            m._forward_pre_hooks.clear()
+            m._forward_hooks.clear()
 
         block_hooks.append((pre_hooks, post_hooks))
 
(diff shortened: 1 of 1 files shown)


src/sparseml/pytorch/sparsification/quantization/helpers.py
```diff
@@ -712,16 +712,16 @@ def _delete_get_block_hooks(
         post_hooks = []
 
         # get first and last Module objects in block by their names
```
natuan, line 714: This comment is no longer correct

src/sparseml/pytorch/sparsification/quantization/helpers.py
```diff
@@ -712,16 +712,16 @@ def _delete_get_block_hooks(
...
-        block_head = get_layer(block[0], module)
-        block_tail = get_layer(block[-1], module)
+        for name in block:
+            m = get_layer(name, module)
 
-        for handle_id, pre_hook_fn in list(block_head._forward_pre_hooks.items()):
-            pre_hooks.append(pre_hook_fn)
-            del block_head._forward_pre_hooks[handle_id]
+            for handle_id, pre_hook_fn in list(m._forward_pre_hooks.items()):
```
natuan, line 718: Wonder if you'd mind reduce the duplicate pattern from these two for loops (make it easier to read)?

natuan (changes requested):

src/sparseml/pytorch/sparsification/quantization/helpers.py
corey-nm: Oh yeah good catch, updated!

src/sparseml/pytorch/sparsification/quantization/helpers.py
corey-nm: Updated - is the new logic easier to read?

natuan: Absolute. Thanks!

bfineran (approved):
tentatively approving as for our current use cases this should be safe (especially since activation shapes/types should not be changed across the kinds of blocks we fuse). but this could potentially cause misuse of certain hooks. maybe we can add an in code comment about this potential.

(ie imagine there is a hook specifically to track the output of a batch norm, now that hook will be tracking the output of the relu - we don't do anything like this but potentially opens this issue up)

natuan:
> tentatively approving as for our current use cases this should be safe (especially since activation shapes/types should not be changed across the kinds of blocks we fuse). but this could potentially cause misuse of certain hooks. maybe we can add an in code comment about this potential.
> 
> (ie imagine there is a hook specifically to track the output of a batch norm, now that hook will be tracking the output of the relu - we don't do anything like this but potentially opens this issue up)

@bfineran just to understand your point: is it right that if there's a hook as in your bn example, the current code (before this fix) we would loose it after fusing, correct?

natuan (approved):

Issue #1282, opened 2022-12-22
Sparsity scheduler always follows default interpolation for GMP


eldarkurtic:
**Describe the bug**
SparseML pruning modifiers support three different interpolation functions for sparsity scheduler:
```python
    :param inter_func: the type of interpolation function to use:
        [linear, cubic, inverse_cubic]
```
but GMPruningModifier always defaults to `cubic`.
A recipe like this:
```yaml
    - !GMPruningModifier
        inter_func: linear                  <--- or `inverse_cubic`
        start_epoch: 0
        end_epoch: 40
        update_frequency: 1.0
        init_sparsity: 0.05
        final_sparsity: 0.85
        mask_type: unstructured
        params: ['sections.0.0.conv1.weight', 'sections.0.0.conv2.weight', 'sections.0.0.conv3.weight']
```
will still use `cubic` interpolation for the sparsity scheduler.

**Expected behavior**
To have GMP modifier using the correct interpolation function from the given yaml recipe.

dbogunowicz:
Hey Eldar, I will take a look into it. Thanks for surfacing it.

eldarkurtic:
@dbogunowicz I have already submitted a fix for it here: https://github.com/neuralmagic/sparseml/pull/1283

PR #1284, opened 2022-12-22
Log overall sparsity of the model

eldarkurtic:
Issues like https://github.com/neuralmagic/sparseml/issues/1282 are very hard to detect when `global_sparsity: True`, as the final model will have the desired target sparsity but the sparsity scheduler might have followed the wrong interpolation function. This PR adds logging of the overall sparsity of the model, which makes this issue easily detectable.

### DIFF
--- a/src/sparseml/pytorch/sparsification/pruning/modifier_pruning_base.py
+++ b/src/sparseml/pytorch/sparsification/pruning/modifier_pruning_base.py
@@ -432,7 +432,7 @@ class BasePruningModifier(ABC, ScheduledUpdateModifier):
                     steps_per_epoch=steps_per_epoch,
                 )
         self.log_scalar(
-            tag="Overall sparsity",
+            tag=f"modifier_sparsity/{self.__class__.__name__}",
             value=num_zeros / (num_params + torch.finfo(torch.float32).eps),
             epoch=epoch,
             steps_per_epoch=steps_per_epoch,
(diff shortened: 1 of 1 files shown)


src/sparseml/pytorch/sparsification/pruning/modifier_pruning_base.py
```diff
@@ -423,6 +431,12 @@ def log_update(
                     epoch=epoch,
                     steps_per_epoch=steps_per_epoch,
                 )
+        self.log_scalar(
+            tag="Overall sparsity",
```
bfineran, line 435: nit: prefer `_` separated

bfineran (commented):
looks good overall, however maybe we should only log when global sparsity is set to True (and then rename tag to `global_sparsity`. thinking about cases where users may have multiple pruning modifiers for different parameters we will see multiple "Overall Sparsities" logged.

eldarkurtic:
@bfineran that's a very good point for multiple modifiers, I've completely missed it.

Regarding logging only when `global_sparsity` is True:
- from practical point of view, I think it would be nice to have all runs logging the same things so it is easier to compare them visually in e.g. wandb dashboard
-  even if we log only when `global_sparsity` is True, multiple modifiers issue can happen because SparseML notion of `global_sparsity` doesn't mean global for the entire model (implying there will be only one modifier); it means the sparsity profile is global only for the given `params` in the specific PruningModifier. There can be multiple global-pruners which are pruning different subsets of `params` with `global_sparsity: True`.
- the latest commit attempts to solve this by logging sparsity with the name of the respective PruningModifier class

jeanniefinks:

PR #1288, opened 2022-12-28
[onnx.utils] fold_conv_bns in-memory ModelProto support

bfineran:
simple change to allow loaded onnx models to be passed directly to `fold_conv_bns`.
this is required for support in on-going ONNX one-shot work

**test_plan:**
manual with related test coverage in separate repo


bfineran:
force landing given the limited scope, test coverage in separate NM repo, and that it is blocking

PR #1312, opened 2023-01-11
Torchvision scaler

corey-nm:
Properly passes scaler to wrap_optim.

# Test plan

Ran
- without `--amp`
- with `--amp`

**Have not tested amp with quantization**

### DIFF
--- a/src/sparseml/pytorch/torchvision/train.py
+++ b/src/sparseml/pytorch/torchvision/train.py
@@ -562,7 +562,11 @@ def main(args):
             distillation_teacher=args.distill_teacher,
         )
         optimizer = manager.modify(
-            model, optimizer, steps_per_epoch=steps_per_epoch, epoch=args.start_epoch
+            model,
+            optimizer,
+            steps_per_epoch=steps_per_epoch,
+            epoch=args.start_epoch,
+            wrap_optim=scaler,
         )
 
     lr_scheduler = _get_lr_scheduler(
--- a/src/sparseml/pytorch/torchvision/train.py
+++ b/src/sparseml/pytorch/torchvision/train.py
@@ -592,7 +592,7 @@ def main(args):
         if args.distributed:
             train_sampler.set_epoch(epoch)
         if manager is not None and manager.qat_active(epoch=epoch):
-            scaler = None
+            scaler._enabled = False
             model_ema = None
 
         train_metrics = train_one_epoch(
--- a/src/sparseml/pytorch/torchvision/train.py
+++ b/src/sparseml/pytorch/torchvision/train.py
@@ -561,13 +561,17 @@ def main(args):
             loggers=logger,
             distillation_teacher=args.distill_teacher,
         )
-        optimizer = manager.modify(
+        step_wrapper = manager.modify(
             model,
             optimizer,
             steps_per_epoch=steps_per_epoch,
             epoch=args.start_epoch,
             wrap_optim=scaler,
         )
+        if scaler is None:
+            optimizer = step_wrapper
+        else:
+            scaler = step_wrapper
 
     lr_scheduler = _get_lr_scheduler(
         args, optimizer, checkpoint=checkpoint, manager=manager


src/sparseml/pytorch/torchvision/train.py
```diff
@@ -561,9 +561,17 @@ def log_metrics(tag: str, metrics: utils.MetricLogger, epoch: int, epoch_step: i
...
+            model,
+            optimizer,
+            steps_per_epoch=steps_per_epoch,
+            epoch=args.start_epoch,
+            wrap_optim=scaler,
         )
+        if scaler is None:
+            optimizer = step_wrapper
+        else:
+            scaler = step_wrapper
```
corey-nm, line 574: Will re-ask this question here:

If using quantization & amp, the scaler will be set to None down below. This means that train_one_epoch will directly using optimizer instead of scaler. However the manager.modify wrapped the scaler.

Does that mean I need to re-call manager.modify on the optimizer after amp is disabled?

corey-nm: Or should I disable scaler by doing `scaler._enabled = False`? That seems to be what transformers integration does

corey-nm: Resolved in meeting - setting enabled flag for scaler down below

KSGulin: As mentioned offline, just setting `scaler._enabled = False` is enough. The optimizer step will be taken care of by torch here:
https://github.com/pytorch/pytorch/blob/d8e795ecd53670682bd3b2e5ff1f378402b147d5/torch/cuda/amp/grad_scaler.py#L313

KSGulin (approved):

bfineran (approved):
nice

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -568,9 +568,17 @@ def log_metrics(tag: str, metrics: utils.MetricLogger, epoch: int, epoch_step: i
             loggers=logger,
             distillation_teacher=args.distill_teacher,
         )
-        optimizer = manager.modify(
-            model, optimizer, steps_per_epoch=steps_per_epoch, epoch=args.start_epoch
+        step_wrapper = manager.modify(
```
natuan, line 571: @corey-nm We should pass the loggers into here so that from now on logging could be handled both by the modifiers (relevant for e.g. logging different quantization losses) and by the loss update of the wrapped manager (i.e. the RecipeManagerStepWrapper). It also prepares us well for distillation in the future.

I think we also need to recompute the steps per epoch passed in, taking into account grad accum steps and multi gpus.  Look at this line and those right above: https://github.com/neuralmagic/sparseml/blob/main/src/sparseml/transformers/sparsification/trainer.py#L272

After that, we can remove all the logging from this script (except perhaps for the python logger).

natuan (changes requested):

src/sparseml/pytorch/torchvision/train.py
natuan: Btw, looks like we can unify scaler and optimizer into "step_wrapper" and just pass it into the `train_one_epoch`(instead of having 2 separate/mutually exclusive objects)?

corey-nm: 1. the loggers were passed into manager.initializer() in the call right above this - does that not pass the loggers to all the modifiers?
2. steps_per_epoch does take into account gradient accumulation, but not multi gpu. i can fix the multi gpu in a separate pr
3. yeah the scaler/optimizer interactions are weird. you need to pass optimizer to scaler.step, and you also need to call optimizer.zero_grad() through the optimizer. it seems like if they were designed differently they could be combined. we could create a new class that wraps them, but i'm not sure the indirection it would add would be worth it. thoughts?

natuan: I see, you're right about the loggers init. 
Thanks for taking on 2. For 3, no I didn't mean to push for it unless it's an easy clean up.

natuan: I think let's push this in and continue to improve it

natuan (approved):

PR #1317, opened 2023-01-12
[Layer Wise Distillation] Minor fixes

KSGulin:
This PR adds fixes a couple bugs discovered during implementation of distillation for yolov5.

Fixes
- Don't enforce type checking for the final model output when using PerLayerDistillation modifier. Or generally, allow said override for any distillation modifier. This is added to account for instances where a model will produce different output in eval and train mode, such as yolov5, and cause the final output between the student and teacher to not align.
- Don't clear handles immediately when no end epoch is set

Test Plan
Covered by existing unit tests

### DIFF
--- a/src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
+++ b/src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
@@ -155,8 +155,6 @@ class PerLayerDistillationModifier(BaseDistillationModifier):
         self._teacher_output_shapes: Dict[str, torch.Size] = {}
         self._loaded_projection = None
 
-        self._compare_model_output = False
-
     def _reset_cache(self):
         self._cached_student_output.clear()
         self._cached_teacher_output.clear()
--- a/src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
+++ b/src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
@@ -155,6 +155,8 @@ class PerLayerDistillationModifier(BaseDistillationModifier):
         self._teacher_output_shapes: Dict[str, torch.Size] = {}
         self._loaded_projection = None
 
+        self._compare_model_output = False
+
     def _reset_cache(self):
         self._cached_student_output.clear()
         self._cached_teacher_output.clear()
@@ -354,7 +356,7 @@ class PerLayerDistillationModifier(BaseDistillationModifier):
         steps_per_epoch: int,
     ):
         super().update(module, optimizer, epoch, steps_per_epoch)
-        if epoch >= self.end_epoch:
+        if self.end_epoch > 0 and epoch >= self.end_epoch:
             for handle in self._student_handles:
                 handle.remove()
             for handle in self._teacher_handles:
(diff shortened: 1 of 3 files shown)


src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -354,7 +356,7 @@ def update(
         steps_per_epoch: int,
     ):
         super().update(module, optimizer, epoch, steps_per_epoch)
-        if epoch >= self.end_epoch:
+        if self.end_epoch > 0 and epoch >= self.end_epoch:
```
rahul-tuli, line 359: nit: maybe this reads better?
```suggestion
        if 0 < self.end_epoch <= epoch:
```

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
```diff
@@ -155,6 +155,8 @@ def __init__(
         self._teacher_output_shapes: Dict[str, torch.Size] = {}
         self._loaded_projection = None
 
+        self._compare_model_output = False
```
rahul-tuli, line 158: Could we set up a way to override this? If my understanding is correct this would not check the types for  all usages of `PerLayerDistillationModifier`? Does my argument make sense?

rahul-tuli (dismissed):

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
KSGulin: Yep I do like that better

corey-nm:
> This is added to account for instances where a model will produce different output in eval and train mode, such as yolov5, and cause the final output between the student and teacher to not align.

Can you expand on this? What are the types/values produced in eval vs train?

src/sparseml/pytorch/sparsification/distillation/modifier_per_layer.py
KSGulin: It only skips the type check for the final model output, not for the layers in between. This final output actually goes unused by `PerLayerDistillationModifier`

KSGulin:
> > This is added to account for instances where a model will produce different output in eval and train mode, such as yolov5, and cause the final output between the student and teacher to not align.
> 
> Can you expand on this? What are the types/values produced in eval vs train?

In train mode it returns only train outputs (a list) and in eval mode in returns a tuple of inference outputs and train outputs together. Train outputs are the pre-sigmoid values to be used for the loss calculation and the inference outputs are the post-processed values ready for NMS

corey-nm:
> In train mode it returns only train outputs (a list) and in eval mode in returns a tuple of inference outputs and train outputs together. Train outputs are the pre-sigmoid values to be used for the loss calculation and the inference outputs are the post-processed values ready for NMS

Can we just ensure that the teacher/student are both in the same mode?

KSGulin:
> > In train mode it returns only train outputs (a list) and in eval mode in returns a tuple of inference outputs and train outputs together. Train outputs are the pre-sigmoid values to be used for the loss calculation and the inference outputs are the post-processed values ready for NMS
> 
> Can we just ensure that the teacher/student are both in the same mode?

We want to keep the teacher in train mode so that all layers (e.g. batchnorm) behave as expected.

Also, as noted in my response to @rahul-tuli, the teacher output being type checked here isn't used by `PerLayerDistillationModifier`. I think outside of this specific yolov5 instance, it may not be a good idea to type check an unrelated output

corey-nm (approved):
🚀

bfineran (approved):

PR #1318, opened 2023-01-12
Adding config to wandb logger and calling LoggerManager.save(recipe)

corey-nm:
1. If args.recipe is a local file path, it will appear under the files section of wandb:
image

2. In the overview tab you can now view all command line parameters in the config table

image


bfineran (approved):
nice

abhinavnmagic (approved):

PR #1326, opened 2023-01-17
Update RN50 for new quantization modifier

anmarques:
This PR fixes a long-standing issue w/ quantization of RN50.

For performance purposes the inputs to the residual add operation must be quantized. In the old flow this was achieved by allowing the output of conv modules to be quantized. The side effect is that, since the inputs are also quantized, this led to duplicate pairs of Q/DQ when one conv follows another. The duplicate pairs are removed during ONNX export, but that results to drop in accuracy when evaluating the ONNX.

The new quantization modifier offers more control over the where quantization is applied, and removes the duplicate Q/DQ pairs. However, since the add operation is not a module, its inputs are not quantized for the old model definition. In addition the quantization modifier cannot quantize modules w/ more than one input, so one cannot simply wrap the add operation into a module.

To fix this, I defined a NoOp module, which is applied to the inputs to the add operation separately. The NoOps do nothing on the FP32 model, and simply quantize the inputs during QAT.

I verified that the changes lead to the correct ONNX graph.

For backward compatibility testing, I evaluated the FP32 and INT8 versions of the pruned95 model (zoo:cv/classification/resnet_v1-50/pytorch/sparseml/imagenet/pruned95-none and zoo:cv/classification/resnet_v1-50/pytorch/sparseml/imagenet/pruned95_quant-none) for accuracy and exported the corresponding models to onnx.


src/sparseml/pytorch/models/classification/resnet.py
```diff
@@ -140,26 +140,17 @@ def required(in_channels: int, out_channels: int, stride: int) -> bool:
         return in_channels != out_channels or stride > 1
 
 
-class _AddReLU(Module):
+class _NoOp(Module):
```
bfineran, line 143: we might want a more descriptive input here since this will be printed out as part of the model definition. maybe `_AddInput`

anmarques: Good suggestion

PR #1328, opened 2023-01-19
EfficientNet definition for new quantization modifier

anmarques:
- Remove SiLU w/ hard-coded QATWrappers
- Add MulInput module (no op) that enables control of quantization of SE input


rahul-tuli (approved):

shubhra (approved):

src/sparseml/pytorch/models/classification/efficientnet.py
```diff
@@ -82,26 +82,12 @@ def forward(self, a: Tensor, b: Tensor):
...
-        self.qat_wrapper_kwargs = {
-            "num_inputs": 1,
-            "num_outputs": 0,
-        }
-
-
-class QATSiLUOutput(SiLU):
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+class MulInput(Module):
```
abhinavnmagic, line 85: PyTorch has an `Identity` Module as part of its `nn` library. Why don't we just use that?

anmarques: The identity module is not quantized by default

abhinavnmagic (approved):

PR #1330, opened 2023-01-20
Sparse yolov8

corey-nm:
Yolov8 is organized in the following structure:
1. There is the `ultralytics.yolo.engine.trainer.BaseTrainer` class, which has all the training logic
2. Each of v8's `ClassificationTrainer`, `DetectionTrainer`, and `SegmentationTrainer` inherit from `BaseTrainer`
3. The CLI for each of the trainers is handle separately with configuration files (a copy of which I added as `default.yaml`)

In the target branch `yolov8-feature-branch` I created a `class SparseTrainer` that inherits from `BaseTrainer` and copies some of the methods directly from BaseTrainer that we need to override. Additionally I added the following sub classes which will serve as our copies of the concrete trainers from yolov8.

```python
class SparseDetectionTrainer(SparseTrainer, DetectionTrainer):
    pass


class SparseClassificationTrainer(SparseTrainer, ClassificationTrainer):
    pass


class SparseSegmentationTrainer(SparseTrainer, SegmentationTrainer):
    pass
```

This PR modifies the `SparseTrainer` class with the necessary sparseml code. It contains the following features:
1. Receive `recipe`, `recipe_args`, `checkpoint_path` from configuration
2. Load checkpoint or model from zoo
3. Update DDP implementation to use our classes (this is one of the sketchiest parts IMO)
4. Initialize/use our managers
5. Initializer our loggers
6. Interact with EMA/AMP
7. Interact with yolov8's weird gradient accumulation

# Test plan

I ran training with no arguments (yolov8 training), and with the following recipe:

```yaml
version: 1.1.0

# General variables
num_epochs: 55
quantization_epochs: 5
quantization_lr: 1.e-5
final_lr: 1.e-9

training_modifiers:
  - !EpochRangeModifier
    start_epoch: 0
    end_epoch: eval(num_epochs)

    start_epoch: eval(num_epochs - quantization_epochs)
    end_epoch: eval(num_epochs)
    lr_func: cosine
    init_lr: eval(quantization_lr)
    final_lr: eval(final_lr)

pruning_modifiers:
  - !ConstantPruningModifier
    start_epoch: 0.0
    params: __ALL_PRUNABLE__

quantization_modifiers:
  - !QuantizationModifier
    start_epoch: eval(num_epochs - quantization_epochs)
    submodules:
      - model
    custom_quantizable_module_types: ['SiLU']
    exclude_module_types: ['SiLU']
    quantize_conv_activations: False
    disable_quantization_observer_epoch: eval(num_epochs - quantization_epochs + 2)
    freeze_bn_stats_epoch: eval(num_epochs - quantization_epochs + 1)
```

And verified that training graphs appear in wandb


bfineran (approved):
LGTM overall - would it not be possible to inherit more from `BaseTrainer`?

corey-nm:
> LGTM overall - would it not be possible to inherit more from `BaseTrainer`?

Actually yeah now that I think about it we should be able to modify lr scheduler without overriding _setup_train entirely - will make this change.

corey-nm:
> > LGTM overall - would it not be possible to inherit more from `BaseTrainer`?
> 
> Actually yeah now that I think about it we should be able to modify lr scheduler without overriding _setup_train entirely - will make this change.

@bfineran updated

src/sparseml/pytorch/yolov8/trainers.py
```diff
@@ -58,17 +114,117 @@ def train(self):
...
 
     def resume_training(self, ckpt):
-        return super().resume_training(ckpt)
+        # NOTE: this method is called at the end of super()._setup_train()
+        super().resume_training(ckpt)
+
+        if ckpt is not None:
+            # resume - set manager from checkpoint
+            if "recipe" not in ckpt:
+                raise ValueError("resume is set not checkpoint does not have recipe")
```
KSGulin, line 123: I think this is a typo

src/sparseml/pytorch/yolov8/trainers.py
```diff
@@ -58,17 +114,117 @@ def train(self):
             self._do_train(int(os.getenv("RANK", -1)), world_size)
 
     def resume_training(self, ckpt):
-        return super().resume_training(ckpt)
+        # NOTE: this method is called at the end of super()._setup_train()
+        super().resume_training(ckpt)
+
+        if ckpt is not None:
+            # resume - set manager from checkpoint
+            if "recipe" not in ckpt:
```
KSGulin, line 122: would this cause issues with nominal yolov5 flows?

src/sparseml/pytorch/yolov8/trainers.py
```diff
@@ -58,17 +114,117 @@ def train(self):
...
+                loggers.append(WANDBLogger(init_kwargs=dict(config=config)))
+            except ImportError:
+                warnings.warn("Unable to import wandb for logging")
+            self.logger_manager = LoggerManager(loggers)
+
+        if self.manager is not None:
+            self.args.epochs = self.manager.max_epochs
+
+            if self.manager.learning_rate_modifiers:
+                self.scheduler = _NullLRScheduler()
```
KSGulin, line 161: Nice!

src/sparseml/pytorch/yolov8/trainers.py
```diff
@@ -58,17 +114,117 @@ def train(self):
...
+                self.checkpoint_manager, self.manager
+            )
+        else:
+            manager = self.manager if self.manager is not None else None
+
         ckpt = {
-            "epoch": self.epoch,
+            "epoch": epoch,
             "best_fitness": self.best_fitness,
             "model": deepcopy(de_parallel(self.model)).half(),
```
KSGulin, line 229: Should probably be updated to save as weight dict, to avoid the same pickling issues as in yolov5

KSGulin (changes requested):
Love it. An inheritance-based, forkless integration got me feeling some type of way. Left a few comments, otherwise LGTM

src/sparseml/pytorch/yolov8/trainers.py
corey-nm: Oooh good call. I think we also expect `"state_dict"` key instead of `"model"` right? At least for torchvision we did... Does yolov5 use model or state_dict?

src/sparseml/pytorch/yolov8/trainers.py
corey-nm: Good call out, I think we can just remove

src/sparseml/pytorch/yolov8/trainers.py
KSGulin: it's still "model" in `yolov5`. I think it'd be fine to keep it that way in yolov8 as well. Also note same should apply to ema. ema will also need some conditional logic for saving when it's `None` during quantizations

corey-nm: Good call, will fix that!

dbogunowicz (approved):

bfineran (approved):

src/sparseml/pytorch/yolov8/trainers.py
```diff
@@ -58,33 +114,142 @@ def train(self):
...
-            "ema": deepcopy(self.ema.ema).half(),
-            "updates": self.ema.updates,
+            "model": deepcopy(de_parallel(self.model)).half().state_dict(),
             "optimizer": self.optimizer.state_dict(),
             "train_args": self.args,
             "date": datetime.now().isoformat(),
             "version": __version__,
         }
 
+        if self.ema is not None:
```
KSGulin, line 235: I think the ema issue may need some additional workarounds to solve. Detailed comment incoming

KSGulin (changes requested):

KSGulin:
Spotted a couple concerns that can come up with ema with the updated flow:

1. [This line](https://github.com/ultralytics/ultralytics/blob/15b3b0365ab2f12993a58985f3cb7f2137409a0c/ultralytics/yolo/engine/trainer.py#L338) will cause issues when we disable EMA. I imagine similar issues may come up with the validation/export flows. I think the cleanest solution here would probably be an ema wrapper class that can be turned off

2. This is more of a nit, but I think we should be saving something under the `ema` and `updates` keys even if ema is turned off. Specifically, thinking about this through the use case of continued fine-tuning on a quantized checkpoint. Even if we go with the wrapper class, I'm not convinced we'd want to save the full disabled ema state though. Maybe saving ema as `None` and reconstructing a randomly initialized disabled ema class upon loading of a checkpoint with a `None` ema would be a sensible solution. Otherwise we'd have to either pickle the ema or have some robust state re-loading logic, which I don't think it worthwhile since we shouldn't care about the state of a disabled ema

corey-nm:
@KSGulin pushed update. still working on testing, but figure you can review while i do that

corey-nm:
@KSGulin FYI i'm merging this into feature branch, and will continue work in #1342

PR #1331, opened 2023-01-20
Adding train cli script to yolov8

corey-nm:
This PR adds a training script around our yolov8 classes. It overrides the relevant classes with our own sparse trainer classes. All the options & help strings were pulled directly from yolov8's default.yaml file.

Tested along with #1330


KSGulin (approved):
LGTM!

PR #1334, opened 2023-01-23
Update: Torchvision checkpoint saving to include arch_key

rahul-tuli:
This PR adds support to save the arch_key along with the checkpoint for sparseml's torchvision integration:

The fix involved two steps:
1)  Save: the `arch_key` in the checkpoint dict, 
2) Update: Model creation code, to try `ModelRegistry.create` even when `arch_key` is `None`; This tries to read the `arch_key` from the model checkpoint, if key is not present the relevant functions still raise the respective errors, else it is read from the checkpoint and returned, this was done to avoid re-loading the checkpoint

After this PR the checkpoint looks like the following (checked after 1 epoch from a training run):
```python
In [1]: import torch

In [2]: checkpoint = torch.load("./checkpoint.pth")

In [3]: checkpoint.keys()
Out[3]: dict_keys(['state_dict', 'optimizer', 'args', 'arch_key', 'epoch', 'recipe'])

In [4]: checkpoint["arch_key"]
Out[4]: 'resnet50'

```

### DIFF
--- a/src/sparseml/pytorch/torchvision/train.py
+++ b/src/sparseml/pytorch/torchvision/train.py
@@ -368,7 +368,7 @@ def main(args):
 
     _LOGGER.info("Creating model")
     local_rank = args.rank if args.distributed else None
-    model = _create_model(
+    model, arch_key = _create_model(
         arch_key=args.arch_key,
         local_rank=local_rank,
         pretrained=args.pretrained,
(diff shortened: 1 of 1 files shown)


dbogunowicz (approved):

corey-nm (approved):
nice!

dbogunowicz (approved):

bfineran (approved):

PR #1329, opened 2023-01-20
Adding check for weight.requires_grad in get_prunable_layers

corey-nm:
This PR resolves an issue (traceback below) I am seeing with yolov8 when using the recipe (also below). The fix is to also add a requirement that `layer.weight.requires_grad == True` in `get_prunable_layers`.

# Recipe

```yaml
version: 1.1.0

# General variables
num_epochs: 55
quantization_epochs: 5
quantization_lr: 1.e-5
final_lr: 1.e-9

training_modifiers:
  - !EpochRangeModifier
    start_epoch: 0
    end_epoch: eval(num_epochs)

  - !LearningRateFunctionModifier
    start_epoch: eval(num_epochs - quantization_epochs)
    end_epoch: eval(num_epochs)
    lr_func: cosine
    init_lr: eval(quantization_lr)
    final_lr: eval(final_lr)

pruning_modifiers:
  - !ConstantPruningModifier
    start_epoch: 0.0
    params: __ALL_PRUNABLE__

quantization_modifiers:
  - !QuantizationModifier
    start_epoch: eval(num_epochs - quantization_epochs)
    submodules:
      - model
    custom_quantizable_module_types: ['SiLU']
    exclude_module_types: ['SiLU']
    quantize_conv_activations: False
    disable_quantization_observer_epoch: eval(num_epochs - quantization_epochs + 2)
    freeze_bn_stats_epoch: eval(num_epochs - quantization_epochs + 1)
```

# Traceback

```python
  File "<snip>/ultralytics/yolo/engine/trainer.py", line 402, in optimizer_step
    self.scaler.step(self.optimizer)
  File "<snip>/sparseml/src/sparseml/pytorch/optim/manager.py", line 172, in step
    return self._perform_wrapped_step(*args, **kwargs)
  File "<snip>/sparseml/src/sparseml/pytorch/optim/manager.py", line 209, in _perform_wrapped_step
    self._wrapped_manager.update(
  File "<snip>/sparseml/src/sparseml/pytorch/optim/manager.py", line 589, in update
    mod.scheduled_update(module, optimizer, epoch, steps_per_epoch)
  File "<snip>/sparseml/src/sparseml/pytorch/sparsification/modifier.py", line 620, in scheduled_update
    self.update(module, optimizer, epoch=epoch, steps_per_epoch=steps_per_epoch)
  File "<snip>/sparseml/src/sparseml/pytorch/sparsification/modifier.py", line 452, in wrapper
    out = func(*args, **kwargs)
[...]

PR #1336, opened 2023-01-24
[Feature Branch] Yolov8

corey-nm:
Adds yolov8 training script to the repo.

# Test plan

## Recipe

```yaml
version: 1.1.0

training_modifiers:
  - !EpochRangeModifier
    start_epoch: 0
    end_epoch: 15

pruning_modifiers:
  - !GMPruningModifier
    init_sparsity: 0.05
    final_sparsity: 0.10
    start_epoch: 5.0
    end_epoch: 10.0
    update_frequency: 1.0
    params: ["model.0.conv.weight", "model.1.conv.weight"]

quantization_modifiers:
  - !QuantizationModifier
    start_epoch: 11.0
    freeze_bn_stats_epoch: 12.0
    disable_quantization_observer_epoch: 13.0
    ignore: ["Upsample", "Concat", "SiLU"]
```

There are two parts to this:
1. the DFL final layer in the yolo models are detached (they do not require grads). Therefore, the pruner must ignore them as this currently crashes (see potential fix in #1329).
2. Due to ultralytics adding special attributes to top layer modules of the model, we must ignore Upsample & Concat for quantization. These attributes are [set on this line](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/nn/tasks.py#L451), and [used on this line](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/nn/tasks.py#L53). Without ignoring Upsample & Concat, these two layers get replaced with `QuantWrapper`, which does not have these special attributes set.

### Base ✅

`CUDA_VISIBLE_DEVICES=0 python src/sparseml/pytorch/yolov8/train.py --name dense-base`

### From local yolo checkpoint ✅

After running the base training command above:

`CUDA_VISIBLE_DEVICES=0 python src/sparseml/pytorch/yolov8/train.py --name dense-local-yolo --model runs/detect/dense-base/weights/best.pt`

Should properly print:
```
...
Transferred 355/355 items from pretrained weights
...
```

### Continuing an interrupted run ✅

After running the base training command above:
`CUDA_VISIBLE_DEVICES=0 python src/sparseml/pytorch/yolov8/train.py --name dense-local-yolo --resume --model runs/detect/dense-base/weights/last.pt`

Properly prints:
```
...
Loading local ultralytics checkpoint
...
Transferred 355/355 items from pretrained weights
...
Resuming training from <path>/weights/last.pt from epoch 5 to 15 total epochs
```

### Base ✅

`CUDA_VISIBLE_DEVICES=0 python src/sparseml/pytorch/yolov8/train.py --name pq --recipe yolov8-pq.yaml`

- Should train for 15 epochs.
- Pruning should be logged & change in wandb
- QuantizationModifier should be logged & wandb after it is applied

### Continuing from an interrupted run before quantization ✅

Run training for a few epochs, ctrl+c, then use the following command:

`CUDA_VISIBLE_DEVICES=0 python src/sparseml/pytorch/yolov8/train.py --resume --model <path>/weights/last.pt`

You should see the following message:

```
...
Loading local sparseml checkpoint
...
Applying structure from un-finished recipe in checkpoint at epoch <>
Loaded previous weights from sparseml checkpoint
...
```

### Continuing from an interrupted run after quantization ✅

Run training for a few epochs until quantization is applied, ctrl+c, then use the following command:

`CUDA_VISIBLE_DEVICES=0 python src/sparseml/pytorch/yolov8/train.py --resume --model <path>/weights/last.pt`

You should see:
```
...
Loading local sparseml checkpoint
...
Applying structure from un-finished recipe in checkpoint at epoch <>
...
```

### **NOT SUPPORTED** Fine tuning a quantized model with no recipe provided ❌ 

Run training with recipe to completion, then run:

`CUDA_VISIBLE_DEVICES=0 python src/sparseml/pytorch/yolov8/train.py --model <path>/weights/last.pt`

You should see:
```
...
Loading local sparseml checkpoint
...
Applying structure from completed recipe in checkpoint at epoch <>
...
```

### DIFF
--- a/src/sparseml/pytorch/yolov8/trainers.py
+++ b/src/sparseml/pytorch/yolov8/trainers.py
@@ -33,7 +33,8 @@ from ultralytics import __version__
 from ultralytics.nn.tasks import attempt_load_one_weight
 from ultralytics.yolo.engine.model import YOLO
 from ultralytics.yolo.engine.trainer import BaseTrainer
-from ultralytics.yolo.utils import LOGGER
+from ultralytics.yolo.utils import LOGGER, yaml_load
+from ultralytics.yolo.utils.checks import check_yaml
 from ultralytics.yolo.utils.dist import (
     USER_CONFIG_DIR,
     ddp_cleanup,
@@ -94,14 +95,6 @@ class SparseTrainer(BaseTrainer):
                 Model(self.model), model_suffix=".pt"
             )
 
-        if (
-            self.args.checkpoint_path is not None
-            and self.args.checkpoint_path.startswith("zoo:")
-        ):
-            self.args.checkpoint_path = download_framework_model_by_recipe_type(
-                Model(self.args.checkpoint_path), model_suffix=".pt"
-            )
-
         self.manager: Optional[ScheduledModifierManager] = None
         self.checkpoint_manager: Optional[ScheduledModifierManager] = None
         self.logger_manager: LoggerManager = LoggerManager(log_python=False)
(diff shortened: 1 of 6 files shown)


KSGulin:
Should we move this to sparseml/yolov8, on the same directory level as yolov5? For context, yolov5 was placed there for high visibility, as it is it's own thing like transformers is, rather than a native pytorch implementation

KSGulin:
@corey-nm @bfineran is there a plan to integrate into the `ultralytics/ultralytics` upstream as we are in yolov5? If so, may be good to keep that in mind with this integration. Since we're going for a forkless integration, would the target then be to have two separate integrations? One in sparseml and one in the ultralytics upstream?

corey-nm:
> Should we move this to sparseml/yolov8, on the same directory level as yolov5? For context, yolov5 was placed there for high visibility, as it is it's own thing like transformers is, rather than a native pytorch implementation

I would say high visibility will be achieved through documentation (README/sample code/etc). I don't expect people will be discovering we support yolov8 through looking at the src directory, and this way we know it depends on pytorch.

corey-nm:
> is there a plan to integrate into the `ultralytics/ultralytics` upstream as we are in yolov5? If so, may be good to keep that in mind with this integration. Since we're going for a forkless integration, would the target then be to have two separate integrations? One in sparseml and one in the ultralytics upstream?

If we get it into upstream I would vote for removing our sparseml version of it. I haven't heard plans of contributing to upstream though?

KSGulin:
> I would say high visibility will be achieved through documentation (README/sample code/etc). I don't expect people will be discovering we support yolov8 through looking at the src directory, and this way we know it depends on pytorch.

Fair enough, I think the argument could be made either way and I can't say I feel too strongly about either option. But I do feel that it's counterintuitive to have yolov5 and yolov8 in different directories

KSGulin:
> If we get it into upstream I would vote for removing our sparseml version of it. I haven't heard plans of contributing to upstream though?

My understanding is that an upstream push will be a part of the yolov8 integration. Probably worth further discussion with the team

KSGulin:
Nice workaround for the detached layers. To complete testing coverage I'd recommend adding:

- Continuing an interrupted run with the --resume keyword (base)
- Continuing an interrupted run with the --resume keyword (recipe)
- Continuing fine-tuning quantized model without recipe

src/sparseml/pytorch/yolov8/trainers.py
```diff
@@ -0,0 +1,548 @@
...
+
+class SparseDetectionTrainer(SparseTrainer, DetectionTrainer):
+    pass
+
+
+class SparseClassificationTrainer(SparseTrainer, ClassificationTrainer):
+    pass
+
+
+class SparseSegmentationTrainer(SparseTrainer, SegmentationTrainer):
```
bfineran, line 398: nice

bfineran (approved):
LGTM @corey-nm! As a quick follow up - let's get the PR description for this adapted into a rough README (to be iterated on by product/research)

rahul-tuli (approved):

PR #1337, opened 2023-01-24
Saving both original and final recipes in torchvision train script

corey-nm:
Feature request was for any arguments passed via `--recipe-args` to be reflected in saved file in wandb.

This PR does the following:
1. Always saves a file for recipe, regardless of if args.recipe is a zoo stub or actual file
2. Saves the original unmodified/unsubstituted recipe as `{args.output_dir}/original_recipe.yaml`
3. Saves final recipe with variables from `--recipe-args` and other variables substituted as `{args.output_dir}/final_recipe.yaml`

Example in wandb:
image

image

image

# Testing plan

Tested with command from README with both no recipe-args and recipe-args specified.

### DIFF
--- a/src/sparseml/pytorch/torchvision/train.py
+++ b/src/sparseml/pytorch/torchvision/train.py
@@ -24,7 +24,6 @@ import warnings
 from functools import update_wrapper
 from types import SimpleNamespace
 from typing import Callable, Optional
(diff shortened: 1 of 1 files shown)


natuan (approved):
LGTM. Thank you @corey-nm!

dbogunowicz (approved):

PR #1338, opened 2023-01-24
Removing distill_teacher model from args

corey-nm:
Fixes the bug `AttributeError: Can't pickle local object '_create_cache_output_hook.<locals>.forward_hook_fn'` with distillation.

This occurred because `args` are pickled in the checkpoint, and the whole distillation teacher model was being put into `args`. Moved it to a local variable instead.

# Test plan

Ran training for 1 epoch with recipe from asana ticket and verified save after fix.


natuan (approved):
Beautiful bug (fix)!

anmarques (approved):

bfineran (approved):

PR #1339, opened 2023-01-25
Fix editing a list inside a loop

anmarques:
Logic for converting signed to unsigned initializers was editing the initializer list inside a loop. Replaced by 2 stages: first create new initializers and the ones to remove, then edit the initializer list.

### DIFF
--- a/src/sparseml/pytorch/sparsification/quantization/quantize_qat_export.py
+++ b/src/sparseml/pytorch/sparsification/quantization/quantize_qat_export.py
@@ -263,8 +263,7 @@ def _convert_signed_to_unsigned(model: ModelProto):
 
     for init in to_remove:
         model.graph.initializer.remove(init)
-    for init in to_append:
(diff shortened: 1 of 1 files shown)


src/sparseml/pytorch/sparsification/quantization/quantize_qat_export.py
```diff
@@ -1583,6 +1586,7 @@ def quantize_torch_qat_export(
     _convert_quantizable_gemm_no_activations(model)
     quantize_resnet_identity_add_inputs(model)
     _remove_duplicate_quantize_ops(model)
+    _convert_signed_to_unsigned(model)
```
abhinavnmagic, line 1589: Is there a reason behind shifting this signed to unsigned conversion at the last?

abhinavnmagic (approved):

rahul-tuli (approved):

PR #1340, opened 2023-01-25
Fix referenced before assignment error

abhinavnmagic:
This should not reverse, but just fix: https://github.com/neuralmagic/sparseml/pull/1338


natuan (approved):

corey-nm (approved):
Nice fix!!

PR #1341, opened 2023-01-26
Compute overall accuracy using number of correct predictions

abhinavnmagic:


src/sparseml/pytorch/torchvision/train.py
```diff
@@ -138,7 +138,7 @@ def train_one_epoch(
                 # Reset ema buffer to keep copying weights during warmup period
                 model_ema.n_averaged.fill_(0)
 
-        acc1, acc5 = utils.accuracy(output, target, topk=(1, 5))
+        acc1, _1, acc5, _5 = utils.accuracy(output, target, topk=(1, 5))
         batch_size = image.shape[0]
         metric_logger.update(loss=loss.item(), lr=optimizer.param_groups[0]["lr"])
         metric_logger.meters["acc1"].update(acc1.item(), n=batch_size)
```
corey-nm, line 144: Does the new total param need to be added to this & acc5 like in evaluate?

src/sparseml/pytorch/torchvision/utils.py
```diff
@@ -210,6 +213,7 @@ def accuracy(output, target, topk=(1,)):
         for k in topk:
             correct_k = correct[:k].flatten().sum(dtype=torch.float32)
             res.append(correct_k * (100.0 / batch_size))
+            res.append(correct_k)
```
corey-nm, line 216: Is this understanding of the changes correct?

Before:
```
self.total += value * n
```
where value = correct_k * 100 / batch_size, and n = batch size,
so `correct_k * 100` would be added to total.

Now `correct_k` is being added to the total.

So since we are removing the factor of 100, that is why you added it back to the global_avg function?

abhinavnmagic: Yes, essentially! By doing this, we avoid inconsistencies with floating point due to extra multiplication and divisions.

src/sparseml/pytorch/torchvision/train.py
abhinavnmagic: For evaluation, it is added since we wish to compute the global accuracy.
For training, global avg doesn't make sense since the model keeps evolving. However, we can include if we wish to maintain uniformity. Let me know if that's preferred and I will update.

natuan: yeah no we don't want to compute global accuracy during training. Nice to have is the running average across batches to smooth the curve, but i can still live w/o it.

corey-nm (approved):
lgtm! 👍

natuan:
@abhinavnmagic After this is approved, before merging can you also use the fix here to do at least a full training with the model you're working on? I'd like to make sure we won't see anything unexpected.

corey-nm (dismissed):

src/sparseml/pytorch/torchvision/utils.py
```diff
@@ -210,6 +213,7 @@ def accuracy(output, target, topk=(1,)):
         for k in topk:
             correct_k = correct[:k].flatten().sum(dtype=torch.float32)
             res.append(correct_k * (100.0 / batch_size))
+            res.append(correct_k * 100.0)
```
natuan, line 216: @abhinavnmagic Isn't correct_k already the number of correct predictions? I'm confused as to why we need to the multiplication here.

abhinavnmagic: I initially didn't multiply by 100 here; and adjusted this in the global_avg function. However, that results in the scaling of losses by 100 (during training run), which is something we don't want. Due to this reason, I moved the scaling to the implementation here. I do agree that this results in this multiplication here, but I don't have any other alternative.

natuan: ok at least having the scaling here is less confusing than in the global_avg.

natuan (approved):

anmarques (approved):

PR #1343, opened 2023-01-26
Ignoring Gemm -> Q/Dq -> Softmax in GemmToQLinearMatMul transform

corey-nm:
When there is a Softmax after a MatMul, this pass improperly assumes the Q/Dq is for itself, rather than for the softmax.

# Original quantized graph

image

# Current deepsparse graph from exporters

image

Note that this is invalid!!

# deepsparse graph after this PR

image

### DIFF
--- a/src/sparseml/exporters/transforms/gemm_to_qlinearmatmul.py
+++ b/src/sparseml/exporters/transforms/gemm_to_qlinearmatmul.py
@@ -12,8 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import logging
-
 from onnx import ModelProto, helper, numpy_helper
 
 from sparseml.exporters.transforms.onnx_transform import OnnxTransform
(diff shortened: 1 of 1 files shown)


src/sparseml/exporters/transforms/gemm_to_qlinearmatmul.py
```diff
@@ -93,7 +90,10 @@ def transform(self, model: ModelProto) -> ModelProto:
             output_dequant = match.children[0][1]
             if output_dequant is not None:
                 output_dequant_child = graph.get_node_single_child(output_dequant)
-                if output_dequant_child and output_dequant_child.op_type == "Gemm":
+                if output_dequant_child and output_dequant_child.op_type in {
+                    "Gemm",
+                    "Softmax",
```
corey-nm, line 95: Will this adversely affect transformers? Not sure if MHA in transformers gets exported as Gemm or MatMul

bfineran: generally MatMul, also we have not been targeting output quantization there anyways

Issue #1347, opened 2023-01-30
OBSPruningModifier fails if params are non-contiguous


eldarkurtic:
**Describe the bug**
Pruning fails if some params are non-contiguous in memory, for example when channels_last format is used for Conv layers.
Thanks to @Godofnothing for pointing out the issue.

**To Reproduce**
Run pruning of any model with channel_last format used in the Conv layers.

PR #1348, opened 2023-01-30
Use more robust reshape instead of view

eldarkurtic:
Fixes https://github.com/neuralmagic/sparseml/issues/1347 by replacing `.view` calls with more robust `.reshape`.


bfineran (approved):
LGTM

PR #1357, opened 2023-02-01
Yolov8 checkpoints v2

corey-nm:
Improved loading logic. Will test completely with test plan in feature branch PR before merging.

New logic is:
1. Create model
2. Create managers
3. Apply structure from checkpoint
4. Load state dict

Additionally, this handles the finetuning over quantized model case, where checkpoint_manager is not None and manager is None

### DIFF
--- a/src/sparseml/pytorch/yolov8/trainers.py
+++ b/src/sparseml/pytorch/yolov8/trainers.py
@@ -33,7 +33,8 @@ from ultralytics import __version__
 from ultralytics.nn.tasks import attempt_load_one_weight
 from ultralytics.yolo.engine.model import YOLO
 from ultralytics.yolo.engine.trainer import BaseTrainer
-from ultralytics.yolo.utils import LOGGER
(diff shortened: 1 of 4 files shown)


src/sparseml/pytorch/yolov8/trainers.py
```diff
@@ -138,32 +130,80 @@ def train(self):
...
+
+        else:
+            # resuming
+            # yolo will populate this when the --resume flag
+            assert self.args.recipe is not None
+            LOGGER.info(
+                "Applying structure from un-finished recipe in checkpoint "
+                f"at epoch {ckpt['epoch']}"
+            )
+            self.manager.apply_structure(self.model, epoch=ckpt["epoch"])
```
bfineran, line 205: instead of apply structure, if we are resuming why not initialize the manager to the epoch we are resuming at?

bfineran (commented):
lgtm pending comment

dbogunowicz (approved):

PR #1367, opened 2023-02-07
add support for deeplake datasets

mwitiderrick:


src/sparseml/pytorch/torchvision/train.py
```diff
@@ -1174,9 +1222,40 @@ def new_func(*args, **kwargs):
...
+    "--use_deeplake", default=None, type=bool, help="Whether to use deeplake datasets"
+)
+@click.option(
+    "--deeplake_token", default=None, type=str, help="Token to authenticate download"
+)
+@click.option(
+    "--deeplake_image_column",
+    default="images",
+    type=str,
+    help="Image column of the dataset",
```
dbogunowicz, line 1245: What is image column?

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -1174,9 +1222,40 @@ def new_func(*args, **kwargs):
...
+)
+@click.option(
+    "--deeplake_label_column",
+    default="labels",
+    type=str,
+    help="Label column of the dataset",
```
dbogunowicz, line 1251: Same

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -1174,9 +1222,40 @@ def new_func(*args, **kwargs):
...
+)
+@click.option(
```
dbogunowicz, line 1253: Can't we just reuse the original `image_size` argument?

dbogunowicz: `training_crop` size of whatever

src/sparseml/pytorch/torchvision/export_onnx.py
```diff
@@ -54,13 +56,6 @@
     help="A path to a previous checkpoint to load the state from "
     "and resume the state for exporting, or a zoo stub.",
 )
-@click.option(
```
dbogunowicz, line 57: Why are we removing this argument? Don't we want to have both options allowed for the user?

src/sparseml/pytorch/torchvision/export_onnx.py
```diff
@@ -134,24 +146,49 @@ def main(
...
-    dataset = torchvision.datasets.ImageFolder(
-        dataset_path / "val",
-        presets.ClassificationPresetEval(
-            crop_size=img_crop_size,
-            resize_size=img_resize_size,
-            interpolation=InterpolationMode(interpolation),
-        ),
+    tform = deeplake_transforms.Compose(
+        [
+            deeplake_transforms.RandomRotation(20),  # Image augmentation
```
dbogunowicz, line 167: why do we need to specify image augmentation for onnx export? Those two things are not related.

dbogunowicz: also, I would move the whole deeplake logic into a helper function, as not to overclutter this script.

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -340,45 +339,94 @@ def main(args):
...
+            dataset_test,
+            batch_size=args.batch_size,
+            sampler=test_sampler,
+            num_workers=args.workers,
+            pin_memory=True,
+        )
+    elif args.use_deeplake:
+        tform = deeplake_transforms.Compose(
+            [
+                deeplake_transforms.RandomRotation(20),  # Image augmentation
```
dbogunowicz, line 385: Where does `20` come from? Is it consistent with the alternative creation of the dataset (through the `args.dataset_path`). We do not want to introduce some hidden arguments under the hood that are obfuscated from the user.

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -340,45 +339,94 @@ def main(args):
...
+                deeplake_transforms.Resize(args.deeplake_image_size),
+                deeplake_transforms.ToTensor(),  # Must convert to pytorch tensor for subsequent operations to run
+                deeplake_transforms.Normalize([0.5], [0.5]),
```
dbogunowicz, line 388: Similar situation here: [0.5] and [0.5] may not work if we need a specific normalization statistics required by Imagenet

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -340,45 +339,94 @@ def main(args):
...
+            ]
+        )
+        ds_train = deeplake.load(
```
dbogunowicz, line 391: I'd hide it in a helper function in a separate auxiliary script, `train.py` should be fairly compact and readable.

src/sparseml/pytorch/torchvision/train.py
```diff
@@ -886,7 +934,7 @@ def new_func(*args, **kwargs):
     type=str,
     help="json parsable dict of recipe variable names to values to overwrite with",
 )
-@click.option("--dataset-path", required=True, type=str, help="dataset path")
```
dbogunowicz, line 889: Maybe it would a good idea to check whether the user has specified either a sufficient set of arguments to create a dataset from deeplake or from generic dataset, and throw error if both pathways or none of them is enabled.

tests/sparseml/tensorflow_v1/optim/test_modifier_lr.py
```diff
@@ -889,13 +889,13 @@ def test_lrs_with_manager(optim_lambda):
             elif epoch < 10:
                 expected = 0.01 * 0.9 ** (epoch - 5)
             elif epoch < 12:
-                expected = 0.01 * 0.9 ** 4
```
dbogunowicz, line 892: Formatting still seems to be not done correctly in several scripts.

dbogunowicz (commented):
Left some notes

PR #1290, opened 2023-01-04
Add integrations standardization table

corey-nm:
This will be the markdown file that documents the features each integration support.

Any official integration should add a column to all tables in this markdown, and add either a ✅ if that feature is supported and a ❌ or blank if the feature is not supported.

This will provide a central way to track what each integration supports. Since this is all in 1 file, minimal changes should be needed if integrations are added in the future or if a new row needs to be added.

PR #1385, opened 2023-02-14
[integration-standardization] status table and page schemas + sweep script + template

bfineran:
Ports status page and table from #1290 into `FeatureStatusPage` and `FeatureStatusTable` schemas.  Includes script harness for generating markdown status pages from yaml configs, aggregated status page, and template. Aggregated page and template written to `status/`, individual pages written next to their yaml configs nested in `src/`

the template yaml file is auto generated from the schemas by running `status/integration_status.py`

**test_plan:**
manually check generated `status_template.status.yaml`

### DIFF
--- a/status/integration_status.py
+++ b/status/integration_status.py
@@ -83,7 +83,7 @@ class DatasetsStatusTable(FeatureStatusTable):
 class CheckpointsStatusTable(FeatureStatusTable):
     original_integration_checkpoints: FeatureStatus = Field()
     sparsezoo_checkpoints: FeatureStatus = Field()
-    best_checkpoint: FeatureStatus = Field()
+    best_dense_checkpoint: FeatureStatus = Field()
     best_pruned_checkpoint: FeatureStatus = Field()
     best_pruned_quantized_checkpoint: FeatureStatus = Field()
     recipe_saved_to_checkpoint: FeatureStatus = Field()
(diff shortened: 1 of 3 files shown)


status/integration_status.py
```diff
@@ -1,4 +1,185 @@
...
+)
+
+
+class BaseTrainingStatusTable(FeatureStatusTable):
+
+    cli: FeatureStatus = Field()
+    api: FeatureStatus = Field()
+    dense_training: FeatureStatus = Field()
+    gradient_accumulation: FeatureStatus = Field()
+    DP: FeatureStatus = Field()
```
KSGulin, line 39: nit: may be better to rename to lowercase (same for ddp, ema, amp). Having an all caps class field would be somewhat unexpected

status/integration_status.py
```diff
@@ -1,4 +1,185 @@
...
+
+    @property
+    def name(self) -> str:
+        return "Datasets"
+
+
+class CheckpointsStatusTable(FeatureStatusTable):
+    original_integration_checkpoints: FeatureStatus = Field()
+    sparsezoo_checkpoints: FeatureStatus = Field()
+    best_checkpoint: FeatureStatus = Field()
```
KSGulin, line 86: nit
```suggestion
    best_dense_checkpoint: FeatureStatus = Field()
```

KSGulin (approved):
Really clean implementation. Left a couple nits to address, but approved otherwise

status/integration_status.py
KSGulin: Thoughts on also adding formatting that would transform e.g. dense_training -> Dense Training

PR #1386, opened 2023-02-14
[integration-standardization] YAML configs for existing integrations

bfineran:
YAML configs manually created from results compiled in #1290 with some fields updated based on progress since that PR was opened

**test_plan:**
pending auto creation of MD files in follow up PR


KSGulin (approved):
Nice!

PR #1388, opened 2023-02-14
[integration-standardization] readme

bfineran:
also adds `make status` command

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
 .PHONY: build docs test
 
 BUILDDIR := $(PWD)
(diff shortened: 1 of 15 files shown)


KSGulin (commented):
Looks great, I would just add a GHA step to the quality check that makes sure `make status` has been applied, similarly to how `make quality` is checked

bfineran:
agreed this would be a great feature especially for automation - let's add this as an enhancement later on though. for local checks I imagine just going through `git diff` will need to do

PR #1393, opened 2023-02-21
Adding MatMulToMatMulIntegerCastMul

corey-nm:
Also enables optional nodes in parent_ops

Tested with sparsifyml & unit tests

### DIFF
--- a/src/sparseml/exporters/transforms/matmul_add_to_matmulinteger_add_cast_mul.py
+++ b/src/sparseml/exporters/transforms/matmul_add_to_matmulinteger_add_cast_mul.py
@@ -22,14 +22,15 @@ from sparseml.exporters.transforms.utils import (
     any_of,
     get_quantization_params,
     get_structural_matches,
+    optional_node,
 )
 from sparseml.onnx.utils import ONNXGraph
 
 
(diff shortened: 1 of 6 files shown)


src/sparseml/exporters/transforms/matmul_add_to_matmulinteger_add_cast_mul.py
```diff
@@ -41,7 +42,7 @@ class MatMulToMatMulIntegerAddCastMul(OnnxTransform):
     |         |
     | input   Dq
     |   |     |
-    |  Q/Dq   Transpose
+    |  Q/Dq   optional Transpose
```
bfineran, line 45: do we test when this exists and doesn't exist? Potentially recall some issues around optional when doing intiial opt experiments

bfineran (approved):
LGTM pending comment

src/sparseml/exporters/transforms/matmul_add_to_matmulinteger_add_cast_mul.py
corey-nm: Yeah it wasn't supported when specified in parents, part of this PR fixes that bug.

There's a test added also in this PR for it

dbogunowicz (approved):

PR #1396, opened 2023-02-23
[Fix] Explicitly set the providers parameter when instantiating InferenceSession

dbogunowicz:
Fix for: https://app.asana.com/0/1203126676641557/1204007135425947

Reason: Since ORT 1.9, you are required to explicitly set the providers parameter when instantiating InferenceSession.


anmarques (approved):
Confirmed that the fix works

KSGulin (approved):

PR #1403, opened 2023-03-01
Added RigL pruning modifier

Godofnothing:
Implementation of the RigLPruningModifier - a sparse training procedure proposed in the paper https://arxiv.org/abs/1911.11134. 

**Description**

This sparse training procedure performs sparse training of the model with periodic pruning if a certain fraction of weights with smallest magnitude and regrowth (unmasking) of the same fraction according to the magnitude of the gradients, i.e zero weights with large gradients become trainable again. 

This modifier implements the three sparsity distribution strategies proposed in the original paper:  
- Uniform
- Erdos-Renyi
- Erdos-Renyi Kernel

**Usage**

This modifier requires a `GradSampler` (similarly to `OBSPruningModifier`) and collects a single gradient for score estimation.

### DIFF
--- a/src/sparseml/pytorch/sparsification/pruning/modifier_pruning_rigl.py
+++ b/src/sparseml/pytorch/sparsification/pruning/modifier_pruning_rigl.py
@@ -39,10 +39,7 @@ from sparseml.pytorch.utils.logger import BaseLogger
 from sparseml.pytorch.sparsification.pruning.scorer import PruningParamsGradScorer
 
 
-__all__ = [
-    "RigLPruningModifier",
-    "RigLPruningParamsScorer"
-]
+__all__ = ["RigLPruningModifier", "RigLPruningParamsScorer"]
 
 
 _LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


src/sparseml/pytorch/sparsification/pruning/modifier_pruning_rigl.py
```diff
@@ -0,0 +1,443 @@
...
+        params: Union[str, List[str]],
+        leave_enabled: bool = True,
+        global_sparsity: bool = True,
+        mask_type: str = "unstructured",
+        sparsity_strategy: str = "erdos_renyi_kernel",
+        init_update_fraction: float = 0.3,
+    ):
+        super().__init__(
+            params=params,
+            init_sparsity=sparsity * (1 - init_update_fraction),
```
bfineran, line 158: will need to override the inherited `init_sparsity` modifier prop to set `serializable=False`

src/sparseml/pytorch/sparsification/pruning/modifier_pruning_rigl.py
```diff
@@ -0,0 +1,443 @@
...
+        sparsity: float,
+        start_epoch: float,
+        end_epoch: float,
+        update_frequency: float,
+        params: Union[str, List[str]],
+        leave_enabled: bool = True,
+        global_sparsity: bool = True,
+        mask_type: str = "unstructured",
+        sparsity_strategy: str = "erdos_renyi_kernel",
+        init_update_fraction: float = 0.3,
```
bfineran, line 154: should add `ModifierProp` properties for both of these for YAML serialization support

bfineran (commented):
Thanks for the PR @Godofnothing, LGTM overall to merge to `src/` we should get some basic tests in for this modifier (lifecycle and YAML serialization similar to the others).  Depending on scope of usage for now, we could also push directly to the `research/` directory (would just need to add an import of the modifier for it to get registered before recipe instantiation)

src/sparseml/pytorch/sparsification/pruning/modifier_pruning_rigl.py
eldarkurtic: probably the same applies to `final_sparsity`

Godofnothing: @bfineran , @eldarkurtic  thanks for you comments, I have fixed this issue in the next commit.

jeanniefinks:
Per the main [README announcement](https://github.com/neuralmagic/sparseml/blob/main/README.md), SparseML is being deprecated by early June 2, 2025. Closing the PR as work has been suspended; thank you for the inputs and support!

PR #1421, opened 2023-03-09
Update docker infra

rahul-tuli:
This PR is the Sparseml Equivalent of DeepSparse diff https://github.com/neuralmagic/deepsparse/pull/934


KSGulin (approved):
Great stuff

docker/Dockerfile
```diff
@@ -1,58 +1,129 @@
...
 
-ARG GIT_CHECKOUT
-# Install SparseML
-# if $GIT_CHECKOUT is not specified - just install from pypi
-RUN if [ -z "${GIT_CHECKOUT}" ] ; then pip3 install --no-cache-dir --upgrade sparseml[dev,torchvision] ; fi
+FROM base as prod
+ARG VENV
+COPY --from=build $VENV $VENV
+ENV PATH="${VENV}/bin:$PATH"
+HEALTHCHECK CMD python -c 'import sparseml'
```
bfineran, line 127: is this healthcheck run on every sub-container?

bfineran, line 127: would also suggest importing `sparseml.pytorch.sparsification` where applicable to make sure the torch install also went through

bfineran (approved):
LGTM, see comment and let's also update the README if necessary

docker/Dockerfile
rahul-tuli: The healthcheck is run once per build, and not for each individual stage.

rahul-tuli: maybe that's not such a great idea, cause we also have a sparseml-base container that doesn't come with torch. I'm not sure if we can supply multiple healthchecks per Dockerfile, but I'll investigate, the only way we could add this functionality is by supplying a specific custom healthcheck or even a Run command for the relevant stages.

Would it make sense to add this improvement in a subsequent PR?

rahul-tuli:
Note: I'll be updating the docker readmes on both sparseml and deepsparse side in separate PRs, will also link them out here when ready

PR #1423, opened 2023-03-09
Explicitly setting all parameter requires_grad=True

corey-nm:
It seems like in upstream yolov8, they explicitly set all param requires_grad to False

https://github.com/ultralytics/ultralytics/blob/790f9c067c9a3548acf7c6925cea14de9ad77787/ultralytics/yolo/engine/exporter.py#L189

So any model we use from upstream we need to set requires_grad to True for exporting

Test plan: Ran `sparseml.ultralytics.export --model yolov8s.pt --recipe <failing recipe from issue>`

### DIFF
--- a/src/sparseml/yolov8/trainers.py
+++ b/src/sparseml/yolov8/trainers.py
@@ -533,7 +533,6 @@ class SparseYOLO(YOLO):
         else:
             return super()._load(weights)
 
-    @smart_inference_mode()
     def export(self, **kwargs):
         """
         Export model.
@@ -569,6 +568,8 @@ class SparseYOLO(YOLO):
                 f"Detected one-shot recipe: {one_shot}. "
                 "Applying it to the model to be exported..."
             )
+            for p in self.model.parameters():
+                p.requires_grad = True
             manager = ScheduledModifierManager.from_yaml(one_shot)
             manager.apply(self.model)
             recipe = (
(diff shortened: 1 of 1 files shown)


dbogunowicz (commented):
Great!

dbogunowicz (dismissed):

src/sparseml/yolov8/trainers.py
```diff
@@ -636,9 +637,7 @@ def export(self, **kwargs):
                 "Recipe checkpoint detected, saving the "
                 f"recipe to the deployment directory {deployment_folder}"
             )
-            ScheduledModifierManager.from_yaml(recipe).save(
-                os.path.join(deployment_folder, "recipe.yaml")
-            )
+            recipe.save(os.path.join(deployment_folder, "recipe.yaml"))
```
dbogunowicz, line 640: 👀

corey-nm: Oh yeah meant to tag you, I was getting issues when using this with one-shot, because one shot creates a recipe. Assume this is needed when one-shot is not present?

corey-nm: Updated

src/sparseml/yolov8/trainers.py
```diff
@@ -533,7 +533,6 @@ def _load(self, weights: str):
         else:
             return super()._load(weights)
 
-    @smart_inference_mode()
```
abhinavnmagic, line 536: I guess this won't lead to any unintended effects for non one-shot pathways?

abhinavnmagic (approved):

dbogunowicz (approved):

PR #1424, opened 2023-03-10
Utility functions for stage epochs

KSGulin:
When applying recipes at intermediate epochs, it can be non-intuitive how to select the correct epoch, as the epoch used during and saved by training scripts is not the same one that should be used in a staged recipe, as epochs for later stages are shifted up by epochs from prior stages. The intent of the addition of these utility functions is to create a standard approach, where recipes applied at intermediate epochs should always be applied at `manager.get_last_start_epoch() + train_epoch`

**Test plan**
Tested with a Yolov5 usecase. See the associated Yolov5 PR here: neuralmagic/yolov5#189

### DIFF
--- a/src/sparseml/optim/manager.py
+++ b/src/sparseml/optim/manager.py
@@ -650,6 +650,14 @@ class BaseManager(BaseObject):
 
                 stage_max_min[stage] = (epoch_min, epoch_max)
 
+            # Post-process to replace -1's with their real values
+            epochs_list = list(stage_max_min.values())
+            for i, stage, epochs in enumerate(stage_max_min.items()):
+                if epochs[0] == -1:
+                    stage_max_min[stage][0] = epochs_list[i - 1][1] if i > 0 else 0
+                if epochs[1] == -1 and i < len(epochs_list) - 1:
+                    stage_max_min[stage][1] = epochs_list[i + 1][0]
+
             return stage_max_min
 
     def get_last_start_epoch(self) -> float:
--- a/src/sparseml/optim/manager.py
+++ b/src/sparseml/optim/manager.py
@@ -676,4 +676,5 @@ def _max_modifier_epoch(modifiers: Iterable[BaseModifier]) -> float:
         [math.ceil(mod.start_epoch) for mod in modifiers if mod.start_epoch > -1]
     )
     vals.extend([math.ceil(mod.end_epoch) for mod in modifiers if mod.end_epoch > -1])
+
     return max(vals) if len(vals) > 0 else -1
--- a/src/sparseml/optim/manager.py
+++ b/src/sparseml/optim/manager.py
@@ -24,7 +24,7 @@ import math
 from collections import OrderedDict
 from copy import deepcopy
 from functools import cmp_to_key
-from typing import Any, Dict, Generator, List, Optional, Tuple, Union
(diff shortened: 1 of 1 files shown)


src/sparseml/optim/manager.py
```diff
@@ -622,6 +622,45 @@ def phase(self, epoch: float) -> Optional[str]:
...
+
+            return stage_max_min
+
+    def get_last_start_epoch(self) -> float:
+        """
+        Return the start epoch of the last stage in the recipe. Useful for applying
+        recipes at the correct epoch in a staged run
+        """
+        stage_max_min = self.get_start_end_epochs()
+        last_stage_epochs = stage_max_min[next(reversed(stage_max_min))]
```
corey-nm, line 661: clever!

src/sparseml/optim/manager.py
```diff
@@ -622,6 +622,45 @@ def phase(self, epoch: float) -> Optional[str]:
...
+                epoch_min = min(epoch_floors) if epoch_floors else -1
+
+                epoch_ceils = [
+                    math.ceil(mod.start_epoch)
+                    for mod in mod_list
+                    if mod.start_epoch > -1
+                ] + [math.ceil(mod.end_epoch) for mod in mod_list if mod.end_epoch > -1]
+                epoch_max = max(epoch_ceils) if epoch_ceils else -1
+
+                stage_max_min[stage] = (epoch_min, epoch_max)
```
corey-nm, line 651: Since you are doing floor/ceil over both start/end epoch, this isn't technically the start & end epochs right? What's the reasoning for including both in epoch_min/epoch_max? Does something go wrong if you just include start_epoch for epoch_min and end_epoch for epoch_max?

KSGulin: Good question and I'm not 100% sure on the answer. I copied the logic from the modifier props `min_epochs` and `max_epochs` and I assumed they're covering some edge cases, but a good edge case doesn't come to mind right now. 

Actually, a counter-example edge case for how this logic may break comes to mind. If we have a modifier with a positive value end epoch and no start epoch, we could return a value that doesn't make sense here

@dbogunowicz @bfineran what is the original logic meant to cover?

src/sparseml/optim/manager.py
```diff
@@ -622,6 +622,54 @@ def phase(self, epoch: float) -> Optional[str]:
...
+
+            # Post-process to replace -1's with their real values
+            epochs_list = list(stage_max_min.values())
+            for i, stage, epochs in enumerate(stage_max_min.items()):
+                if epochs[0] == -1:
+                    stage_max_min[stage][0] = epochs_list[i - 1][1] if i > 0 else 0
+                if epochs[1] == -1 and i < len(epochs_list) - 1:
```
KSGulin, line 658: By this logic, if there is no explicit end epoch on the last stage, we're keeping it at -1. This is the only -1 we would potentially keep. Comment if you think this should be handled differently

KSGulin:
@corey-nm thanks for the review. I just pushed up another commit as I realized the case of -1's wasn't fully covered

bfineran (dismissed):

corey-nm (dismissed):
nice helpers!

corey-nm (dismissed):

corey-nm (approved):

bfineran (approved):

PR #1433, opened 2023-03-13
[Cherry-pick] Utility functions for stage epochs (#1424)

KSGulin:
Cherry-pick of #1424 for hot-fix 1.4.3

### DIFF
--- a/src/sparseml/optim/manager.py
+++ b/src/sparseml/optim/manager.py
@@ -24,7 +24,7 @@ import math
 from collections import OrderedDict
 from copy import deepcopy
 from functools import cmp_to_key
-from typing import Any, Dict, Generator, List, Optional, Union
+from typing import Any, Dict, Generator, Iterable, List, Optional, Tuple, Union
 
 from sparseml.optim.modifier import BaseModifier, BaseObject, ModifierProp
 from sparseml.sparsification.types import SparsificationTypes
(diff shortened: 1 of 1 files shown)


src/sparseml/optim/manager.py
```diff
@@ -561,6 +529,44 @@ def qat_active(self, epoch: float) -> bool:
...
+        staged manager, map 'all' to the the min and max epochs
+        """
+        if isinstance(self.modifiers, List):
+            return OrderedDict({"all": (self.min_epochs, self.max_epochs)})
+        else:
+            stage_max_min = OrderedDict()
+            for stage, mod_list in self.modifiers.items():
+                epoch_min = _min_modifier_epoch(mod_list)
+                epoch_max = _max_modifier_epoch(mod_list)
+                stage_max_min[stage] = (epoch_min, epoch_max)
```
rahul-tuli, line 544: This is absolutely not a blocker, especially cause this is a cherry-pick, but I feel `stage_min_max` is a better name for this variable

rahul-tuli (approved):

dbogunowicz (approved):

src/sparseml/optim/manager.py
KSGulin: That's a good point. Will be addressed on main, but will keep as is for release

PR #1437, opened 2023-03-15
Enabling OBS training-aware

abhinavnmagic:


dbogunowicz:
@abhinavnmagic Please check whether my recent changes are working correctly, by running export script e.g.
```bash
sparseml.ultralytics.export --one-shot {recipe_with_obs}.yaml --save_one_shot_torch True --model yolov8s.pt
```

src/sparseml/yolov8/trainers.py
```diff
@@ -571,7 +604,20 @@ def export(self, **kwargs):
             for p in self.model.parameters():
                 p.requires_grad = True
             manager = ScheduledModifierManager.from_yaml(one_shot)
-            manager.apply(self.model)
+
+            # TODO: I hate the fact that to scale this functionality to other tasks
+            # we might have to (worst case scenario) copy and paste many
+            # arguments from the train.py. This is not ideal.
```
abhinavnmagic, line 610: Can we not just arguments for data and device to export CLI?

src/sparseml/yolov8/trainers.py
```diff
@@ -571,7 +604,20 @@ def export(self, **kwargs):
...
+            overrides = self.overrides.copy()
+            overrides["data"] = "coco128.yaml"  # args["data"]
+            overrides["device"] = "cpu"
+            trainer = self.TrainerClass(overrides=overrides)
+
+            manager.apply(
+                self.model,
+                # maybe we could check whether OBS pruner is in the manager?
```
abhinavnmagic, line 618: Manager does have OBS Pruner, as part of `manager.pruning_modifiers`.

src/sparseml/yolov8/trainers.py
abhinavnmagic: I guess that's what you mean by copy and paste? Yes, we would essentially have to maintain two pathways: one-shot and training-aware with both of them having similar support.

PR #1451, opened 2023-03-16
Update usage of analyze_model to model_debug_analysis

rahul-tuli:
Cherry Pick for


bfineran (changes requested):
why do we need a cherry pick here? this change is not on the deepsparse release branch

rahul-tuli:
Mistakenly opened this PR, @bfineran is right and this does not need to be cherry picked. Closing it out now

Issue #1438, opened 2023-03-15
PyTorch Lightning - RuntimeError: cannot register a hook on a tensor that doesn't require gradient


clementpoiret:
**Describe the bug**

When trying to use a SparseML callback with Pytorch Lightning, the following error is raised `RuntimeError: cannot register a hook on a tensor that doesn't require gradient`.
Please note I use a local copy of PyTorch lightning bolt SparseMLCallback because their _TORCH_MAX_VERSION does not match yours (I've made a PR for it to allow PT < 1.13.100. The only difference if that I do not check the `torch.__version__`

**Expected behavior**
It should train :)

**Environment**
Include all relevant environment information:
1. OS [e.g. Ubuntu 18.04]: Arch Linux
2. Python version [e.g. 3.7]: 3.9.15
3. SparseML version or commit hash [e.g. 0.1.0, `f7245c8`]: 1.5.0.20230315
4. ML framework version(s) [e.g. torch 1.7.1]: Torch 1.12.1
5. Other Python package versions [e.g. SparseZoo, DeepSparse, numpy, ONNX]:
6. Other relevant environment information [e.g. hardware, CUDA version]: I tried only on CPU

**To Reproduce**
Exact steps to reproduce the behavior:

Just run the training.py of this repo: https://github.com/clementpoiret/sparseml_register_hook_bug/

Pytorch Lightning SparseMLCallback hasn't been updated for a while, maybe its way to do is outdated?

**Errors**

```
/home/cp264607/anaconda3/envs/SparseML/lib/python3.9/site-packages/lightning_fabric/accelerators/cuda.py:219: UserWarning: Can't initialize NVML
  warnings.warn("Can't initialize NVML")
/home/cp264607/anaconda3/envs/SparseML/lib/python3.9/site-packages/pytorch_lightning/trainer/connectors/accelerator_connector.py:737: UserWarning: You passed `Trainer(accelerator='cpu', precision=16)` but native AMP is not supported on CPU. Using `precision='bf16'` instead.
  rank_zero_warn(
Using bfloat16 Automatic Mixed Precision (AMP)
GPU available: False, used: False
TPU available: False, using: 0 TPU cores
IPU available: False, using: 0 IPUs
HPU available: False, using: 0 HPUs
Loading `train_dataloader` to estimate number of stepping batches.
[...]

PR #1463, opened 2023-03-20
Fix hook registration when some parameters do not require grad

clementpoiret:
This PR fixes https://github.com/neuralmagic/sparseml/issues/1438

When setting trainable=false to some parameters using a Training Modifier, the training fails because SparseML tries to create hooks on parameters that do not require grad.

I do not know if this is the correct way to do that, but it fixed the issue I had


bfineran:
Hi @clementpoiret thanks for the PR! This will safely fix your error however since it skips the layers silently at the mask creator level this hides it from the user and also potentially would create a mismatch between the modifier targets and what actually gets pruned. We could potentially move where this happens to address these issues.

bfineran:
Hi @clementpoiret were you able to get unblocked locally on this?

PR #1465, opened 2023-03-21
Disable EMA for YOLOv8 during QAT

natuan:
This PR disables EMA for quantization with QAT and uses the (non-averaged) model during validation (otherwise, by the fault the EMA is used, which is not updated for QAT causing constant metrics being logged).

Another issue coming up: with this new validation, I can no longer run with multiple GPUs.

### DIFF
--- a/src/sparseml/yolov8/validators.py
+++ b/src/sparseml/yolov8/validators.py
@@ -36,7 +36,7 @@ class SparseValidator(BaseValidator):
         if self.training:
             self.device = trainer.device
             self.data = trainer.data
-            if trainer.manager.quantization_modifiers:
+            if trainer.manager and trainer.manager.quantization_modifiers:
                 # Since we disable the EMA model for QAT, we validate the non-averaged
                 # QAT model
                 model = de_parallel(trainer.model)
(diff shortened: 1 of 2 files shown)


natuan:
> @natuan what happens if you attempt to run with multiple gpus?

The training part is fine, but the validation appears to be frozen. I'm not completely sure how to fix it. I created this ticket: https://app.asana.com/0/1203126676641557/1204227804256169/f

src/sparseml/yolov8/validators.py
```diff
@@ -36,7 +36,13 @@ def __call__(self, trainer=None, model=None):
         if self.training:
             self.device = trainer.device
             self.data = trainer.data
-            model = trainer.ema.ema or trainer.model
+            if trainer.manager.quantization_modifiers:
+                # Since we disable the EMA model for QAT, we validate the non-averaged
+                # QAT model
+                model = de_parallel(trainer.model)
```
corey-nm, line 42: 👍 this was the original reason i thought to just always use trainer.model, but implemented more generally!

anmarques (dismissed):

anmarques (approved):

corey-nm (approved):

PR #1471, opened 2023-03-22
[ONNX] override_model_input_shape helper function

bfineran:
simple helper function to override an ONNX graph's input shape to a target shape

added to set a hardcoded starting shape for yolov5 graph exports with dynamic shape. (dynamic shape export is required to change image size, however deepsparse engine requires a static input shape and static input shapes provide good information for model analysis and shape inference)

**test_plan:**
manual, see yolov5 pr for extensive testing

### DIFF
--- a/src/sparseml/onnx/utils/helpers.py
+++ b/src/sparseml/onnx/utils/helpers.py
@@ -74,6 +74,7 @@ __all__ = [
     "get_tensor_shape",
     "get_tensor_dim_shape",
     "set_tensor_dim_shape",
+    "override_model_input_shape",
 ]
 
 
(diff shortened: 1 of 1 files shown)


src/sparseml/onnx/utils/helpers.py
```diff
@@ -1233,3 +1234,25 @@ def set_tensor_dim_shape(tensor: onnx.TensorProto, dim: int, value: int):
     :param value: new shape for the given dimension
     """
     tensor.type.tensor_type.shape.dim[dim].dim_value = value
+
+
+def override_model_input_shape(model: Union[str, onnx.ModelProto], shape: List[int]):
+    """
+    Set the shape of the first input of the given model to the given shape.
```
dbogunowicz, line 1241: can we be more specific that this only works for cv-like models that assume input BCHW?

bfineran: this works for any onnx model, no? shape is generic

dbogunowicz (approved):

KSGulin (approved):

PR #1478, opened 2023-03-23
[ONNX] support for setting/getting string onnx tensor shapes

bfineran:
simple fix to support setting str tensor shapes to the `dim_param` one of attribute of dim in tensor shape.

supports a YOLOv5 patch

### DIFF
--- a/src/sparseml/onnx/utils/helpers.py
+++ b/src/sparseml/onnx/utils/helpers.py
@@ -1224,7 +1224,7 @@ def get_tensor_dim_shape(tensor: onnx.TensorProto, dim: Union[int, str]) -> int:
     """
     return (
         tensor.type.tensor_type.shape.dim[dim].dim_value
-        or tensor.type.tensor_type.shape.dim[dim].dim_value
+        or tensor.type.tensor_type.shape.dim[dim].dim_param
     )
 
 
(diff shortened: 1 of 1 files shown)


src/sparseml/onnx/utils/helpers.py
```diff
@@ -1216,24 +1216,30 @@ def get_tensor_shape(tensor: onnx.TensorProto) -> List[int]:
...
+def get_tensor_dim_shape(tensor: onnx.TensorProto, dim: Union[int, str]) -> int:
     """
     :param tensor: ONNX tensor to get the shape of a dimension of
     :param dim: dimension index of the tensor to get the shape of
     :return: shape of the tensor at the given dimension
     """
-    return tensor.type.tensor_type.shape.dim[dim].dim_value
+    return (
+        tensor.type.tensor_type.shape.dim[dim].dim_value
+        or tensor.type.tensor_type.shape.dim[dim].dim_value
```
rahul-tuli, line 1227: Am I missing something or do we have the same value on both sides of `or`?

rahul-tuli (changes requested):
did we mean dim_param on line 1227?

src/sparseml/onnx/utils/helpers.py
bfineran: good catch

bfineran, line 1227:
```suggestion
        or tensor.type.tensor_type.shape.dim[dim].dim_param
```

rahul-tuli (approved):

dbogunowicz (approved):

PR #1486, opened 2023-03-28
[Yolov8] Applying pruning modifier masks to EMA model

corey-nm:
# Summary

Currently sparsification is not applied to the EMA model, even though EMA model is still active during sparsification. Notably we only deactivate the EMA model once quantization begins. This PR applies all sparsification masks to the EMA model if they exist and the modifier is active.

Note that without this, the EMA weights would end up converging to the correct values over time, since they would start decaying to 0.

# Testing plan:
1. manually verified that this applies mask to weights
2. Training completed


dbogunowicz (approved):

rahul-tuli (dismissed):

bfineran (dismissed):
LGTM @corey-nm can we move this into a helper function or maybe method of base pruning modifier? we will also need this functionality in YOLOv5 and any future EMA flows

abhinavnmagic:
Additionally, I think adding this to the callbacks is a cleaner implementation. 
Either on_train_batch_start, or on_train_batch_end should work!

abhinavnmagic:
Adding to optimizer_step increase the training time quit a bit (**2x**). We need not update the EMA at each step; just before evaluation would be much better from an efficiency perspective.

anmarques (commented):
I'm afraid that updating the mask on ema for every optimizer step is an overkill. Sure, it gets ema to a correct state, but I'd rather we did this every time we eval and/or save the model

corey-nm:
Okay yep I'll move it to before evaluation

corey-nm:
@anmarques @abhinavnmagic updated to use callback as suggested, only done once right before validation now, so it's much faster.

anmarques (approved):

abhinavnmagic (approved):

PR #1491, opened 2023-03-29
Modify YoloV8 arch for quantization

natuan:
To allow for more control in quantizing residual adds in Yolov8, this PR marks their inputs with custom identity modules that can be manipulate from the recipe. It does so by overwrite ultralytics's Bottleneck with SparseML's version.

It also overwrite ultralytics's Conv with separately instantiated SiLU to help to resolve naming issue in the recipe.

Qualification: manually verify quantization graph before training loop.


corey-nm (approved):
Nice implementation! Super clean 👍

anmarques (approved):

PR #1493, opened 2023-03-29
Fixing training-validation metric mismatch

abhinavnmagic:
~~1. Copy EMA weights to model once training finishes, following ultralytics behavior.
2. validation metrics are re-computed once the training finishes with the saved checkpoints~~

1. Load EMA weights if available in the checkpoint, else default to model weights.
2. Save EMA weights only if EMA is enabled.


corey-nm (approved):
This LGTM, assuming you tested with sparse/quantized recipe and the numbers matched? I guess I'm surprised that we don't need to update the Validator class to use trainer.model if ema is disabled to get the metrics to match

anmarques (dismissed):

corey-nm (dismissed):

anmarques (approved):

PR #1497, opened 2023-03-30
ONNX export update for YOLOv8

anmarques:
Adds a function that allows a dequantize node to be propagated down from a split node. That allows the dequantize node to be close to other ops that might be quantized after the split, and enable the proper conversion.

### DIFF
--- a/src/sparseml/yolov8/trainers.py
+++ b/src/sparseml/yolov8/trainers.py
@@ -205,22 +205,13 @@ class SparseTrainer(BaseTrainer):
         LOGGER.info("Loaded previous weights from sparseml checkpoint")
         return ckpt
 
-    def _modify_arch_for_quantization(self):
-        layer_map = {"Bottleneck": Bottleneck, "Conv": Conv}
-        for name, layer in self.model.named_modules():
-            cls_name = layer.__class__.__name__
-            if cls_name in layer_map:
-                submodule_path = name.split(".")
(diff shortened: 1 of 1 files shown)


src/sparseml/onnx/utils/helpers.py
```diff
@@ -796,7 +796,12 @@ def get_node_output_nodes(model: ModelProto, node: NodeProto) -> List[NodeProto]
     for output_id in get_node_outputs(model, node):
         nodes.extend(get_nodes_by_input_id(model, output_id))
 
-    return nodes
+    unique_nodes = []
+    for node in nodes:
+        if node not in unique_nodes:
+            unique_nodes.append(node)
+
+    return unique_nodes
```
bfineran, line 804: it this fine?
```suggestion
    return list(set(nodes))
```

bfineran (commented):
LGTM overall - will need to add this to the `ONNXToDeepSparse` converter as well

src/sparseml/onnx/utils/helpers.py
anmarques: Great suggestion!

anmarques: Had to revert change. The set construction failed w/ onnx nodes

anmarques:
> LGTM overall - will need to add this to the `ONNXToDeepSparse` converter as well

Will add ONNXToDeepSparse in a future PR

bfineran (approved):

PR #1505, opened 2023-04-03
Update RN50 for QAT

anmarques:
Introduce "no op" modules to the inputs for the residual add operation. These modules allow these inputs to be quantized by the new quantization modifier, addressing a long standing issue w/ RN50 models. Prior to this, the quantization of residual adds was only possible via a construction that created duplicate quantization nodes and hurt accuracy.

Testing plan:
1. Completed a QAT run of RN50 using the schedule of the pruned 95% model (https://sparsezoo.neuralmagic.com/models/cv%2Fclassification%2Fresnet_v1-50%2Fpytorch%2Fsparseml%2Fimagenet%2Fpruned95_quant-none) implemented via a recipe that uses the new quantization modifier:
https://wandb.ai/neuralmagic/alexandre_debug/runs/cwne0smk?workspace=user-neuralmagic

2. Evaluated the existing model from the zoo to verify that the changes do not affect the ability to load legacy models.

### DIFF
--- a/src/sparseml/pytorch/models/classification/resnet.py
+++ b/src/sparseml/pytorch/models/classification/resnet.py
@@ -140,26 +140,9 @@ class _IdentityModifier(Module):
         return in_channels != out_channels or stride > 1
 
 
-class _AddReLU(Module):
-    """
-    Wrapper for the FloatFunctional class that enables QATWrapper used to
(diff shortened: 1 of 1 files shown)


bfineran (approved):
nice

natuan (approved):

PR #1509, opened 2023-04-04
Enabling OBS training-aware and one-shot

abhinavnmagic:
This PR adds the support for running OBS on yolov8 models from main. This would enable both training-aware and one-shot pathways for OBS+yolov8.

Testing plan:
1. Run training-aware OBS for yolov8s:
a) cmd: ```sparseml.ultralytics.train --recipe ~/recipes/yolov8/obs/recipe_obs_spdyprofile_75sp_50epochs.yaml --model yolov8s.pt --batch 64 --data coco.yaml --deterministic --project abhinav_yolov8_obs --name ta-bs64-numgrads512-gpu4-spdyprofile75sp50epochs256fisher```
b) [Wandb](https://wandb.ai/neuralmagic/abhinav_yolov8_obs/runs/zzjcq2si?workspace=user-neuralmagic)

2. Run one-shot OBS for yolov8s:
a) cmd: ```sparseml.ultralytics.export --model yolov8s.pt --save-dir ~/experiments/obs-one-shot/ --one-shot <recipe> --save-one-shot-torch True"

### DIFF
--- a/src/sparseml/pytorch/optim/mask_pruning_scorer.py
+++ b/src/sparseml/pytorch/optim/mask_pruning_scorer.py
@@ -151,7 +151,7 @@ class PruningParamsGradScorer(PruningParamsScorer, ABC):
             dist.destroy_process_group(self._dist_group)
 
     def _broadcast_list_from_main(self, val: Any) -> Any:
-        if not self._is_ddp:
+        if not self._is_ddp and self._dist_group is None:
             return val
         dist.broadcast_object_list(val, src=0, group=self._dist_group)
         return val
--- a/src/sparseml/pytorch/optim/mask_pruning_scorer.py
+++ b/src/sparseml/pytorch/optim/mask_pruning_scorer.py
@@ -151,7 +151,7 @@ class PruningParamsGradScorer(PruningParamsScorer, ABC):
             dist.destroy_process_group(self._dist_group)
 
     def _broadcast_list_from_main(self, val: Any) -> Any:
-        if not self._is_ddp and self._dist_group is None:
+        if not self._is_ddp:
             return val
         dist.broadcast_object_list(val, src=0, group=self._dist_group)
         return val
--- a/src/sparseml/pytorch/sparsification/pruning/modifier_pruning_obs.py
+++ b/src/sparseml/pytorch/sparsification/pruning/modifier_pruning_obs.py
@@ -389,6 +389,8 @@ class OBSPruningParamsScorer(PruningParamsGradScorer):
         mask_type: str,
     ):
         super().__init__(params)
+        super().__init__(params, dist_backend="nccl")
+        # super().__init__(params, dist_backend="gloo")
         self._num_grads = num_grads
         self._damp = damp
         self._fisher_block_size = fisher_block_size
--- a/src/sparseml/pytorch/sparsification/pruning/modifier_pruning_obs.py
+++ b/src/sparseml/pytorch/sparsification/pruning/modifier_pruning_obs.py
@@ -388,7 +388,6 @@ class OBSPruningParamsScorer(PruningParamsGradScorer):
         fisher_block_size: int,
         mask_type: str,
     ):
-        # super().__init__(params)
         super().__init__(params, dist_backend="nccl")
         self._num_grads = num_grads
         self._damp = damp
--- a/src/sparseml/pytorch/sparsification/pruning/modifier_pruning_obs.py
+++ b/src/sparseml/pytorch/sparsification/pruning/modifier_pruning_obs.py
@@ -294,7 +294,6 @@ class OBSPruningModifier(BaseGradualPruningModifier):
                 recomputation_sparsity=recomputation_sparsity,
             )
 
-        torch.cuda.empty_cache()
         self._last_applied_sparsity = to_apply_sparsities
         if self._scorer._is_main_proc:
             self._scorer._enabled_grad_buffering = False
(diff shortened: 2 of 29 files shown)


src/sparseml/pytorch/optim/mask_pruning_scorer.py
```diff
@@ -151,7 +151,7 @@ def on_pruning_end(self):
             dist.destroy_process_group(self._dist_group)
 
     def _broadcast_list_from_main(self, val: Any) -> Any:
-        if not self._is_ddp:
+        if not self._is_ddp and self._dist_group is None:
```
KSGulin, line 154: I believe self._dist_group should be None any time `self._is_ddp` is off

src/sparseml/pytorch/sparsification/pruning/modifier_pruning_obs.py
```diff
@@ -388,7 +387,7 @@ def __init__(
         fisher_block_size: int,
         mask_type: str,
     ):
-        super().__init__(params)
+        super().__init__(params, dist_backend="nccl")
```
KSGulin, line 390: This doesn't seem to be used anywhere. Guessing it's a dev relic?

src/sparseml/pytorch/utils/sparsification.py
```diff
@@ -46,6 +47,8 @@
     "GradSampler",
 ]
 
+logger = logging.getLogger()
```
KSGulin, line 50: nit
```suggestion
_LOGGER = logging.getLogger(__name__)
```

src/sparseml/pytorch/utils/sparsification.py
```diff
@@ -254,4 +257,11 @@ def iter_module_backwards(
                         pbar.update(1)
                     if computed_grads >= num_grads:
                         break
+                if computed_grads < num_grads:
+                    logger.warning(
```
KSGulin, line 261: To be applied with the nit above
```suggestion
                    _LOGGER.warning(
```

KSGulin (commented):
Looks good! Just a few minor nits. Could you also update the PR with a description. Would OBS be usable from main if this lands, or is it still run from a research branch? If the former, would be nice to have an example command as well.

src/sparseml/yolov8/trainers.py
```diff
@@ -322,11 +322,58 @@ def _setup_train(self, rank, world_size):
...
+        self.device = torch.device("cuda", rank)
+        LOGGER.info(
+            f"DDP settings: RANK {rank}, WORLD_SIZE {world_size}, DEVICE {self.device}"
+        )
+        torch.distributed.init_process_group(
+            "nccl" if torch.distributed.is_nccl_available() else "gloo",
+            rank=rank,
+            world_size=world_size,
+            timeout=timedelta(seconds=7200),
+        )
```
corey-nm, line 348: Is this used anywhere?

corey-nm (approved):
lgtm

src/sparseml/pytorch/sparsification/pruning/modifier_pruning_obs.py
abhinavnmagic: Specifying `dist_backend` enables creation of `self._dist_group`. This results in initializing a separate group for scorer. Without this, `self._dist_group` is set to None.

src/sparseml/yolov8/trainers.py
abhinavnmagic: Yes. This is used by ultralytics to set up DDP training. The only change wrt ultralytics implementation is the `timeout` increase from 1800s to 7200s. This helps avoid timeout errors with scorer.
I will add in a comment to make this explicit.

src/sparseml/pytorch/sparsification/pruning/modifier_pruning_obs.py
KSGulin: Ah yeah I misread the code

src/sparseml/yolov8/trainers.py
corey-nm: Ahhh gotcha, thanks!

corey-nm (approved):
👍

KSGulin (approved):

src/sparseml/pytorch/utils/sparsification.py
eldarkurtic: @abhinavnmagic I think we don't need this check because our dataloader are infinite, i.e. they will just cycle and start loading from beginning

abhinavnmagic: Yes, we know thats true for the current dataloader for yolov8.
It's a good practice to keep this check as we expand to other datasets, dataloaders or tasks.

eldarkurtic: But this is not specific to Yolov8, it is the case for all other projects as well (e.g. all our current NLP integrations) have infinite dataloaders because we create them like that intentionally when building dataloaders for gradient-sampling. 

I think this feature of breaking when `computed_grads < num_grads` has three major problems that we should consider before introducing these changes:
1.  it is not backward compatible, thus all of our OBS/MFAC examples for NLP models might start crashing 
2. it might be very annoying when working with NLP datasets, because majority of GLUE tasks are extremely small and this will require from users (and us as well) to iterate a few times with different values of `num_grads` until we hit the correct value for each dataset
3. our recipes won't be consistent over different datasets, and for example, we won't be able to reuse a recipe from MNLI (large GLUE dataset) on SST2 (small GLUE dataset)

abhinavnmagic: The implementation doesn't break when `computed_grads < num_grads`. All it does is that if the for loop finishes before collecting the required number of gradients (which it shouldn't since we have infinite dataloaders), it issues a warning to the user. That's it. There is no change in the runtime behavior of the program.
It's important to note that this warning is outside the gradient collection for loop. I guess it's really to miss from the way GitHub shows diffs.

PR #1514, opened 2023-04-10
 Support exporting > 2Gb transformer models

dbogunowicz:
Since are now supporting big ONNX models, we are:
- enforcing that when exporting such a model, its weights will be dumped to `model.data` file
- setting `tokenizer.json` to an optional file

The rationale for making `tokenizer.json` optional was explained by @natuan in the thread a while ago.


src/sparseml/pytorch/utils/exporter.py
```diff
@@ -570,7 +570,23 @@ def export_onnx(
 
         # clean up graph from any injected / wrapped operations
         _delete_trivial_onnx_adds(onnx_model)
-    onnx.save(onnx_model, file_path)
+
+    try:
+        # alternatively we can just check the size of the onnx file on disk
```
dbogunowicz, line 575: To decide whether we should save the model with/without external data we could either check its disk size (feels over-engineered) or simply check whether we fail on saving all the data directly to protobuf. If yes, we switch to exporting with the external data file.

src/sparseml/transformers/export.py
```diff
@@ -91,12 +91,14 @@
 __all__ = ["export_transformer_to_onnx", "load_task_model"]
 
 MODEL_ONNX_NAME = "model.onnx"
```
dbogunowicz, line 93: Introducing the following convention for creating the `deployment` folder:
Files that need to be copied over: `model.onnx`, 'tokenizer_config.json`, 'config.json`. 
Files that can be copied will only raise a warning, not an error when absent: `model.data` (because not every model is >2Gb), `tokenizer.json` because we have seen many times that this file is not mandatory for all the transformer-like models. Confirming this with Tuan.

src/sparseml/transformers/export.py
```diff
@@ -91,12 +91,14 @@
...
-DEPLOYMENT_FILES: List[str] = [
+EXTERNAL_ONNX_DATA_NAME = "model.data"
```
dbogunowicz, line 94: I wonder whether we should dump all the information into a single `data` file, or maybe have multiple files, one per parameter - this is what torch.onnx.save(...) does. However, given that we want to minimize the number of files per stub in sparsezoo, sticking to one file is probably the way to go. This is also how `optimum` does this: they give you onnx model and the respective single data file.

src/sparseml/pytorch/utils/exporter.py
```diff
@@ -570,7 +570,23 @@ def export_onnx(
 
         # clean up graph from any injected / wrapped operations
         _delete_trivial_onnx_adds(onnx_model)
-    onnx.save(onnx_model, file_path)
```
dbogunowicz, line 573: Perhaps we can already to saving to external formats in line L557. There are a couple of things to clarify:
- is there API support (`torch.onnx.export(model, (input, ), './large_model.onnx', use_external_data_format=True)` only available in torch >= 1.15)
- we will need to resave the weights after the "onnx file fixes" L563 anyway, so maybe not worth the trouble.

bfineran (approved):

src/sparseml/transformers/export.py
```diff
@@ -113,7 +114,6 @@ def load_task_model(task: str, model_path: str, config: Any) -> Module:
         return SparseAutoModel.question_answering_from_pretrained(
             model_name_or_path=model_path,
             config=config,
```
dbogunowicz, line 116:
```suggestion
            config=config,
            model_type="model",
```

src/sparseml/yolov8/trainers.py
```diff
@@ -729,6 +729,7 @@ def export(self, **kwargs):
             pass
 
         validate_onnx(complete_path)
+
```
dbogunowicz, line 732:
```suggestion
```

KSGulin (approved):

PR #1520, opened 2023-04-13
[YOLOv8] Fix support for `--dataset_dir` argument

dbogunowicz:
This PR: #1513 was not tested rigorously and was failing for all the pathways. This PR fixes this.

Testing:
- tested validation pathway
- tested export pathway (generating samples and using the data for calculating gradients at one-shot)
- tested training pathway

### DIFF
--- a/src/sparseml/yolov8/export.py
+++ b/src/sparseml/yolov8/export.py
@@ -80,10 +80,10 @@ from sparseml.yolov8.trainers import SparseYOLO
     help="cuda device, i.e. 0 or 0,1,2,3 or cpu. Device to run on",
 )
 @click.option(
-    "--datasets-dir",
+    "--dataset_path",
     type=str,
     default=None,
-    help="Path to override default datasets dir.",
+    help="Path to override default dataset path.",
 )
 def main(**kwargs):
 
--- a/src/sparseml/yolov8/export.py
+++ b/src/sparseml/yolov8/export.py
@@ -80,7 +80,7 @@ from sparseml.yolov8.trainers import SparseYOLO
     help="cuda device, i.e. 0 or 0,1,2,3 or cpu. Device to run on",
 )
 @click.option(
-    "--dataset_path",
+    "--dataset-path",
     type=str,
     default=None,
     help="Path to override default dataset path.",
--- a/src/sparseml/yolov8/export.py
+++ b/src/sparseml/yolov8/export.py
@@ -12,11 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
 
 import click
 from sparseml.yolov8.trainers import SparseYOLO
-from ultralytics.yolo.utils import USER_CONFIG_DIR, get_settings, yaml_save
 
 
 # Options generated from
@@ -88,12 +86,6 @@ from ultralytics.yolo.utils import USER_CONFIG_DIR, get_settings, yaml_save
     help="Path to override default datasets dir.",
 )
 def main(**kwargs):
-    if kwargs["datasets_dir"] is not None:
-        settings = get_settings()
-        settings["datasets_dir"] = os.path.abspath(
-            os.path.expanduser(kwargs["datasets_dir"])
-        )
-        yaml_save(USER_CONFIG_DIR / "settings.yaml", settings)
 
     model = SparseYOLO(kwargs["model"])
     model.export(**kwargs)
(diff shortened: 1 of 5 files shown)


src/sparseml/yolov8/val.py
```diff
@@ -74,16 +73,12 @@
 @click.option(
     "--datasets-dir",
     type=str,
-    default=None,
+    default="/home/ubuntu/damian/sparseml/funny_dir",
```
anmarques, line 76: Maybe leave default as None 😄

anmarques (changes requested):
Looks good to me. I just recommend replacing --datasets-dir by --dataset-path where the later refers to the complete path to the dataset rather than it's parent directory

bfineran (dismissed):

bfineran:
LGTM pending comment

src/sparseml/yolov8/val.py
dbogunowicz, line 76:
```suggestion
    default=None,
```

src/sparseml/yolov8/export.py
```diff
@@ -82,18 +80,12 @@
     help="cuda device, i.e. 0 or 0,1,2,3 or cpu. Device to run on",
 )
 @click.option(
-    "--datasets-dir",
+    "--dataset_path",
```
anmarques, line 83: Let's use --dataset-path, making it consistent w/ other arguments (dash instead of underscore)

src/sparseml/yolov8/train.py
```diff
@@ -212,18 +211,15 @@
     "--copy-paste", type=float, default=0.0, help="segment copy-paste (probability)"
 )
 @click.option(
-    "--datasets-dir",
+    "--dataset_path",
```
anmarques, line 214: Let's use --dataset-path, making it consistent w/ other arguments (dash instead of underscore)

src/sparseml/yolov8/utils/helpers.py
```diff
@@ -69,3 +77,42 @@ def create_grad_sampler(
...
+    and finally save it to the current working directory.
+    This allows to create load data yaml config files that point
+    to the arbitrary directories on the disk.
+
+     :param data: name of the dataset (e.g. "coco.yaml")
+     :param dataset_path: path to the dataset directory
+     :return: a path to the new yaml config file
+        (saved in the current working directory)
+    """
+    if os.path.basename(dataset_path) != os.path.splitext(data)[0]:
```
anmarques, line 96: We don't have to impose this constraint. For instance, I doing something like ~/datasets/my_coco should still be acceptable with coco.yaml.

The dataset path doesn't have to be related to the name of the dataset

src/sparseml/yolov8/val.py
```diff
@@ -72,18 +71,15 @@
 )
 @click.option("--plots", default=False, is_flag=True, help="show plots during training")
 @click.option(
-    "--datasets-dir",
+    "--dataset_path",
```
anmarques, line 74: Let's use --dataset-path, making it consistent w/ other arguments (dash instead of underscore)

anmarques (changes requested):

src/sparseml/yolov8/train.py
dbogunowicz, line 214:
```suggestion
    "--dataset-path",
```

src/sparseml/yolov8/export.py
dbogunowicz, line 83:
```suggestion
    "--dataset-path",
```

anmarques (approved):

src/sparseml/yolov8/utils/helpers.py
```diff
@@ -69,3 +77,36 @@ def create_grad_sampler(
...
+    from the Ultralytics dataset repo, overwrite its 'path'
+    attribute (dataset root dir) to point to the `dataset_path`
+    and finally save it to the current working directory.
+    This allows to create load data yaml config files that point
+    to the arbitrary directories on the disk.
+
+     :param data: name of the dataset (e.g. "coco.yaml")
+     :param dataset_path: path to the dataset directory
+     :return: a path to the new yaml config file
+        (saved in the current working directory)
```
rahul-tuli, line 94: Is there an extra space here?

rahul-tuli (approved):

PR #1521, opened 2023-04-13
Convert input for quantized YOLOv8

anmarques:
Added call to function that skips the quantization of the input if the model is quantized. After that it expects the input to be quantized.

Testing plan:
Exported YOLOv8s dense quantized. Verified that the input is quantized and the QuantizationLinear node is removed. Tested for accuracy and latency on deepsparse.

### DIFF
--- a/src/sparseml/yolov8/trainers.py
+++ b/src/sparseml/yolov8/trainers.py
@@ -28,6 +28,7 @@ import torch
 
 from sparseml.optim.helpers import load_recipe_yaml_str
 from sparseml.pytorch.optim.manager import ScheduledModifierManager
+from sparseml.pytorch.sparsification.quantization import skip_onnx_input_quantize
 from sparseml.pytorch.utils import ModuleExporter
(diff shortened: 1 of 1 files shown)


bfineran (dismissed):

src/sparseml/yolov8/trainers.py
```diff
@@ -711,7 +712,13 @@ def export(self, **kwargs):
             else ["output0"],
         )
 
-        onnx.checker.check_model(os.path.join(save_dir, name))
+        complete_path = os.path.join(save_dir, name)
+        try:
+            skip_onnx_input_quantize(complete_path, complete_path)
+        except Exception:
+            pass
```
natuan, line 719: @anmarques If the skip func above throws runtime error, wouldn't the check_model below fails? Is it safe to pass the exception?

anmarques: The Exception object contemplates RuntimeError

bfineran (approved):

mgoin (approved):

PR #1522, opened 2023-04-13
Fixing yolov8-seg validation

abhinavnmagic:
The PR enables segmentation evaluation for yolov8 models, which currently fails. It implements a function that extends `detach` to PyTorch expected types ie. list or tuples.

Testing:
- Evaluation works for both detection and segmentation models obtained form either SparseML or ultraltyics

### DIFF
--- a/src/sparseml/yolov8/utils/helpers.py
+++ b/src/sparseml/yolov8/utils/helpers.py
@@ -15,14 +15,16 @@
 import os
 import warnings
 from argparse import Namespace
-from typing import Any, Dict
+from typing import Any, Dict, List, Tuple, Union
+
+import torch
 
 from ultralytics.yolo.data.dataloaders.v5loader import create_dataloader
 from ultralytics.yolo.engine.model import DetectionModel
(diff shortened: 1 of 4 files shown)


src/sparseml/yolov8/utils/helpers.py
```diff
@@ -69,3 +71,14 @@ def create_grad_sampler(
         / train_loader.batch_size,
     )
     return grad_sampler
+
+
+def detach(x: Union[torch.Tensor, List, Tuple]):
```
natuan, line 76: Would it be better to move this func into sparseml/pytorch/utils/helpers.py so it could be reused more?

natuan (changes requested):

src/sparseml/yolov8/utils/helpers.py
abhinavnmagic: Yes, definitely. If we believe it would reused, that's a better place.

natuan (approved):

dbogunowicz (approved):

PR #1523, opened 2023-04-18
[Bug-fix] YOLOv5 integration test output pathing

KSGulin:
There is currently a failing post-commit yolov5 integration test caused by an incorrect pathing assumption in the integration test. This PR fixes it by updating the final output path to refer to an `.onnx` file rather than a `.pt` file

**Test plan**
Tested locally with `make testinteg TARGETS=yolov5 SPARSEML_TEST_CADENCE=commit`


rahul-tuli (approved):
Pathlib is so nice!

bfineran (approved):

PR #1539, opened 2023-04-26
Fix ONNX export for MobileBERT

anmarques:
Simple fixes for ONNX export of MobileBERT.

Before the fix, a MatMul in the Embedding section of MobileBERT was not being converted to MatMulInteger, even though the inputs are quantized.

In short, a DequantizeLinear node used as part of the embedding quantization must be propagated down a few Slice and Concat nodes such that it can sit next to the MatMul node. This allows the proper pattern matching to convert that MatMul to MatMulInteger.

This behavior was already present in the onnx export logic, but there is a data type check on the embedding weights and it expected uint8. However, the weights were defined as int8 (conversion to uint8 happens at a later step). This PR adds logic to support int8, and also accounts for non-zero zero-point.

**Testing plan:**
Did the onnx export for the model below and checked that the MatMul is converted to ConvInteger. Checked that deepsparse now supports 99.35% of ops. Accuracy matches value reported in the zoo.

zoo:nlp/question_answering/mobilebert-none/pytorch/huggingface/squad/14layer_pruned50_quant-none-vnni

### DIFF
--- a/src/sparseml/pytorch/sparsification/quantization/quantize_qat_export.py
+++ b/src/sparseml/pytorch/sparsification/quantization/quantize_qat_export.py
@@ -1761,7 +1761,7 @@ def _propagate_mobilebert_embedding_quantization(model: ModelProto):
             continue
 
         embedding_array = numpy_helper.to_array(embedding_initializer)
-        if embedding_array.dtype != numpy.uint8:
+        if embedding_array.dtype not in [numpy.uint8, numpy.int8]:
             continue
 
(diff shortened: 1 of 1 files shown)


eldarkurtic:
I have just tested export to onnx and eval-downstream of the exported model, and everything looks as expected

eldarkurtic (approved):

mgoin (approved):
Excellent fix, thanks!

PR #1544, opened 2023-05-02
[Transformers] Allow for model-less dataset creation

KSGulin:
Currently, a pytorch model is required to create a text-classification or token-classification dataset. This PR removes the requirement and allows instead a config to be used directly.

Test plan
Local testing with DeepSparse deployment folder

### DIFF
--- a/src/sparseml/transformers/text_classification.py
+++ b/src/sparseml/transformers/text_classification.py
@@ -705,7 +705,7 @@ def _get_tokenized_and_preprocessed_raw_datasets(
     ) = _get_label_info(data_args, raw_datasets)
 
     train_dataset = predict_dataset = eval_dataset = None
-    config = config if model else config
+    config = model.config if model else config
     if not main_process_func:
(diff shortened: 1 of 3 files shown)


src/sparseml/transformers/text_classification.py
```diff
@@ -705,6 +705,7 @@ def _get_tokenized_and_preprocessed_raw_datasets(
     ) = _get_label_info(data_args, raw_datasets)
 
     train_dataset = predict_dataset = eval_dataset = None
+    config = config if model else config
```
rahul-tuli, line 708: Did we mean to do this maybe?

```suggestion
    config = model.config if model else config
```

bfineran (commented):
LGTM pending @rahul-tuli's coment

src/sparseml/transformers/text_classification.py
KSGulin: Ah shoot good catch

bfineran (approved):

rahul-tuli (approved):
Gucci!

PR #1556, opened 2023-05-10
allow patch versions of torchvision 0.14 beyond .0

bfineran:
as written, only 0.14.0 would be allowed


bfineran:
tested locally, merging

PR #1568, opened 2023-05-18
Choose quantization modifier based on dynamic set of modifier properties

KSGulin:
This PR makes the selection between LegacyQuantizationModifier and QuantizationModifier more robust by basing it on the full difference in modifier props between the two modifiers.

**Test plan**
Tested locally with a QA training and export script


bfineran (approved):

rahul-tuli (approved):
LGTM!

PR #1569, opened 2023-05-18
update get_named_layers_and_params_by_regex in src/sparseml/pytorch/u…

ohaijen:
Update get_named_layers_and_params_by_regex in src/sparseml/pytorch/utils/helpers.py to still match in the DDP training case, where module. is prepended to layer names.  I came across this issue when trying to run the suggested training command in src/sparseml/pytorch/torchvision on multiple GPUs.


jeanniefinks:
Per the main [README announcement](https://github.com/neuralmagic/sparseml/blob/main/README.md), SparseML is being deprecated by early June 2, 2025. Closing the PR as work has been suspended; thank you for the inputs and support!

PR #1570, opened 2023-05-19
Scipy version depending on python

natuan:
Without this change, I got the following error during installation:

ERROR: scikit-image 0.20.0 has requirement scipy<1.9.2,>=1.8; python_version <= “3.9”, but you’ll have scipy 1.10.1 which is incompatible.


bfineran (approved):
nice, wasn't aware of this syntax!

dbogunowicz (approved):

PR #1571, opened 2023-05-19
Export files required by OPT to load the tokenizer

dbogunowicz:
OPT does not contain a "tokenizer.json", but instead requires the following files:
"special_tokens_map.json"
"vocab.json"
"merges.txt"
to create the tokenizer from pretrained.

I.e if those three files are present in the `deployment` model this line of code will execute:
```python
model_path = "deployment"
tokenizer = AutoTokenizer.from_pretrained(
            model_path,
        )
```


src/sparseml/transformers/export.py
```diff
@@ -97,7 +97,14 @@
     "tokenizer_config.json",
     "config.json",
 ]
-OPTIONAL_DEPLOYMENT_FILES: List[str] = [EXTERNAL_ONNX_DATA_NAME, "tokenizer.json"]
+
+OPTIONAL_DEPLOYMENT_FILES: List[str] = [
+    EXTERNAL_ONNX_DATA_NAME,
+    "tokenizer.json",
+    "special_tokens_map.json",
```
natuan, line 104: to make this self-explains, I would put these 3 new files under a separate list with hint that it's for OPT model, before you do the merge.

natuan (changes requested):

src/sparseml/transformers/export.py
dbogunowicz: Crossed my mind, makes sense.

bfineran: agree

src/sparseml/transformers/export.py
```diff
@@ -91,13 +91,18 @@
...
     "tokenizer_config.json",
     "config.json",
 ]
-OPTIONAL_DEPLOYMENT_FILES: List[str] = [EXTERNAL_ONNX_DATA_NAME, "tokenizer.json"]
+EXTERNAL_ONNX_DATA_NAME = ["model.data"]
+OPT_TOKENIZER_FILES = ["special_tokens_map.json", "vocab.json", "merges.txt"]
+
+OPTIONAL_DEPLOYMENT_FILES: List[str] = []
+OPTIONAL_DEPLOYMENT_FILES.extend(EXTERNAL_ONNX_DATA_NAME)
+OPTIONAL_DEPLOYMENT_FILES.extend(OPT_TOKENIZER_FILES)
```
anmarques, line 104: tokenizer.json is no longer in the list

src/sparseml/transformers/export.py
```diff
@@ -91,13 +91,18 @@
...
+MANDATORY_DEPLOYMENT_FILES = [
     MODEL_ONNX_NAME,
     "tokenizer_config.json",
     "config.json",
 ]
-OPTIONAL_DEPLOYMENT_FILES: List[str] = [EXTERNAL_ONNX_DATA_NAME, "tokenizer.json"]
+EXTERNAL_ONNX_DATA_NAME = ["model.data"]
+OPT_TOKENIZER_FILES = ["special_tokens_map.json", "vocab.json", "merges.txt"]
+
+OPTIONAL_DEPLOYMENT_FILES: List[str] = []
```
dbogunowicz, line 102:
```suggestion
OPTIONAL_DEPLOYMENT_FILES: List[str] = ["tokenizer.json"]
```

anmarques (approved):

natuan (approved):

PR #1573, opened 2023-05-24
[KV cache] Properly set the static dimensions of the kv cache inputs/outputs

dbogunowicz:
This PR enforces that the kv cache input/outputs injected into the model have all their "static" dimensions already set to the appropriate values.

E.g.
for OPT, the kv_cache inputs have dimensions 
`["num_attention_heads", "sequence_len", "kv_cache_hidden_dims"]`.
Regardless of the sequence length that will be eventually set in the deepsparse pipeline, the first and last dimensions of this input will always be constant (this is specified by the architecture type, not the length of the input sequence). 

This is why we can set the dimensions in advance to 
`16, "sequence_len", 64]` (for opt-350m). 

Example use:
```python
model = onnx.load("deployment/model.onnx", load_external_data=False)
model = KeyValueCacheInjector(model_path = "deployment").apply(model)
model.graph.input[5]
```
```bash
name: "past_key_values.1.value"
type {
  tensor_type {
    elem_type: 1
    shape {
      dim {
        dim_value: 16
      }
      dim {
        dim_param: "past_sequence_len"
      }
      dim {
        dim_value: 64
      }
    }
  }
}
```


KSGulin (approved):
LGTM! As discussed, generic support for getting number of attention heads and hidden size from config should be added with the CodeGen PRs

src/sparseml/exporters/kv_cache_injector.py
```diff
@@ -43,44 +43,82 @@ def __init__(
...
         :param inplace: If True, the model will be modified in place.
             If False, a copy of the model will be made and modified.
         """
         self.inplace = inplace
-        if model_type.lower() not in _SUPPORTED_ARCHITECTURES:
+        self.config = self.get_config(model_path)
+
+        if not self.config["model_type"] in _SUPPORTED_ARCHITECTURES:
             raise ValueError(
-                f"`model_type` must be one of {_SUPPORTED_ARCHITECTURES}, "
```
bfineran, line 71: let's instead warn and just set to a dynamic value so as to not block users who want to try new models

src/sparseml/exporters/kv_cache_injector.py
```diff
@@ -43,44 +43,82 @@ def __init__(
...
+        model_path = Path(model_path) if isinstance(model_path, str) else model_path
+
+        if not model_path.is_dir():
+            raise ValueError(
+                f"`model_path` is expected to be a directory, found {model_path}"
+            )
+        config_file = [
+            file for file in model_path.iterdir() if file.name == "config.json"
+        ]
+        if len(config_file) != 1:
```
bfineran, line 111: how would this happen?

src/sparseml/exporters/transforms/kv_cache/cache_keys_and_values.py
```diff
@@ -102,6 +108,11 @@ class CacheKeysAndValues(OnnxTransform):
 
     """
 
+    def __init__(self, num_attention_heads: int, hidden_size_kv_cache: int):
```
bfineran, line 111: let's consider defaulting to dynamic as discussed above

bfineran (changes requested):

src/sparseml/exporters/kv_cache_injector.py
dbogunowicz: Removed this `if-` statement and defaulted to logging. My intention here was to have some kind of check so that the transformation does not fail silently on the config ingestion.

src/sparseml/exporters/kv_cache_injector.py
dbogunowicz: Good idea.

src/sparseml/exporters/kv_cache_injector.py
```diff
@@ -108,13 +111,11 @@ def get_config(self, model_path: Union[str, Path]) -> Dict[str, Any]:
...
             file for file in model_path.iterdir() if file.name == "config.json"
         ]
-        if len(config_file) != 1:
-            raise ValueError(
-                f"Expected to find exactly one config.json file in {model_path}, "
-                f"found {len(config_file)}"
-            )
+        config_file = config_file[0]
+
+        _LOGGER.warn(f"Found config file {config_file}")
```
dbogunowicz, line 116: not warn but info

src/sparseml/exporters/kv_cache_injector.py
```diff
@@ -0,0 +1,138 @@
...
+_LOGGER = logging.getLogger(__name__)
+
+_SUPPORTED_ARCHITECTURES = ["opt"]
+
+
+class KeyValueCacheInjector(BaseExporter):
+    def __init__(
+        self,
+        model_path: str,
+        inplace: bool = True,
```
dbogunowicz, line 40: kwargs = {num_attention_heads, hidden_kv_cache_dims} (optional)

PR #1586, opened 2023-05-27
Keep model in CPU during ONNX export

natuan:
This is to fix OOM issue when exporting 13b OPT model.

### DIFF
--- a/src/sparseml/transformers/export.py
+++ b/src/sparseml/transformers/export.py
@@ -75,10 +75,12 @@ import logging
 import math
 import os
 import shutil
+from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


dbogunowicz (approved):
Looks good to me. Just wondering, why:
`model.cpu()` did not work in the first place? Was the `Trainer` ignoring this piece of logic?

natuan:
> Looks good to me. Just wondering, why: `model.cpu()` did not work in the first place? Was the `Trainer` ignoring this piece of logic?

It's too late at that point

anmarques (approved):

PR #1589, opened 2023-05-29
[KV cache] Input/output KV cache to include `batch` dimension.

dbogunowicz:


KSGulin (approved):
LGTM

src/sparseml/exporters/transforms/kv_cache/cache_keys_and_values.py
```diff
@@ -200,40 +202,46 @@ def create_cache(
...
+        hidden_size_kv_cache,
+    ]
     cache_output_dims = [
+        batch_size,
         num_attention_heads,
         "past_sequence_len + 1",
         hidden_size_kv_cache,
     ]
 
+    cache_input_name_reshaped = f"{cache_input_name}_reshaped"
```
bfineran, line 222: getting a bit long now - let's move this cache reshape node(s) injection to a helper function

src/sparseml/exporters/transforms/kv_cache/cache_keys_and_values.py
```diff
@@ -255,18 +263,66 @@ def create_cache(
         cache_output_dims,
     )
 
+    graph = ONNXGraph(model)
+
+    if node.op_type == "QLinearMatMul" and cache_input_idx == 1:
+        cache_input_idx = 3  # QLinearMatMul B matrix is at idx 3, not 1
+
+    cache_parent = graph.get_node_single_parent(node, index=cache_input_idx)
+    if isinstance(cache_parent, NodeProto) and cache_parent.op_type == "Transpose":
```
bfineran, line 272: does this guarantee the cache input dims will be in the exact expected order of the pipeline/engine?

bfineran (changes requested):

src/sparseml/exporters/transforms/kv_cache/cache_keys_and_values.py
dbogunowicz: Yes, intended do it on the CodeGen KV Cache PR, but it fits better here.

src/sparseml/exporters/transforms/kv_cache/cache_keys_and_values.py
dbogunowicz: i think so. where do you see any inconsistencies?

PR #1590, opened 2023-05-29
[KV Cache] Support for CodeGen

dbogunowicz:
The implementation of the KV Cache injection for the CodeGen model.
The model that was used for the injection is exported using `transformers==4.30.2` (anticipating the incoming internal transformers upgrade)

## Feature Preview
see: https://github.com/neuralmagic/deepsparse/pull/1078

## Notes:
- Decided to add a `KVCacheConfig` class for a clearer interface for setting up KV Cache injection parameter. I like how it makes things more structured, but I see potential disadvantages of its presence. Something I'd definitely like to talk about.
- Made `CacheKeysAndValues` transformation more robust and extendable. 
- With this PR in place, the integration of the BLOOM model should not take more than a few hours, all the building blocks are there.
- The transformations for injecting `positions` input are lightweight, they share components such as `add_position_input` methods or a helper function for deleting orphaned nodes.

#### Positions

Range gets replaced with the `positions` input

##### Without KV Cache Injection

image

##### With KV Cache Injection
image

##### Without KV Cache Injection

image

##### With KV Cache Injection

image

##### With KV Cache Injection

image

##### Without KV Cache Injection

image

##### With KV Cache Injection
("Add" node is required, see: https://github.com/huggingface/transformers/blob/main/src/transformers/models/opt/modeling_opt.py#LL105C32-L105C32)
image

##### Without KV Cache Injection

image

## Testing Plan

- [x] Test that the generated model gives the correct output in the pipeline (CodeGen)
- [x] Test that the generated model gives correct output in the pipeline (OPT)
- [x] Test the pathway, where the user provides kwargs instead of model_path to the `KVCacheInjector` transformation.


src/sparseml/exporters/transforms/kv_cache/position_embeddings_adjustment_codegen.py
```diff
@@ -0,0 +1,450 @@
...
+        if not position_embeddings_nodes:
+            pattern_not_found = _ROTARY_POSITION_EMBEDDING_PATTERN
+        if pattern_not_found:
+            raise RuntimeError(
+                f"Unable to find nodes with that match the expected pattern {pattern_not_found}"
+            )
+        return position_embeddings_nodes, rotary_position_embedding_nodes
+
+    @classmethod
+    def add_positions_input(cls, model: ModelProto) -> ModelProto:
```
dbogunowicz, line 110: Think about the potential way to share it across LLMs: unify `position_embeddings...` classes

src/sparseml/exporters/transforms/kv_cache/position_embeddings_adjustment_codegen.py
```diff
@@ -0,0 +1,450 @@
...
+
+        ```
+        def fixed_pos_embedding(x, seq_dim=1, seq_len=None):
+            dim = x.shape[-1]
+            if seq_len is None:
+                seq_len = x.shape[seq_dim]
+            inv_freq = 1.0 / (10000 ** (torch.arange(0, dim, 2) / dim))
+            sinusoid_inp = (
+                torch.einsum("i , j -> i j",
+                torch.arange(seq_len, dtype=torch.float), # This is the POSITION EMBEDDING
```
dbogunowicz, line 145: Positions, not embeddings

src/sparseml/exporters/transforms/kv_cache/position_embeddings_adjustment_codegen.py
```diff
@@ -0,0 +1,450 @@
...
+        without the kv_cache we apply the following slicing to the `duplicate_interleave` output:
+        [None, 0: length_current, None, :]
+
+        however, what we want to do after kv cache injection is to apply the following slicing:
+        [None, length_cache: length_current + length_cache, None, :]
+        e.g.
+        [None, 5: 6, None, :]
+
+        Note: for now I have decided to set the:
+        - offset (length_cache)
```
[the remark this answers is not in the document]
dbogunowicz: Add relevant screenshots from Netrone!

src/sparseml/exporters/kv_cache_injector.py
```diff
@@ -71,25 +78,31 @@ def __init__(
         self.inplace = inplace
         self.config = self.get_config(model_path)
 
-        if not self.config["model_type"] in _SUPPORTED_ARCHITECTURES:
+        if self.config["model_type"] not in PositionEmbeddingTransformation.__members__:
```
bfineran, line 81: as mentioned in #1573 let's enable some default behavior here

src/sparseml/exporters/kv_cache_injector.py
```diff
@@ -71,25 +78,31 @@ def __init__(
...
-        hidden_size_kv_cache = self.config["hidden_size"] // num_attention_heads
-
+        # TODO: Employ some elegant, unified logic that will work for both OPT and Codegen (and future LLMs)
+        # (probably copy some logic from AutoConfig.from_pretrained()
+        if self.config["model_type"] == "opt":
+            num_attention_heads = self.config["num_attention_heads"]
+            hidden_size_kv_cache = self.config["hidden_size"] // num_attention_heads
+        else:
+            num_attention_heads = self.config["n_head"]
+            hidden_size_kv_cache = self.config["n_embd"] // num_attention_heads
```
bfineran, line 94: let's move to a helper function like `_get_num_attention_heads_and_hidden_size`

src/sparseml/exporters/transforms/kv_cache/cache_keys_and_values.py
```diff
@@ -206,41 +211,66 @@ def create_cache(
...
+            "past_sequence_len",
+            hidden_size_kv_cache,
+        ]
+        cache_output_dims = [
+            batch_size,
+            num_attention_heads,
+            "past_sequence_len + 1",
+            hidden_size_kv_cache,
+        ]
+    else:
```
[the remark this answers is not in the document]
bfineran: elaborating this is a must because the runtime is going to expect/need consistency

src/sparseml/exporters/transforms/kv_cache/cache_keys_and_values.py
```diff
@@ -206,41 +211,66 @@ def create_cache(
...
+    For the sake of experimentation with CodeGen, enforcing some 
+    particular assumptions about the structure of the KV Cache:
+    - KV Cache is 4D (this is the expected shape for the MatMul operation)
+    - The key tensors has the following shape: 
+        [batch_size, num_attention_heads, past_sequence_len, hidden_size_kv_cache]
+    - The value tensors has the following shape:
+        [batch_size, past_sequence_len, num_attention_heads, hidden_size_kv_cache]
+    This can all be generalized later, when we make the KV Cache work for Codegen
+    """
+    if cache_input_name.endswith("key"):
```
bfineran, line 225: what does this condition mean?

src/sparseml/exporters/transforms/kv_cache/cache_keys_and_values.py
```diff
@@ -206,41 +211,66 @@ def create_cache(
...
-    )
-
-    reshape_out_initializer = numpy_helper.from_array(
-        numpy.array(
-            [batch_size, num_attention_heads, -1, hidden_size_kv_cache],
-            dtype=numpy.int64,
-        ),
-        reshape_out_initializer_name,
-    )
+    # reshape_in_initializer = numpy_helper.from_array(
```
bfineran, line 259: nit: cleanup

bfineran: ignoring through line 346 for now due to code state

src/sparseml/exporters/transforms/kv_cache/cache_keys_and_values.py
```diff
@@ -356,19 +386,50 @@ def _find_key_value_matmul_pairs(
...
+                raise ValueError(
+                    "While traversing the graph to find a Softmax that precedes "
+                    f"the candidate for a `value` MatMul: {node.name}, found a node "
+                    f"with multiple parents {parent.name}. "
+                    "It is assumed that the graph that connects the Softmax "
+                    "node and the `value` MatMul node is a linear chain of nodes "
+                    "and thus none of the encountered nodes should have multiple "
+                    "parents"
+                )
+            parent = graph.get_node_parents(parent)[0]
```
bfineran, line 431: can use `get_node_single_parent` - additionally why make another call in line 431?

src/sparseml/exporters/transforms/kv_cache/position_embeddings_adjustment_codegen.py
```diff
@@ -0,0 +1,450 @@
...
+    3. Update the rotary positional embeddings
+        (`self._update_rotary_pos_embedding(...)`)
+
+    For description of every step, please refer to the docstrings of the
+    corresponding methods.
+    """
+
+    # TODO: To refactor so different PositionEmbeddingsAdjustment[LLM name]
+    # classes can share functionalities more efficiently
+    POSITIONS_NAME = "positions"
```
bfineran, line 57: at very least there should be a shared helper function to inject the input name

src/sparseml/exporters/transforms/kv_cache/position_embeddings_adjustment_codegen.py
```diff
@@ -0,0 +1,450 @@
...
+
+        pattern_not_found = None
+        if not rotary_position_embedding_nodes:
+            pattern_not_found = _POSITION_EMBEDDING_PATTERN
+        if not position_embeddings_nodes:
+            pattern_not_found = _ROTARY_POSITION_EMBEDDING_PATTERN
+        if pattern_not_found:
+            raise RuntimeError(
+                f"Unable to find nodes with that match the expected pattern {pattern_not_found}"
+            )
```
bfineran, line 106: why do we raise here instead of continuing?

src/sparseml/exporters/transforms/kv_cache/position_embeddings_adjustment_codegen.py
```diff
@@ -0,0 +1,450 @@
...
+from sparseml.onnx.utils.helpers import get_node_output_nodes, get_nodes_by_output_id
+
+
+__all__ = ["PositionEmbeddingsAdjustmentCodeGen"]
+
+
+# regex to find the node that computes position embeddings
+_POSITION_EMBEDDING_PATTERN = r"/transformer/h.\d+/attn/Range_1"
+# regex to find the node that computes rotary position embeddings
+_ROTARY_POSITION_EMBEDDING_PATTERN = r"/transformer/h.\d+/attn/Slice_(10|9|5|4)"
```
bfineran, line 34: we should avoid using regex for matching and instead rely on node type patterns where possible - especially important as exported names are subject to change both between torch/onnx versions as well as slight changes to model version/size

src/sparseml/exporters/transforms/kv_cache/position_embeddings_adjustment_codegen.py
```diff
@@ -0,0 +1,450 @@
...
+        Here, the `seq_len` is an integer that describes the total number of tokens in the input sequence.
+        This means that e.g.
+        length_cache = layer_past[0].shape[-2] # -> 5
+        length_current = key.shape[1] # -> 1
+        seq_len = length_cache + length_current # -> 6
+
+        hence positional embedding -> torch.arange(seq_len, dtype=torch.float) = np.array([0., 1., 2., 3., 4., 5.])
+
+        Clearly, this is a dynamic value (it depends on the kv cache sequence). It is not yet clear to me whether we
+        could make it static or whether it should remain dynamic and be derived from the static 'positions' input.
```
bfineran, line 160: let's talk more about this

src/sparseml/exporters/transforms/kv_cache/position_embeddings_adjustment_codegen.py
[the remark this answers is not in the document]
bfineran: let's take a look at these and revisit

src/sparseml/exporters/transforms/kv_cache/position_embeddings_adjustment_codegen.py
```diff
@@ -0,0 +1,450 @@
...
+            orphaned_nodes = self._find_orphaned_nodes(model, graph, node)
+            nodes_to_delete.extend(orphaned_nodes)
+
+        graph.delete_nodes(nodes_to_delete)
+        graph.update()
+        graph.delete_unused_initializers()
+
+        return model
+
+    def _find_orphaned_nodes(
```
bfineran, line 424: can be shared, no?

bfineran (changes requested):

src/sparseml/exporters/kv_cache_injector.py
dbogunowicz: Solved in a different PR

src/sparseml/exporters/transforms/kv_cache/position_embeddings_adjustment_codegen.py
dbogunowicz: Your previous function traverses the graph top-bottom, mine goes "upstream". We can potentially discuss creating a single helper function.

src/sparseml/onnx/utils/graph_editor.py
```diff
@@ -450,3 +451,58 @@ def prune_model_one_shot_iter(
         pruned_weight_val = prune_unstructured(weight.val, sparsity)
         update_model_param(model, weight.name, pruned_weight_val)
         yield (index + 1) / len(nodes)
+
+
+def find_orphaned_nodes(model: ModelProto, node: NodeProto) -> List[NodeProto]:
```
bfineran, line 456: would be nice to make this a method of `ONNXGraph`

dbogunowicz: Sounds good

src/sparseml/exporters/kv_cache_injector.py
```diff
@@ -80,63 +89,53 @@ def __init__(
...
-        config_file = config_file[0]
-
-        _LOGGER.info(f"Found config file {config_file}")
-
-        with open(config_file) as f:
-            config = json.load(f)
-
-        return config
+        if positions_adjustment is not None:
+            transforms += [positions_adjustment()]
```
KSGulin, line 137: Nice

src/sparseml/exporters/kv_cache_injector.py
```diff
@@ -80,63 +89,53 @@ def __init__(
...
-                    f"Proceeding with transformation, but may require additional "
-                    f"customization..."
-                )
-
-            num_attention_heads = self.config["num_attention_heads"]
-            hidden_size_kv_cache = self.config["hidden_size"] // num_attention_heads
+
+        if model_path is not None:
+            # get the parameters from the config
+            self.config = get_kv_cache_config(model_path)
```
KSGulin, line 99: If I understand correctly, the `kv_cache_config` is a hardcoded config for each model type. the model_path here is just used to determined the model type right?

src/sparseml/onnx/utils/graph_editor.py
```diff
@@ -242,6 +242,58 @@ def delete_unused_initializers(self):
...
         )  # delete inits that have no edge
 
+    def find_orphaned_nodes(self, node: NodeProto) -> List[NodeProto]:
+        """
+        Given a node, that is to be removed from the graph, find all nodes that
+        will be orphaned as a result of the removal. Orphaned nodes are nodes
+        that will have no inputs after the removal of the given node.
+        The method traverses the graph upwards from the given node until
+        a node with multiple outputs is found. All nodes that are traversed
+        are considered orphaned and will be removed.
```
KSGulin, line 252: It seems there is a potential for this to fail an edge case where the first node with multi-outputs has all its outputs route to different part of the same orphaned subgraph i.e. the multi-output node will be orphaned as well. Is it foreseeable that this would come up in LLM graphs?

KSGulin (approved):
Looks great. Left a couple questions, but they're non-blocking

src/sparseml/exporters/kv_cache_injector.py
dbogunowicz: Yes, exactly. `model_path` is a path to the directory that contains HF's `config.json`. Inside this config, you will find information about the model type and the architecture (num of attention heads etc). All this information is being used to pick the correct "hardcoded config" (semi-hardcoded, since if you have the same model of different size, the "hardcoded config" will be robust against it) and then pass it on in the injection process.

src/sparseml/onnx/utils/graph_editor.py
dbogunowicz: Don't follow, could you maybe sketch the graph for that case?

src/sparseml/exporters/kv_cache_injector.py
```diff
@@ -80,63 +89,53 @@ def __init__(
...
-                    f"Model type {self.config.model_type} is currently not supported. "
-                    f"Supported model types: {_SUPPORTED_ARCHITECTURES}."
-                    f"Proceeding with transformation, but may require additional "
-                    f"customization..."
-                )
-
-            num_attention_heads = self.config["num_attention_heads"]
-            hidden_size_kv_cache = self.config["hidden_size"] // num_attention_heads
+
+        if model_path is not None:
```
bfineran, line 97: logic is getting a bit heavy for init function - suggest breaking out into helper function if we can do so cleanly, i.e.
```
_get_transforms_from_config()
_get_transforms_from_kwargs()
```

src/sparseml/exporters/transforms/kv_cache/cache_keys_and_values.py
```diff
@@ -29,6 +29,7 @@
 
 _LOGGER = logging.getLogger(__name__)
 
+ALLOWED_NODES_BEFORE_SOFTMAX = ["Cast"]
```
bfineran, line 32: extensible, nice

src/sparseml/exporters/kv_cache_injector.py
bfineran, line 99: what happens if config is `None` looks like we crash (after a warning) instead of defaulting?

src/sparseml/exporters/transforms/kv_cache/positions_adjustment_codegen.py
```diff
@@ -0,0 +1,89 @@
...
+    PositionsAdjustmentBase,
+)
+from sparseml.exporters.transforms.utils.matching import get_structural_matches
+from sparseml.onnx.utils import ONNXGraph
+
+
+__all__ = ["PositionsAdjustmentCodeGen"]
+
+
+class PositionsAdjustmentCodeGen(PositionsAdjustmentBase):
```
bfineran, line 27: ..... this is it? if a version upgrade simplifies what we need to do this significantly (both in terms of engineering hours as well as de-risking model/inference errors) definitely push for prioritize upgrade!

great work here

src/sparseml/onnx/utils/graph_editor.py
bfineran: in general, nodes don't have multiple outputs unless potentially they are a graph output - but agree we should look into this edge case. @KSGulin can you sync w/ @dbogunowicz or propose a solution (this should be a simple fix if it is indeed an issue, no?)

bfineran (approved):
looks great pending comments

src/sparseml/exporters/transforms/kv_cache/positions_adjustment_codegen.py
dbogunowicz: I was totally overestimating the amount of work that is needed to do the transformer upgrade. Shame on me 😊

src/sparseml/exporters/transforms/kv_cache/cache_keys_and_values.py
dbogunowicz: Yes, spoiler alert, for BLOOM you'd have to add a Reshape into this set ;)

src/sparseml/exporters/kv_cache_injector.py
dbogunowicz: Good idea.

src/sparseml/exporters/kv_cache_injector.py
dbogunowicz: You mean when there are no `kwargs` and `config = None` right? Yes, this is a corner case that I should cover, good point.

src/sparseml/onnx/utils/graph_editor.py
dbogunowicz: @KSGulin Feel free to sync with me spontaneously today or tomorrow, or otherwise, we can land this and create a small patch for this next week.

src/sparseml/exporters/kv_cache_injector.py
dbogunowicz: Second thought, this is already covered. If config is `None` we will end up raising an error. I think this makes sense, defaulting here is not a good idea IMO

KSGulin (approved):

bfineran (approved):

PR #1559, opened 2023-05-11
[WIP] 'Generic' KV cache injection support

bfineran:

PR #1616, opened 2023-06-08
ONNXToDeepSparse matmul integer conversion patches for whipser support

bfineran:
Quantization experiments for the whipser model resulted with a few `MatMulInteger` ops that did not have their weights properly folded. This is primarily due to the non-parameterized matmul matching improperly not skipping over parameterized matmuls (see fixes in [src/sparseml/exporters/transforms/matmul_to_matmulinteger_cast_mul.py](https://github.com/neuralmagic/sparseml/compare/whisper-onnx-to-deepsparse-patches?expand=1#diff-91a487ce10f7dec24c6c93183bfd8398e02168c8f835ccdddbffe06802285b9f)).

The longer fixes in the other two files update pattern matching to convert parameterized matmuls without bias adds, this were lifted from my updates in #1559 that are still pending review. We can merge these first as they are tangential to that PR as well

**test_plan:**
Improper conversion before:
Screenshot 2023-06-08 at 5 20 05 PM

Fixed conversion after:
Screenshot 2023-06-08 at 5 20 33 PM


dbogunowicz (dismissed):

KSGulin (approved):

KSGulin (commented):
Looks like the exporter tests are failing. Otherwise LGTM

KSGulin (approved):

PR #1618, opened 2023-06-09
Update sparseml to support torch 2.0

dsikka:
- Updated max version of `torch` to be 2.0, updated `torchvision` max to be `0.15.1`
- Updated max `torchaudio` to be `2.0.1`
- Tested and passed all `pytorch` tests 
- Tested updated package versions in training using `sparseml.image_classification.train`. Used the `ImageNette` dataset and `ResNet-50` model from SparseZoo

```
# Epoch and Learning-Rate variables
num_epochs: 3.0
init_lr: 0.0005

# quantization variables
quantization_epochs: 1.0

training_modifiers:
  - !EpochRangeModifier
    start_epoch: 0.0
    end_epoch: eval(num_epochs)

    final_lr: 0.0
    init_lr: eval(init_lr)
    lr_func: cosine
    start_epoch: 0.0
    end_epoch: eval(num_epochs)

# Phase 1 Sparse Transfer Learning / Recovery
sparse_transfer_learning_modifiers:
  - !ConstantPruningModifier
    start_epoch: 0.0
    params: __ALL_PRUNABLE__

# Phase 2 Apply quantization
sparse_quantized_transfer_learning_modifiers:
  - !QuantizationModifier
    start_epoch: eval(num_epochs - quantization_epochs)
```

- Trained using the sample recipe above and then exported in onnx using `sparseml.image_classification.export_onnx`
- Confirmed size of the onnx export model to match another model trained/exported using torch 1.13.1 (23 mb) and validated graph using Netron thanks to @KSGulin

### DIFF
--- a/src/sparseml/pytorch/base.py
+++ b/src/sparseml/pytorch/base.py
@@ -49,7 +49,7 @@ __all__ = [
 
 
 _TORCH_MIN_VERSION = "1.0.0"
-_TORCH_MAX_VERSION = "1.13.100"  # set bug to 100 to support all future 1.9.X versions
+_TORCH_MAX_VERSION = "2.0.0"  # set bug to 100 to support all future 1.9.X versions
 
 
 def check_torch_install(
--- a/src/sparseml/pytorch/base.py
+++ b/src/sparseml/pytorch/base.py
@@ -49,7 +49,7 @@ __all__ = [
 
 
 _TORCH_MIN_VERSION = "1.0.0"
-_TORCH_MAX_VERSION = "2.0.0"
+_TORCH_MAX_VERSION = "2.0.100"
 
 
 def check_torch_install(
--- a/src/sparseml/pytorch/base.py
+++ b/src/sparseml/pytorch/base.py
@@ -49,7 +49,7 @@ __all__ = [
 
 
 _TORCH_MIN_VERSION = "1.0.0"
-_TORCH_MAX_VERSION = "2.0.0"  # set bug to 100 to support all future 1.9.X versions
+_TORCH_MAX_VERSION = "2.0.0"
 
 
 def check_torch_install(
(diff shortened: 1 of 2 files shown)


bfineran (commented):
thanks @dsikka - one last thing to look for is exporting quantized models with `convert_qat=True` and verifying that the weights are properly compressed (ie load a quantized resnet pytorch model from the sparsezoo, export it to onnx and verify that the file size is the expected quantized one)

dbogunowicz:
Oh, wow, this seems to be much less effort than expected no?

bfineran (dismissed):

src/sparseml/pytorch/base.py
```diff
@@ -49,7 +49,7 @@
 
 
 _TORCH_MIN_VERSION = "1.0.0"
-_TORCH_MAX_VERSION = "1.13.100"  # set bug to 100 to support all future 1.9.X versions
+_TORCH_MAX_VERSION = "2.0.0"  # set bug to 100 to support all future 1.9.X versions
```
rahul-tuli, line 52: Could we remove this outdated comment?

rahul-tuli (commented):
LGTM after outdated comment removal!

src/sparseml/pytorch/base.py
```diff
@@ -49,7 +49,7 @@
 
 
 _TORCH_MIN_VERSION = "1.0.0"
-_TORCH_MAX_VERSION = "1.13.100"  # set bug to 100 to support all future 1.9.X versions
+_TORCH_MAX_VERSION = "2.0.0"
```
bfineran, line 52: Actually building on Rahul's comment, we should match the patch version to .100 still to support future patches of 2.0.X

bfineran (changes requested):

src/sparseml/pytorch/base.py
dsikka: I.e set `2.0.100` so that we can support all `2.0.x` versions in the future? @bfineran

bfineran (approved):

rahul-tuli (approved):

PR #1625, opened 2023-06-14
[BugFix] Hide `sparseml.benchmark`

rahul-tuli:
This PR is a first step towards deprecating `sparseml.benchmark` utility which is fairly old and will be deprecated/refactored or removed in a future version

For now, this removes the public pathway `sparseml.benchmark`
Additionally raises a deprecation warning if the script is invoked directly

The actual utility wasn't removed as of yet, cause the helpers in this namespace might be required by `sparsify`

### DIFF
--- a/src/sparseml/benchmark/info.py
+++ b/src/sparseml/benchmark/info.py
@@ -91,7 +91,6 @@ sparseml.benchmark --framework onnx --model ~/downloads/model.onnx \
 import argparse
 import logging
 import os
-import warnings
 from abc import ABC, abstractmethod
 from typing import Any, Dict, Iterable, Iterator, Optional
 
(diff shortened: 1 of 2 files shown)


src/sparseml/benchmark/info.py
```diff
@@ -544,6 +545,11 @@ def _parse_args():
 
 
 def _main():
+    warnings.warn(
```
bfineran, line 548: use to handle setting the warning (by default deprecation warnings are suppressed): https://github.com/neuralmagic/sparseml/blob/c3612a19b16a3d60493ff5447c5c860c16e172df/src/sparseml/utils/helpers.py#L827

bfineran (approved):

KSGulin (approved):

PR #1626, opened 2023-06-14
[CLIP]: Sparseml onnx export pathways for open_clip models

dsikka:
### Summary:

- The current onnx export functionality using OpenClip requires torch nightly and opset = 14
- This PR provides an example of how to use sparseml's onnx export functionality to export the visual and text branches of the CLIP model, using [OpenClip](https://github.com/mlfoundations/open_clip)
- OpenClip wraps everything in a `torch.nn.Module` object for the visual branch (ie everything required for the forward pass) but does not do this for the text branch. For the sake of completeness, the `TextModel` class was added (see `clip_models.py`) which wraps the appropriate steps needed in the forward pass. This is then exported 
- Furthermore, for captioning, OpenClip provides thee CoCa models, which requires an additional text-decoder and also uses a separate interface compared to the other models. As such, there is a special `is_coca` case provided in the scripts

### Test Plan:
- Ran this locally (requires openclip and torch nightly to be installed) and created a `convnext_onnx` directory with the `clip_text.onnx` and `clip_visual.onnx` exports
```
 python clip_onnx_export.py --model convnext_base_w_320 --pretrained laion_aesthetic_s13b_b82k --export-path convnext_onnx
```

- Also ran a CoCa-specific command which created the default `clip_onnx` directory with the visual and text exports, along with the text decoder export `clip_text_decoder.onnx`
```
python clip_onnx_export.py --model coca_ViT-L-14 --pretrained laion2b_s13b_b90k
```

### DIFF
--- a/integrations/clip/clip_models.py
+++ b/integrations/clip/clip_models.py
@@ -1,14 +1,29 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 import torch
 import torch.nn as nn
 
 
 class VisualModel(nn.Module):
-    def __init__(self, visual_model, transformations):
+    def __init__(self, visual_model, transformations, output_tokens):
 
         super().__init__()
 
         self.visual_model = visual_model
         self.transformations = transformations
+        self.visual_model.output_tokens = output_tokens
 
     def forward(self, x):
         return self.visual_model(x)
--- a/integrations/clip/clip_models.py
+++ b/integrations/clip/clip_models.py
@@ -17,12 +17,11 @@ import torch.nn as nn
 
 
 class VisualModel(nn.Module):
-    def __init__(self, visual_model, transformations, output_tokens):
+    def __init__(self, visual_model, output_tokens):
 
         super().__init__()
 
         self.visual_model = visual_model
-        self.transformations = transformations
         self.visual_model.output_tokens = output_tokens
 
     def forward(self, x):
@@ -33,7 +32,6 @@ class TextModel(nn.Module):
     def __init__(
         self,
         token_embedding,
-        tokenizer,
         positional_embedding,
         transformer,
         ln_final,
@@ -44,7 +42,6 @@ class TextModel(nn.Module):
         super().__init__()
 
         self.token_embedding = token_embedding
-        self.tokenizer = tokenizer
         self.positional_embedding = positional_embedding
         self.transformer = transformer
         self.ln_final = ln_final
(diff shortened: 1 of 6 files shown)


integrations/clip/clip_models.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def forward(self, input_ids):
+        x = self.token_embedding(input_ids).to(self.cast_dtype)
+        x = x + self.positional_embedding.to(self.cast_dtype)
+        x = x.permute(1, 0, 2)  # NLD -> LND
+        x = self.transformer(x, attn_mask=self.attn_mask)
+        x = x.permute(1, 0, 2)  # LND -> NLD
+        x = self.ln_final(x)  # [batch_size, n_ctx, transformer.width]
+        # take features from the eot embedding (eot_token = highest in each sequence)
+        x = x[torch.arange(x.shape[0]), input_ids.argmax(dim=-1)] @ self.text_projection
```
bfineran, line 63: what semantically is the output here? are we only returning scores for the highest activations?

integrations/clip/clip_models.py
```diff
@@ -0,0 +1,64 @@
...
+        transformer,
+        ln_final,
+        text_projection,
+        attn_mask,
+    ):
+
+        super().__init__()
+
+        self.token_embedding = token_embedding
+        self.tokenizer = tokenizer
```
bfineran, line 47: where is this tokenizer used?

dsikka: Just for the sample batch. Added to the `TextModel` object for the sake of completeness but not needed in the forward pass. Can remove

integrations/clip/clip_models.py
dsikka: In the last time, they seems to only take features from the dimension with the end-of-text token, reducing the dimension from `batch x context_length x transformer.width` to `batch x transformer.width` which is then projected

integrations/clip/clip_models.py
bfineran: makes sense, let's slim down any args not needed for onnx export / forward pass

bfineran (approved):
LGTM pending comments

bfineran (dismissed):

integrations/clip/clip_onnx_export.py
```diff
@@ -0,0 +1,229 @@
...
+
+def _export_text(
+    model: torch.nn.Module,
+    device: str,
+    export_path: Union[str, Path],
+    tokenizer,
+    is_coca: bool,
+    **export_kwargs,
+):
+    module_name = "clip_text.onnx"
```
rahul-tuli, line 86: Would this fail or overwrite the model if existing `clip_text.onnx` is present? Same comment for other hardcoded names

integrations/clip/clip_onnx_export.py
```diff
@@ -0,0 +1,229 @@
...
+
+"""
+Examples on how to use sparsemls's onnx export functionality to export CLIP visual
+and text models using the OpenCLIP API.
+
+Note: This requires torch nightly and openclip to be installed:
+https://github.com/mlfoundations/open_clip
+
+"""
+import argparse
```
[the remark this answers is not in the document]
dsikka: Is this necessary for an example? I'd rather not introduce it at this point in the dev cycle. Potentially in a follow-up PR?

KSGulin (approved):

integrations/clip/clip_onnx_export.py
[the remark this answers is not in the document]
rahul-tuli: Sure, works for me as long as we do scope it in.!

integrations/clip/clip_onnx_export.py
dsikka: It won't fail. Will just write over the existing files.

rahul-tuli (approved):
LGTM!

KSGulin (approved):

PR #1629, opened 2023-06-15
Set lower validation batch size for yolov8

abhinavnmagic:
Ultralytics by default sets the validation batch size as twice the training batch size. This leads to OOM with our integration for YoloV8 m,l and x models.

The fix is to set the validation batch size to be one-fourth of the training batch size.
I tested the changes while training yolov8 models.


anmarques (approved):

src/sparseml/yolov8/trainers.py
```diff
@@ -279,6 +279,10 @@ def _setup_train(self, rank, world_size):
         super()._setup_train(rank, world_size)
         # NOTE: self.resume_training() was called in ^
 
+        if rank in {0, -1}:
+            self.test_loader = self.get_dataloader(self.testset, batch_size=max(1, self.train_loader.batch_size // 4), rank=-1, mode='val')
```
anmarques, line 283: Won't this lead to batch size 0 if the original batch size is between 1 and 3? Maybe make sure that it's at least 1

anmarques (changes requested):

bfineran (approved):

bfineran (approved):

PR #1630, opened 2023-06-16
[Required for MPT models] Expose trust_remote_code flag for HF-transformers

eldarkurtic:
MPT models are not yet integrated in HF-transformers, and therefore to load them for ONNX export we need to explicitly set `trust_remote_code` flag when loading config and model with `Auto` feature from `transformers`.

Without this patch, running `sparseml.transformers.export_onnx ....` on MPT models will crash with: `ValueError: Loading 68e1a8e0ebb9b30f3c45c1ef6195980f29063ae2 requires you to execute the configuration file in that repo on your local machine. Make sure you have read the code there to avoid malicious use, then set the option `trust_remote_code=True` to remove this error.`

### DIFF
--- a/.MAINTAINERS
+++ b/.MAINTAINERS
@@ -9,10 +9,9 @@ rahul-tuli
 KSGulin
 dbogunowicz
 # anmarques
-kylesayrs
 # eldarkurtic
-corey-nm
 # chibukach
 # shubhra
 # abhinavnmagic
 # eiofinov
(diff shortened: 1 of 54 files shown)


natuan (approved):

eldarkurtic:
Changes merged with this PR: https://github.com/neuralmagic/sparseml/pull/1654

PR #1634, opened 2023-06-16
[1.5.2] restrict pandas to <2.0

bfineran:
this fixes a bug in the latest NM 1.5 supported transformers datasets that is incompatible with pandas 2.0. Specifically, a removed kwarg is passed to pandas on `datasets.load_dataset`.

Future releases will support later datasets versions.


rahul-tuli (approved):
LGTM!

PR #1642, opened 2023-06-28
Update setup.py

mgoin:


bfineran (commented):
@mgoin can you give more context on why here?

PR #1643, opened 2023-06-29
YOLOv8 ONNX export fix

anmarques:
Before this PR the ONNX for YOLOv8 would include 3 intermediate outputs. This creates performance issues for the engine. This PR fixes this behavior such that the ONNX will only have the final detection output.

The fix is based on this line of code from the original export script from ultralytics: https://github.com/ultralytics/ultralytics/blob/64f247d6923e28108792cbeea16343606d8cd844/ultralytics/yolo/engine/exporter.py#L180

In short, the Segment and Detection heads have an internal parameter "export" that is set to False by default. When this parameter is False, the heads will output the intermediate values, when it's set to True they output only the final results. The fix simply switches this parameter to True in the export flow.

### DIFF
--- a/src/sparseml/yolov8/trainers.py
+++ b/src/sparseml/yolov8/trainers.py
@@ -48,6 +48,7 @@ from sparsezoo import Model
 from sparsezoo.utils import validate_onnx
 from ultralytics import __version__
 from ultralytics.nn.tasks import SegmentationModel, attempt_load_one_weight
+from ultralytics.nn.modules import Detect, Segment
 from ultralytics.yolo.cfg import get_cfg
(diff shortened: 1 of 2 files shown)


mgoin (dismissed):
i understand removing the outputs (thank you) but unsure if this is intended default behavior from ultralytics and should be upstreamed, or just should stay as a local change in our code

anmarques (commented):
The intermediate outputs are not needed in eval model. Ultralytics has a ONNX export path that does the same changes I implemented in the PR. We implemented our own ONNX export path, and that's where we were missing the parameter change.

mgoin (approved):

andy-neuma (approved):
cool

bfineran (approved):

PR #1649, opened 2023-07-03
Add variable bit width support to ONNXToDeepsparse

KSGulin:
This PR adds support for variable-bit weight quantization in the ONNXToDeepsparse exporter. This affects two steps:

- Conversion of intiailziers to unit8
- Clipping in quantization of weight arrays

**Test Plan**
Local sparsify runs with int-4 weight quantization


bfineran (changes requested):
also need to propagate bit width to clip weights in the target bit range on the actual quantization

KSGulin:
@bfineran good callout. Updated the array quantization routine and propagated the `bit_width` args to address this

jeanniefinks:

PR #1654, opened 2023-07-06
Fix export of all quantized transformer models

eldarkurtic:
Our current transformers export-to-onnx pipeline doesn't work for quantized models. To reproduce: try to export any quantized model with the latest versions of sparseml and neuralmagic/transformers.

There are two reasons:
1. we are passing in `resolved_archive_file=[]` at https://github.com/neuralmagic/sparseml/blob/b73a173ff89c3bb524dcc0a1f7a16a3109a234a1/src/sparseml/transformers/sparsification/trainer.py#L690 which causes error in HF library here: https://github.com/neuralmagic/transformers/blob/0798c9e3b743a7e5c552f943a1a7d52ff63bbffb/src/transformers/modeling_utils.py#L3150 . We can fix this by passing `None` instead of `[]`. 
2. the function `self.model._load_pretrained_model` from transformers returns 6 items (see https://github.com/neuralmagic/transformers/blob/0798c9e3b743a7e5c552f943a1a7d52ff63bbffb/src/transformers/modeling_utils.py#L3323), and in our interface we only accept 5 https://github.com/neuralmagic/sparseml/blob/b73a173ff89c3bb524dcc0a1f7a16a3109a234a1/src/sparseml/transformers/sparsification/trainer.py#L686 which also raises an issue while unpacking the returned tuple from their function.


bfineran (dismissed):
great catch @eldarkurtic this should also fix our broken tests on main. Thank you!

eldarkurtic:
Accidentally pushed changes from this PR as well to enable `--trust_remote_code`, so we can close that PR as well by merging this one into main.

bfineran (approved):

bfineran:
failing tests are unrelated - investigating separately but unable to reproduce outside of GHA. Merging this PR to unblock functionality

PR #1657, opened 2023-07-07
[Sparsify.alpha][Image Classification] use registry dataset is possible when local path to dataset is not available

bfineran:
enables example registry datasets to be downloaded automatically if a local imagefolder of the same name is not found

**testing:**
verified the following previously failing sparsify command now automatically downloads the example dataset and runs
```
sparsify.run sparse-transfer --use-case image-classification --data imagenette
```


KSGulin (approved):

bfineran:
test failures unrelated - merging to unblock testing

PR #1660, opened 2023-07-10
Fix broken `protobuf` ranges

dsikka:
- Fix protobuf ranges to account for updates in other packages and get our pytorch tests working


rahul-tuli (approved):

bfineran (approved):
note that @mgoin recently uncapped the range to enable onnx 1.14. It looks like 3.20.2 is in range so we should be good to both fix our issue for now while keeping onnx 1.14 enabled

rahul-tuli (approved):

bfineran (approved):

PR #1663, opened 2023-07-11
Fix Falcon & MPT MatMul folding during ONNX export

natuan:
Need this new pattern matching for Falcon/MPT during ONNX export to correctly fold MatMuls into MatMulInteger.

Also did some renaming for related funcs, hopefully making their intention clearly.

Qualification: manually tested with Falcon-7b pruned quantized.


natuan:
Close the PR because it's been added as part of ONNXToDeepSparse

PR #1667, opened 2023-07-12
[question answering] Fallback to loading JSONL files without `data` prefix

bfineran:
fixes a bug with custom datasets for `sparseml.transformers.question_answering` where custom JSONL files without a top level `"data"` field wrapping each line

This catches specifically a JSONDecodeError and tries the fallback if so.

**test_plan:**
Reproduced a failing sparsify command and verifies this fixes it:
```
sparsify.run sparse-transfer --use-case question-answering --data local_squad --optim-level 50
```

### DIFF
--- a/src/sparseml/transformers/question_answering.py
+++ b/src/sparseml/transformers/question_answering.py
@@ -20,10 +20,7 @@
 """
 Fine-tuning the library models for question answering integrated with sparseml
 """
-
-# You can also adapt this script on your own question answering task.
-# Pointers for this are left as comments.
-
+import json
 import logging
 import os
 import sys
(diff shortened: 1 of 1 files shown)


src/sparseml/transformers/question_answering.py
```diff
@@ -913,12 +917,22 @@ def _get_raw_dataset(data_args, cache_dir: Optional[str] = None):
...
+
+        try:
+            raw_datasets = load_dataset(
+                extension,
+                data_files=data_files,
+                field="data",
+                cache_dir=cache_dir,
+            )
+        except json.JSONDecodeError:
+            # run without `field="data"` - JSONL files will not be nested uner
```
bfineran, line 929:
```suggestion
            # run without `field="data"` - JSONL files will not always have each line nested under
```

PR #1670, opened 2023-07-13
Support Int4 ONNX Export

Satrat:
Int4 QAT was working out of the box, but the onnx export step was failing because [onnx only supports 8-bit integer quantization](https://onnxruntime.ai/docs/performance/model-optimizations/quantization.html). The fix for this was to trick onnx into thinking we were exporting 8-bit quantized weights by manually adjusting the range of the `FakeQuantize` modules. The weights themselves will still be between [-8, 7], and the original range boundaries are still accessible via the `FakeQuantize.quant_min` and `FakeQuantize.quant_max` attributes (they are deprecated by PyTorch, replaced by `FakeQuantize.activation_post_process.quant_[min\max]`)

### Testing
Run `tests/sparseml/pytorch/test_torch_to_onnx_exporter/test_export_4bit_model` for example usage, or just run `TorchToONNX.export()` on any Int4 QAT model:
```
model = ...
sample_batch = ...
exporter = TorchToONNX(sample_batch)
exporter.export(model, "model.onnx")
```

### Old Behavior
Export fails with error:
```
"name": "SymbolicValueError",
"message": "For (quant_min, quant_max), ONNX allows only (0, 127), (0, 255) and (-128, 127). Got (-8, 7)
```

### New Behavior
Export completes without errors

### DIFF
--- a/src/sparseml/pytorch/torch_to_onnx_exporter.py
+++ b/src/sparseml/pytorch/torch_to_onnx_exporter.py
@@ -28,7 +28,11 @@ from sparseml.exporters.base_exporter import BaseExporter
 from sparseml.exporters.transforms.base_transform import BaseTransform
 from sparseml.pytorch import _PARSED_TORCH_VERSION
 from sparseml.pytorch.opset import TORCH_DEFAULT_ONNX_OPSET
-from sparseml.pytorch.utils.helpers import tensors_module_forward, tensors_to_device
+from sparseml.pytorch.utils.helpers import (
+    adjust_quantization_for_onnx_export,
+    tensors_module_forward,
+    tensors_to_device,
+)
 from sparseml.pytorch.utils.model import is_parallel_model
 from sparsezoo.utils import save_onnx
 
@@ -190,21 +194,7 @@ class _TorchOnnxExport(BaseTransform):
             )
         }
 
-        # adjust 4-bit ranges to 8-bits, since ONNX does not have 4-bit support
-        fake_quant_modules = [
-            m for m in module.modules() if m.__class__.__name__ == "FakeQuantize"
-        ]
-        for quant in fake_quant_modules:
-            # original ranges preserved in quant.quant_min and quant.quant_max
-            quant_range = (
-                quant.activation_post_process.quant_min,
-                quant.activation_post_process.quant_max,
-            )
-            if quant_range == (-8, 7):  # convert int4 range to int8
-                quant.activation_post_process.quant_min = -128
-                quant.activation_post_process.quant_max = 127
-            elif quant_range == (0, 15):  # convert uint4 to uint8
-                quant.activation_post_process.quant_max = 255
+        adjust_quantization_for_onnx_export(module)  # in-place operation
 
         # disable active quantization observers because they cannot be exported
         disabled_observers = []
--- a/src/sparseml/pytorch/torch_to_onnx_exporter.py
+++ b/src/sparseml/pytorch/torch_to_onnx_exporter.py
@@ -190,6 +190,22 @@ class _TorchOnnxExport(BaseTransform):
             )
         }
 
+        # adjust 4-bit ranges to 8-bits, since ONNX does not have 4-bit support
+        fake_quant_modules = [
+            m for m in module.modules() if m.__class__.__name__ == "FakeQuantize"
+        ]
+        for quant in fake_quant_modules:
+            # original ranges preserved in quant.quant_min and quant.quant_max
+            quant_range = (
+                quant.activation_post_process.quant_min,
+                quant.activation_post_process.quant_max,
+            )
+            if quant_range == (-8, 7):  # convert int4 range to int8
+                quant.activation_post_process.quant_min = -128
+                quant.activation_post_process.quant_max = 127
+            elif quant_range == (0, 15):  # convert uint4 to uint8
+                quant.activation_post_process.quant_max = 255
+
         # disable active quantization observers because they cannot be exported
         disabled_observers = []
         for submodule in module.modules():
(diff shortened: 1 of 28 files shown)


src/sparseml/pytorch/torch_to_onnx_exporter.py
```diff
@@ -190,6 +190,22 @@ def transform(
             )
         }
 
+        # adjust 4-bit ranges to 8-bits, since ONNX does not have 4-bit support
```
bfineran, line 193: let's move this to a helper function given it's repeated in `exporter.py`

src/sparseml/pytorch/utils/exporter.py
```diff
@@ -208,17 +208,34 @@ def export_onnx(
         """
         if not export_kwargs:
             export_kwargs = {}
+
+        module = deepcopy(self._module).cpu()  # don't modify the original model
```
bfineran, line 212: I see we deepcopy in `TorchToONNX`, wondering if this will be too much overhead as models get larger, probably not, just something to keep in mind

src/sparseml/pytorch/torch_to_onnx_exporter.py
```diff
@@ -190,6 +190,22 @@ def transform(
...
+            # original ranges preserved in quant.quant_min and quant.quant_max
+            quant_range = (
+                quant.activation_post_process.quant_min,
+                quant.activation_post_process.quant_max,
+            )
+            if quant_range == (-8, 7):  # convert int4 range to int8
+                quant.activation_post_process.quant_min = -128
+                quant.activation_post_process.quant_max = 127
+            elif quant_range == (0, 15):  # convert uint4 to uint8
+                quant.activation_post_process.quant_max = 255
```
bfineran, line 207: to support arbitrary bit widths, lets support any quant range that isn't int8 or uint8 (choosing int vs uint depending on if `quant_min < 0`)

tests/sparseml/pytorch/test_torch_to_onnx_exporter.py
```diff
@@ -24,13 +24,81 @@
...
+    "model,sample_batch",
+    [
+        (MLPNet(), torch.randn(8)),
+        (MLPNet(), torch.randn(10, 8)),
+        (LinearNet(), torch.randn(8)),
+        (LinearNet(), torch.randn(10, 8)),
+        (ConvNet(), torch.randn(1, 3, 28, 28)),
+    ],
+)
+def test_export_4bit_model(tmp_path, model, sample_batch):
```
bfineran, line 74: nice test, to take it a step further, would be good to try running `ONNXToDeepSparseExporter` as well to convert the "Q/DQ" nodes+weights and then verify that the resulting weights are in the correct bit range

bfineran (changes requested):

src/sparseml/pytorch/utils/exporter.py
Satrat: Yeah this could definitely add computation time for larger models, but I think its better to avoid modifying the original model thats passed in to the function. If a user is going to keep using the model after export we don't want to be unexpectedly changing it

bfineran: yes agreed - if we ever go back, we can always track which modules are updated and what their original ranges were

tests/sparseml/pytorch/test_torch_to_onnx_exporter.py
```diff
@@ -24,13 +24,81 @@
...
 from tests.sparseml.pytorch.helpers import ConvNet, LinearNet, MLPNet
 
 
+QUANT_RECIPE = """
+!QuantizationModifier
+    start_epoch: 0.0
+    scheme:
+        input_activations:
+            num_bits: 4
+            symmetric: False
```
mgoin, line 41: it's cool how simple it is to represent this in a recipe. could you also test where the activations are not quantized to the same num_bits as the weights? AFAIK we're targeting 8-bit activations at lowest precision

rahul-tuli (commented):
Nice diff! LGTM pending comments

src/sparseml/pytorch/utils/helpers.py
```diff
@@ -1174,3 +1177,41 @@ def detach(x: Union[torch.Tensor, List, Tuple]):
         return tuple([detach(e) for e in x])
     else:
         raise ValueError("Unexpected type to detach")
+
+
+def adjust_quantization_for_onnx_export(module: torch.nn.Module) -> torch.nn.Module:
+    # supported pytorch ranges are int8 or uint8
+    allowed_ranges = [(0, 127), (0, 255), (-128, 127)]
```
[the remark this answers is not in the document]
Satrat: These are the ranges officially supported by ONNX, the documentation briefly mentions that the (0,127) range is a workaround to support certain machines: https://onnxruntime.ai/docs/performance/model-optimizations/quantization.html#when-to-use-reduce-range-and-per-channel-quantization

bfineran (approved):

src/sparseml/pytorch/utils/helpers.py
```diff
@@ -1174,3 +1177,41 @@ def detach(x: Union[torch.Tensor, List, Tuple]):
...
+            quant_range = (quant.quant_min, quant.quant_max)
+            if quant_range not in allowed_ranges:
+                if quant_range[0] < 0:  # convert signed range to int8
+                    quant.quant_min = -128
+                    quant.quant_max = 127
+                else:  # convert unsigned range to uint8
+                    quant.quant_min = 0
+                    quant.quant_max = 255
+            # don't update observer since ranges are artificially modified
+            quant.observer_enabled[0] = 0
```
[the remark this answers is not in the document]
rahul-tuli: Let's do that in a follow up PR!

rahul-tuli (approved):

PR #1676, opened 2023-07-18
[KV Cache Injection] Causal Mask for CodeGen

dbogunowicz:
This PR:

- implements causal mask injection for the CodeGen model
- begins a process of refactoring the `AdditionalTransforms` objects, so that the `AdditionalTransformsBase` is to share more and more helper methods that can be shared across its children classes and reduce boilerplate code. This process will be continued with the incoming causal mask injections for other LLMs

For testing please refer to: https://github.com/neuralmagic/deepsparse/pull/1127


src/sparseml/pytorch/utils/helpers.py
```diff
@@ -1174,3 +1177,41 @@ def detach(x: Union[torch.Tensor, List, Tuple]):
         return tuple([detach(e) for e in x])
     else:
         raise ValueError("Unexpected type to detach")
+
+
+def adjust_quantization_for_onnx_export(module: torch.nn.Module) -> torch.nn.Module:
```
bfineran, line 1182: this needs rebase

PR #1679, opened 2023-07-20
Channel-Wise Quantization with QDQ Folding

Satrat:
This branch adds a new `strategy` parameter to the QuantizationModifier, which can be set to `channel` to enable independent scale and zero point values for each channel of a quantized weight tensor. This improves the accuracy of `int4` quantization, since each channel can have its own range. 

I updated the `ONNXToDeepsparse` converter to support for tensor scale and zero_points, and made some changes to the dimensions of the Add/Mul operators to work with a channel-wise ConvInteger. This should also work with [Rahul's Channel-wise OBQ PR](https://github.com/neuralmagic/sparsifyml/pull/138), but it still needs to be tested

### Testing
New unit tests were added to `text_torch_to_onnx_exporter.py`, but note that the channel-wise tests are disabled for torch<2.0, as channel-wise quantization is not supported in easier versions

I also modified an existing torchvision tutorial to test the whole end-to-end pipeline of training, export and inference, which can be found here: integrations/torchvision/tutorials/docs-torchvision-channelwise-4bit-transfer-imagenette.ipynb

### Before QDQ Folding
Screenshot 2023-07-25 at 12 21 25 PM

### After QDQ Folding
Screenshot 2023-07-25 at 12 22 03 PM


src/sparseml/pytorch/sparsification/quantization/helpers.py
```diff
@@ -150,6 +154,8 @@ class QConfigProperties:
     weight_dtype: torch.dtype = torch.qint8
     activation_bits: int = 8
     weight_bits: int = 8
+    activation_strategy: str = "tensor"
```
rahul-tuli, line 157: nice

rahul-tuli (approved):
LGTM!

PR #1683, opened 2023-07-21
Revert "Fix onnx export for quantized MPT models 2 (#1647)"

rahul-tuli:
This reverts commit e8ff0f6bd5ed4fcb8d281f30462462b0125b4ddf.
These commit breaks our current YOLOv8 flows (After #1680 ); reverting them for now, a fix will be put in later for MPT models


dsikka (approved):

markurtz (approved):

Satrat (approved):

markurtz:
@eldarkurtic noting here for you, let's talk through potential general solution that we can use here

abhinavnmagic:
I just verified that reverting the earlier commit (https://github.com/neuralmagic/sparseml/commit/e8ff0f6bd5ed4fcb8d281f30462462b0125b4ddf) works.

PR #1685, opened 2023-07-24
Implementation of powerpropagation as a modifier

ohaijen:
Implement Powerpropagation (raising weights to a power in forward pass), see https://arxiv.org/abs/2110.00296

The functionality is implemented as a Modifier that, when activated, wraps pruned modules (for now, only Conv2d or Linear, but can easily be extended later) in a PowerpropagationWrapper, which does the exponentiation. Once the modifier is deactivated, the architecture change is undone.

Testing: wrote new tests, including for the architecture change and corresponding weight updates. Ran a few epochs on imagenet to make sure that training was looking OK (loss going down/accuracy going up).


src/sparseml/pytorch/sparsification/pruning/modifier_powerpropagation.py
```diff
@@ -0,0 +1,327 @@
...
+            return F.linear(x, weight, self.layer.bias)
+        else:
+            raise ValueError(
+                "Powerpropagation only works with Linear and Conv2d layers"
+            )
+
+    def set_alpha(self, new_alpha):
+        with torch.no_grad():
+
+            self.layer.weight *= pow(abs(self.layer.weight), self.alpha / new_alpha - 1)
```
abhinavnmagic, line 90: I didn't quite understand this part. From my understanding, this is the initializing the weight parameters. Doesn't the paper (page 3, second last para) mention this as: `\phi_i=sign(\theta_i) * pow(\abs{\theta_i}, \frac{1}{\alpha})`.

ohaijen: This is a generalization of the formula you quote. So the formula you wrote is for updating the exponent when the current exponent is 1 (this is the `\frac{1}`). So replace the 1 with some other exponent, say `\rho`:
```
\phi_i=sign(\theta_i)* pow(\abs{\theta_i}, \frac{\rho}{\alpha})
= \theta_i*pow(\abs{\theta_i}, \frac{\rho}{\alpha}-1)
```

So this is what I have, except `\theta_i` is `self.layer.weight`, `\alpha` is `new_alpha`, and `\rho` is `self.alpha`.

abhinavnmagic: Yeah, makes sense. Thank you for clarifying.

abhinavnmagic (dismissed):

anmarques (dismissed):

anmarques (approved):

abhinavnmagic (approved):

PR #1691, opened 2023-07-27
[KV Cache Injection] MPT

dbogunowicz:


src/sparseml/exporters/transforms/kv_cache/configs.py
```diff
@@ -102,20 +105,24 @@ class Config:
...
     transpose_value_input=None,
     transpose_key_input=(0, 1, 3, 2),
-    multiply_batch_by_num_att_heads=True,
+    multiply_batch_by_num_att_heads=False,
 )
 
 
 def get_kv_cache_config(
     model_path: str,
-    supported_configs: List[BaseModel] = [OPT_CONFIG, CODEGEN_CONFIG, BLOOM_CONFIG],
```
dbogunowicz, line 118: We need to remove BLOOM, we are not supporting it officially in the nightly.

bfineran: if we have some internal testing, hidden support seems ok

src/sparseml/exporters/transforms/kv_cache/transforms_base.py
```diff
@@ -97,6 +98,76 @@ def find_nodes_by_pattern(
...
+            input. The ordering of nodes is important. The first node in the list
+            must take the causal mask as input. The last node in the list will be
+            connected to the `causal_mask_input_child`.
+        :param initializers_to_add: an optional list of initializers to add to the model
+        :return: model with the causal mask adjusted
+        """
+
+        graph = ONNXGraph(model)
+
+        if nodes_to_add[0].input[0] != self.CAUSAL_MASK_NAME:
```
bfineran, line 144: lgtm overall - let's discuss tomorrow if there's a way we can clean up this flow a bit. needing to format this list arg in a certain way seems a bit like hidden knowledge here

mgoin (commented):
FYI I've tested this with MPT instruct, chat, and replit code to much success

bfineran (approved):

mgoin (approved):

PR #1698, opened 2023-08-02
Enable running pytorch.torchvision.train with distributed data parallel

ohaijen:
Two main changes to enable running the following command in DistributedDataParallel mode:
```
CUDA_VISIBLE_DEVICES=<GPUs> python -m torch.distributed.launch \
 --nproc_per_node <NUM GPUs> \
sparseml.torchvision.train  <TRAIN.PY ARGUMENTS>
```
Before, it would throw an error (if run as above) or run in DataParallel mode (if run from CLI). After this change, running from CLI will still run DP, but running as above will run error-free with DDP, which is generally the desired outcome.

Concretely, the following changes were necessary: 
- Accept local_rank argument set by torch.distributed.launch
- in case of distributed/DDP, ensure model gets sent to single GPU

Additionally, update documentation to explain how to run with DDP.


src/sparseml/pytorch/torchvision/train.py
```diff
@@ -800,7 +811,16 @@ def _create_model(
         raise ValueError(
             f"Unable to find {arch_key} in ModelRegistry or in torchvision.models"
         )
-    model, device, _ = model_to_device(model=model, device=device)
+    ddp = False
+    if local_rank is not None:
+        torch.cuda.set_device(local_rank)
```
bfineran, line 816: nice - looks like it will make a lot of our defaults work out of the box

PR #1700, opened 2023-08-10
[bugfix] don't allow qlinearmatmul to match output Q/DQ block meant for next quantizable node

bfineran:
currently, if a Q/DQ -> MatMul -> Q/DQ -> MatMul series of nodes occurs, the first matmul may be matched as a target for a `QLinearMatMul` conversion even though the output Q/DQ block is meant for the second matmul

this adds a simple check to skip these cases. this issue was brought up in testing of quantizing distilbert by @rahul-tuli 

**test-plan:**
to be tested by @rahul-tuli


rahul-tuli:
[Noting] the commits from this diff have been verified and cherrypicked to #1701

bfineran:
#1701 rebases onto this - closing

PR #1701, opened 2023-08-10
Fix channelwise export

rahul-tuli:
This diff fixes ONNXToDeepSparse QDQ to Quantized graph conversion;

The problems were two fold:
 - Axis was always assumed to be zero; while this works for Convs it will fail for MatMuls (axis = 1) and Gemms (axis selection based on `transA` or `transB` attribute)
 - MatMul node matching errors where some matmuls were incorrectly converted to QlinearMatMuls; fixed by @bfineran in diff #1700 and the commit is cherry-picked in this PR
 
 Test plan(Manual):
 
 Previously failing conversions distilbert qdq and codegen qdq now complete successfully; the generated models have been shared on the `/network`
 
 -> August 14th, 2023 (Monday EST: 14:30 PM) Manually verified QDQ -> Quant conversion using a distilbert model and ran it through deepsparse.benchmark
 
 Test Code:
 ```python
import onnx
from sparseml.exporters.onnx_to_deepsparse import ONNXToDeepsparse

model_path = "/home/rahul/experiments/distilbert/int4-channel-ptq/deployment-qdq/model.onnx"

model = onnx.load(model_path)

exporter = ONNXToDeepsparse()
exporter.apply(model)

new_model_path = "/home/rahul/folded-distil-test.onnx"
onnx.save(model, new_model_path)
```

### DIFF
--- a/src/sparseml/exporters/transforms/utils/add_quantized_conv_matmul_add_ops.py
+++ b/src/sparseml/exporters/transforms/utils/add_quantized_conv_matmul_add_ops.py
@@ -28,6 +28,44 @@ from sparseml.exporters.transforms.utils.helpers import (
 __all__ = ["add_quantized_conv_matmul_add_ops"]
 
 
+def _get_axis(node, input_quantize_node):
+    """
+    return: the axis to be used for channel-wise QDQ -> Quantized
+        folding for Gemm, MatMul and Conv Ops
+    """
+    gemm_input_index_to_attribute = {
+        0: "transA",
+        1: "transB",
+    }
+
+    if node.op_type == "Conv":
+        return 0
+
+    if node.op_type == "MatMul":
+        # axis is always 1 for MatMul
+        return 1
+
+    axis = 1  # default axis is 1
+    target_id_index = int(input_quantize_node.name.split(".")[-2])
+    is_input = "input" in input_quantize_node.name
+
+    if (
+        node.op_type == "Gemm"
+        and is_input
+        and target_id_index in gemm_input_index_to_attribute
+    ):
+
+        attribute_name = gemm_input_index_to_attribute[target_id_index]
+
+        attr = next(
+            (attr for attr in node.attribute if attr.name == attribute_name), None
+        )
+        if attr is None:
+            return 0
+        axis = int(not bool(attr.i if attr else 0))
+    return axis
+
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/sparseml/exporters/transforms/utils/add_quantized_conv_matmul_add_ops.py [hunk 1]
```diff
@@ -28,6 +28,44 @@
 __all__ = ["add_quantized_conv_matmul_add_ops"]
 
 
+def _get_axis(node, input_quantize_node):
```
Satrat, line 31: Nice helper function!

Satrat (approved):

src/sparseml/exporters/transforms/utils/add_quantized_conv_matmul_add_ops.py
```diff
@@ -28,6 +28,44 @@
...
+    if node.op_type == "Conv":
+        return 0
+
+    if node.op_type == "MatMul":
+        # axis is always 1 for MatMul
+        return 1
+
+    axis = 1  # default axis is 1
+    target_id_index = int(input_quantize_node.name.split(".")[-2])
+    is_input = "input" in input_quantize_node.name
```
bfineran, line 50: this logic is specific to Q/DQs added by our sparsify one shot flow only - should instead find the index based on the DQ node - regardless, the weight will always be the "B" matrix so we shouldn't need this

src/sparseml/exporters/transforms/utils/add_quantized_conv_matmul_add_ops.py
```diff
@@ -28,6 +28,57 @@
...
+        if attr is None:
+            return 0
+
+        # axis is 0 if transB is 1
+        #  else 1
+        return int(not bool(attr.i))
+
+    # default axis is 1 based on ONNX spec
+    #  https://github.com/onnx/onnx/blob/main/docs/Operators.md#QuantizeLinear
+    return 1
```
Satrat, line 79: Could we add in a unit test for this case since its a bit convoluted?

Satrat (approved):
LGTM, I suggested a test case for the Gemm case but didn't want to block the merge over it

rahul-tuli:
Closing this out as it's been deprioritized

PR #1714, opened 2023-09-05
Pytorch OBCQ Implementation

Satrat:
This PR adds a PyTorch implementation of OBCQ/SparseGPT to SparseML. It is MUCH faster than the ONNX implementation (hours vs minutes to run) but also much less robust to different models and architectures. This only contains the OPT implementation, LLaMA and MPT will be in a separate PR

### How to Run
An example recipe file is located at `src/sparseml/pytorch/sparsification/obcq/example.yaml`
```
python src/sparseml/pytorch/sparsification/obcq/obcq.py facebook/opt-350m c4 --recipe src/sparseml/pytorch/sparsification/obcq/example.yaml 
```

### Notes
* Start in `src/sparseml/transformers/sparsification/obcq`
* Research implementation of the algorithm is in src/sparseml/experimental and doesn't need to be reviewed (it'll be deleted before this branch gets merged)


src/sparseml/transformers/sparsification/obcq/sparse_gpt_modifier.py
```diff
@@ -0,0 +1,130 @@
...
+
+@PyTorchModifierYAML()
+class SparseGPTModifier(BaseModifier):
+    def __init__(
+        self,
+        sparsity: float = 0.5,
+        block_size: int = 128,
+        quantize: bool = True,
+        dampening_frac: Optional[float] = 0.01,
+        sequential_update: Optional[bool] = True,
```
bfineran, line 37: for serialization, need to mark these as `@ModifierProp`s (see other modifiers for example). also good to add a serialization test (see other modifiers for example)

src/sparseml/transformers/sparsification/obcq/sparse_gpt_modifier.py
```diff
@@ -0,0 +1,130 @@
...
+    def compressible_layers(self):
+        raise NotImplementedError
+
+    def bottom_compressor(self):
+        raise NotImplementedError
+
+    def head_compressor(self):
+        raise NotImplementedError
+
+    def one_shot(self, model, dataloader, initializer_kwargs, finalize_kwargs):
```
bfineran, line 64: would be great to move all this functionality into just initialize and finalize, so at the generic modifier/manager level, one shot can be applied by just calling these two

bfineran (changes requested):
LGTM overall - since this runs in one shot and doesn't follow the normal training lifecycle the other pytorch based modifiers do, the main missing piece is serialization.  The code/names look a bit more on the research side (ie mixed case, abbreviations, generic names, etc) and some doc strings are missing.

src/sparseml/transformers/sparsification/obcq/layer_compressor.py
```diff
@@ -0,0 +1,159 @@
...
+        # Run through the samples in order to compute Hessian matrix
+        nsamples = self.inputs.shape[0]
+        attention_mask = kwargs.get("attention_mask", None)
+        for j in range(nsamples):
+            attn_mask = (
+                attention_mask[j]
+                if isinstance(attention_mask, List)
+                else attention_mask
+            )
+            self.layer(self.inputs[j].unsqueeze(0), attention_mask=attn_mask)[0]
```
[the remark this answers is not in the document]
Satrat: I'm not sure actually, I pulled this line from Tuan's code so let me check with him

PR #1719, opened 2023-09-08
SparseZoo V2 Bug Fixes

Satrat:
Fixing issues caused by the upgrade to SparseZoo V2:
1. ~~`Model` class stores recipes in a list, updated `optim/helpers.py` to reflect this change~~ This was handled instead by updating SparseZoo to restore the default functionality: https://github.com/neuralmagic/sparsezoo/commit/f2cf607e840850e402820df08be9960b2587d35e
2. Changed multiple filename searches to use `-` instead of `_` to reflect change in sparsezoo
3. Removed deprecated test case that searches for `original.md` recipe (these have been renamed to `recipe.md`). Replaced with a test that searches for a non-default recipe name


src/sparseml/optim/helpers.py
```diff
@@ -401,7 +401,10 @@ def _load_yaml_str_from_file(file_path: Union[str, File]) -> str:
     if file_path.startswith("zoo:"):
         # download from zoo stub
         model = Model(file_path)
-        file_path = model.recipes.default.path
+        recipe = model.recipes
+        if isinstance(recipe, List):
```
bfineran, line 405: is the recipe ever not a list? i believe this is what `default` was supposed to handle - selecting the correct recipe out of the list of possible ones. is `default` no longer supported?

Satrat: In the current sparsezoo code `self.recipes` is always a list, see https://github.com/neuralmagic/sparsezoo/blob/main/src/sparsezoo/model/model.py#L153. Some of the unit tests were throwing errors because `model.recipes.default` is no longer valid

bfineran (dismissed):
marking OK to land, however we need to get `recipes.default` back to supported so that `?recipe_type` in stubs will work reliably

rahul-tuli (approved):

horheynm (approved):

PR #1730, opened 2023-09-21
Update transformers/__init__.py to warn about nm_transformers

mgoin:
This would fail if the attribute wasn't present rather than warn

```
  File "sparseml/src/sparseml/transformers/__init__.py", line 43, in _check_transformers_install
    if not _transformers.NM_INTEGRATED:
  File "lib/python3.10/site-packages/transformers/utils/import_utils.py", line 1177, in __getattr__
    raise AttributeError(f"module {self.__name__} has no attribute {name}")
AttributeError: module transformers has no attribute NM_INTEGRATED
```

### DIFF
--- a/src/sparseml/transformers/__init__.py
+++ b/src/sparseml/transformers/__init__.py
@@ -40,11 +40,10 @@ def _check_transformers_install():
     # check for NM integration in transformers version
     import transformers as _transformers
 
-    if not _transformers.NM_INTEGRATED:
+    if not getattr(_transformers, "NM_INTEGRATED", False):
         _LOGGER.warning(
-            "the neuralmagic fork of transformers may not be installed. it can be "
-            "installed via "
-            f"`pip install {_NM_TRANSFORMERS_NIGHTLY}`"
+            "The neuralmagic fork of transformers may not be installed. It can be "
+            "installed via `pip install nm_transformers`"
         )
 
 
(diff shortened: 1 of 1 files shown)


src/sparseml/transformers/__init__.py
```diff
@@ -40,11 +40,10 @@ def _check_transformers_install():
...
     import transformers as _transformers
 
-    if not _transformers.NM_INTEGRATED:
+    if not getattr(_transformers, "NM_INTEGRATED", False):
         _LOGGER.warning(
-            "the neuralmagic fork of transformers may not be installed. it can be "
-            "installed via "
-            f"`pip install {_NM_TRANSFORMERS_NIGHTLY}`"
+            "The neuralmagic fork of transformers may not be installed. It can be "
+            "installed via `pip install nm_transformers`"
```
dbarbuzzi, line 46: nit: the PyPI package name is `nm-transformers` (hyphen, not underscore). Using an underscore instead will work with `pip install` though I think matching the name more accurately is better for consistency/etc.

bfineran (changes requested):
Should raise a bit louder of a warning here since this would mean that the quantized matmuls would not be applied which is a significant difference for sparseml flows

dbogunowicz (approved):
Can we have the same diff for deepsparse?

mgoin:
> Can we have the same diff for deepsparse?

Deepsparse already has a fix to properly access the attribute and I think the critical-ness of using nm-transformers with deepsparse is much less than sparseml. Happy to escalate deepsparse's warning if y'all think it's warranted

PR #1739, opened 2023-09-27
Deprecate SSD Resnet Tests

Satrat:
These models are deprecated and no longer on SparseZoo as of V2, so we can skip this failing test


rahul-tuli (approved):
Ship it!

bfineran (approved):

PR #1740, opened 2023-09-28
Change scheme overrides priority

anmarques:
This PR changes scheme override priority to match first the module name and ten the module type. This allows users to combine both forms of scheme overrides in the same recipe in a useful way.

For instance, a user can override the default behavior for linear layers, but then define a different behavior for specific layers by defining overrides by name. This wasn't possible before


bfineran (dismissed):
names are more granular so I am on board with this

rahul-tuli (approved):

rahul-tuli (approved):

Satrat (approved):

PR #1743, opened 2023-10-02
SmoothQuant integration to SparseGPT

anmarques:
- Adds native SmoothQuant integration to SparseGPT (no longer depends on the original smoothquant library)
- Tested on OPT 1.3b
- Added support to logarithmic activation equalization, a method similar to smoothquant introduced in this paper [https://arxiv.org/abs/2308.15987](https://arxiv.org/abs/2308.15987))
- Built on top of the Llama2 integration


anmarques:
Implementation landed on main

PR #1746, opened 2023-10-03
[ONNXTransform] always delete orphaned node branches after running an onnx transform

bfineran:
adds a pass in the base `ONNXTransform` cleanup steps to run `ONNXGraph.delete_orphaned_node_branches` which is an important additional cleanup step.  This removes unwanted nodes from remaining in the graph after a transform which could lead to issues at inference time.  This function should be safe to run against all transforms and only adds minimal overhead (`O(nodes^2)`) if any orphaned nodes are found

should fix issues in KVCache injected MPT models which seem to consistently get shipped with orphaned nodes remaining

**test_plan:**
verified locally that this unblocked running a pipeline with `zoo:nlg/text_generation/mpt-7b/pytorch/huggingface/mpt_chat/base-none`

future testing should pull down a model with no data files and test injection against it

### DIFF
--- a/src/sparseml/exporters/transforms/onnx_transform.py
+++ b/src/sparseml/exporters/transforms/onnx_transform.py
@@ -102,6 +102,7 @@ class OnnxTransform(BaseTransform):
             model.graph.node.remove(node)
         graph = ONNXGraph(model)
         graph.delete_unused_initializers()
+        graph.delete_orphaned_node_branches()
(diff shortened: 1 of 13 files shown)


anmarques (approved):
Tested on mpt-7b-mpt_pretrain-base and it worked to produce a valid model

bfineran:
tests fixed, merging

PR #1750, opened 2023-10-06
[DO NOT MERGE] Add script to reveal bug in ConstantPruningModifier

rahul-tuli:
Run:
```bash
python bug_constant_pruning.py     
```


rahul-tuli:
The Bug was caught and fixed as a part of #1751 , verified the bug fix works!

PR #1751, opened 2023-10-06
Fix Constant Pruning Modifier

Satrat:
The mask calculation in the constant pruning modifier was inverted, this one character PR fixes it!

### Testing
Constant pruning modifier is tested in this branch: https://github.com/neuralmagic/sparseml/tree/modifier-refactor-constant-pruning-bug

Run `python bug_constant_pruning` to confirm constant pruning modifier is working as intended


rahul-tuli (approved):

bfineran (approved):

rahul-tuli:
Let's Go!!

PR #1752, opened 2023-10-09
[modifier refactor] Add constant pruning tests 

rahul-tuli:
This PR serves as the base branch for all `ConstantPruningModifier` tests:

- [X] End to End tests
- [X] Serialization Bugfix #1753 
- [X] Serialization Test #1755 
- [X] Unit Tests #1756 
- [X] Mask De-registration Bugfix #1773 
- [X] Session reset for all tests #1790


tests/sparseml/core/recipe/test_recipe.py
```diff
@@ -14,30 +14,78 @@
...
+@pytest.mark.parametrize("recipe_str", _valid_recipes())
+def test_serialization(recipe_str):
+    recipe_instance = Recipe.create_instance(recipe_str)
+    serialized_recipe_str = recipe_instance.yaml()
+    recipe_from_serialized = Recipe.create_instance(serialized_recipe_str)
+
+    expected_dict = recipe_instance.dict()
+    actual_dict = recipe_from_serialized.dict()
+
+    assert expected_dict == actual_dict
```
Satrat, line 91: could we test initializing from a dict and from explicitly defining the arguments as well? I believe a lot of the other serialization tests do this

tests/sparseml/modifiers/pruning/constant/test_pytorch.py
```diff
@@ -0,0 +1,173 @@
...
+    # move imports inside so that pytorch is not required unless
+    # running this test
+
+    from sparseml.core import State
+    from sparseml.core.event import Event, EventType
+    from sparseml.core.framework import Framework
+    from sparseml.modifiers.pruning.constant.pytorch import (
+        ConstantPruningModifierPyTorch,
+    )
+    from sparseml.pytorch.utils import tensor_sparsity
```
Satrat, line 88: Since the Makefile is set to ignore this file except when we run the pytorch tests can't we move these to the top of the file?

Makefile
```diff
@@ -25,7 +25,7 @@ ifneq ($(findstring onnx,$(TARGETS)),onnx)
     PYTEST_ARGS := $(PYTEST_ARGS) --ignore tests/sparseml/onnx
 endif
 ifneq ($(findstring pytorch,$(TARGETS)),pytorch)
-    PYTEST_ARGS := $(PYTEST_ARGS) --ignore tests/sparseml/pytorch
+    PYTEST_ARGS := $(PYTEST_ARGS) --ignore tests/sparseml/pytorch --ignore tests/sparseml/modifiers
```
Satrat, line 28: We should think about a better way to separate out the pytorch tests, because eventually we will have other frameworks in this folder. This will also skip any framework-independent base tests we write

tests/sparseml/modifiers/pruning/constant/test_pytorch.py
```diff
@@ -0,0 +1,173 @@
...
+        if "weight" in name
+    }
+    assert manipulated_sparsities != expected_sparsities, "Sparsity manipulation failed"
+
+    # apply modifier
+
+    modifier.on_update(state, event=Event(type_=EventType.OPTIM_PRE_STEP))
+    modifier.on_update(state, event=Event(type_=EventType.OPTIM_POST_STEP))
+    modifier.finalize(state)
+    modifier.on_end(state, None)
```
Satrat, line 127: these should be called in the opposite order

tests/sparseml/modifiers/pruning/constant/test_pytorch.py
```diff
@@ -0,0 +1,173 @@
...
+        optimizer=optimizer(model.parameters(), lr=0.1),
+        start=0,
+    )
+    modifier = ConstantPruningModifierPyTorch(
+        targets="__ALL_PRUNABLE__",
+        start=0,
+        end=1,
+        update=0.5,
+    )
+    modifier.initialize(state)
```
Satrat, line 110: Could we add a check after this call to make sure the masks have been added to the model and that they match up with the 0 locations?

tests/sparseml/modifiers/pruning/constant/test_pytorch.py
```diff
@@ -0,0 +1,173 @@
...
+        for name, param in model.named_parameters()
+        if "weight" in name
+    }
+    assert manipulated_sparsities != expected_sparsities, "Sparsity manipulation failed"
+
+    # apply modifier
+
+    modifier.on_update(state, event=Event(type_=EventType.OPTIM_PRE_STEP))
+    modifier.on_update(state, event=Event(type_=EventType.OPTIM_POST_STEP))
+    modifier.finalize(state)
```
Satrat, line 126: Lets test that the masks were actually removed after this call

tests/sparseml/modifiers/pruning/constant/test_pytorch.py
```diff
@@ -0,0 +1,173 @@
...
+    assert isinstance(
+        type_, ConstantPruningModifierPyTorch
+    ), "PyTorch ConstantPruningModifier not registered"
+
+
+def _setup_modifier_factory():
+    from sparseml.core.factory import ModifierFactory
+
+    ModifierFactory.refresh()
+    assert ModifierFactory._loaded, "ModifierFactory not loaded"
```
Satrat, line 173: This seems like it would be a useful helper function outside of this class, maybe we could move it to a general test helpers file

Satrat (changes requested):
Would be good to test the mask is correctly being added and removed(see comments), but otherwise this is looking good!

Makefile
rahul-tuli: Updated to use a regex for files inside sparseml/modifiers

tests/sparseml/modifiers/pruning/constant/test_pytorch.py
rahul-tuli: Done!

tests/sparseml/modifiers/pruning/constant/test_pytorch.py
rahul-tuli: Done!

tests/sparseml/modifiers/pruning/constant/test_pytorch.py
rahul-tuli: Agree moving this to a helper!

tests/sparseml/modifiers/pruning/constant/test_pytorch.py
rahul-tuli: Done!, however it will throw an error now if we do `pytest --collect-only`, works with makefile though with the regex change.

tests/sparseml/core/recipe/test_recipe.py
rahul-tuli: Added as a part of #1774

Makefile
```diff
@@ -25,7 +25,7 @@ ifneq ($(findstring onnx,$(TARGETS)),onnx)
     PYTEST_ARGS := $(PYTEST_ARGS) --ignore tests/sparseml/onnx
 endif
 ifneq ($(findstring pytorch,$(TARGETS)),pytorch)
-    PYTEST_ARGS := $(PYTEST_ARGS) --ignore tests/sparseml/pytorch --ignore tests/sparseml/modifiers
+    PYTEST_ARGS := $(PYTEST_ARGS) --ignore tests/sparseml/pytorch --ignore-glob 'tests/sparseml/modifiers/*pytorch.py'
```
Satrat, line 28: I think this is going to clash with how I defined the tests for the quantization modifier. Lets sync up on the testing structure before this gets merged in

src/sparseml/modifiers/obcq/utils/sparsegpt.py
```diff
@@ -18,7 +18,13 @@
...
 import torch
 import torch.nn as nn
-import transformers
+
+
+try:
+    import transformers
+except ImportError as err:
+    transformers = None
+    transformers_err = err
```
Satrat, line 27: I think this is also going to clash with a change I made, but I like this solution better! In pair programming today lets go through the merge conflicts

tests/sparseml/helpers.py
```diff
@@ -0,0 +1,82 @@
...
+    ]
+
+
+def should_skip_pytorch_tests():
+    try:
+        import torch  # noqa: F401
+    except ImportError:
+        return True
+
+    return os.getenv("NM_ML_SKIP_PYTORCH_TESTS", False)
```
Satrat, line 82: Why do we need both this and the change to the makefile?

Satrat (commented):
This looks good, but I think our branches have diverged a bit of testing layout. Lets figure it out in pair programming today then get this merged

tests/sparseml/helpers.py
rahul-tuli: The glob in makefile only ignores the modifiers/*pytorch.py files; this was done because a few tests in other files were being picked up in base tests, like a few in test_state.py

Makefile
rahul-tuli: Updated!

src/sparseml/modifiers/obcq/utils/sparsegpt.py
rahul-tuli: updated!

bfineran (dismissed):

rahul-tuli:
The failing test `tests/sparseml/pytorch/modifiers/quantization/test_pytorch.py` is fixed by https://github.com/neuralmagic/sparseml/pull/1790

Satrat (approved):
Looks great pending the session reset PR getting merged!

bfineran (approved):

PR #1758, opened 2023-10-11
SmoothQuant Modifier for OBCQ

Satrat:
This PR implements a productionized SmoothQuantModifier for use in OBCQ, with example recipes for OPT and Llama. 

### Instructions and Testing
To run for OPT, and print out perplexity results
```
python src/sparseml/transformers/sparsification/obcq/obcq.py facebook/opt-1.3b c4
    --recipe src/sparseml/transformers/sparsification/obcq/example.yaml --eval wikitext
```

To run for Llama, you'll need a local copy of the model 
```
python src/sparseml/transformers/sparsification/obcq/obcq.py /local/path/to/llama open_platypus
    --recipe src/sparseml/transformers/sparsification/obcq/example_llama.yaml --eval open_platypus
```


src/sparseml/modifiers/__init__.py
```diff
@@ -18,3 +18,4 @@
 from .obcq import *
 from .pruning import *
 from .quantization import *
+from .smoothquant import *
```
bfineran, line 21: nit: no new line

src/sparseml/modifiers/smoothquant/base.py
```diff
@@ -0,0 +1,65 @@
...
+    logarithmic_equalization: Optional[bool] = False
+    num_calibration_steps: Optional[int] = None
+
+    def on_initialize_structure(self, state: "State", **kwargs):
+        pass  # nothing needed for this modifier
+
+    def on_start(self, state: State, event: Event, **kwargs):
+        pass
+
+    def on_update(self, state: State, event: Event, **kwargs):
```
bfineran, line 58: wonder how much we can do with modifiable models and datasets along with shared quant schemes to get a more generic impl of algorithms like this between torch and onnx

src/sparseml/modifiers/smoothquant/pytorch.py
```diff
@@ -0,0 +1,269 @@
...
+
+            return hook_fn
+
+        for mapping in self.resolved_mappings_:
+            name = mapping.smooth_name
+            layer = mapping.smooth_layer
+            self.hooks_.append(layer.register_forward_hook(create_hook_fn(name)))
+
+    @torch.no_grad()
+    def _calibrate(self, model: ModifiableModelPyTorch, calibration_dataloader: List):
```
bfineran, line 190: probably can move this out to a helper function, we do the same in base quantization modifier right?

src/sparseml/modifiers/smoothquant/pytorch.py
```diff
@@ -0,0 +1,269 @@
...
+        # remove the hooks now that we are done calibrating
+        for hook in self.hooks_:
+            hook.remove()
+        del self.hooks_
+
+    @torch.no_grad()
+    def _apply_smoothing(self):
+        """
+        After calibration, apply smoothing to the activations and push the transform
+        into the following weights by applying the inverse to each balance weight.
```
bfineran, line 234: let's add the smoothing function in this docstring if we can

bfineran (commented):
LGTM - would be great to add some simple unit tests

src/sparseml/modifiers/smoothquant/base.py
Satrat: Yeah I think theres definitely more to be done both here and in some of the other modifiers to make them more generic. I just pushed an update that moves the resolve_mapping function and most of initialization to the base class, and I think with a bit more thought we could move the apply_smoothing over too. The calibration function and the hook setup seem pretty solidly pytorch specific though

src/sparseml/modifiers/smoothquant/base.py
```diff
@@ -0,0 +1,170 @@
...
+    :param migration_strength: Intensity of smoothing to perform (0-1 range)
+    :param mappings: list activation layers to smooth, and the which layers to offset
+    the smoothing to for each activation
+    :param ignore: list of layers to ignore, even if they match a regex in mappings
+    :param logarithmic_equalization: Whether to use a logarithmic scale for smoothing
+    :param num_calibration_steps: number of samples to use for calibration, or None to
+    use the whole dataset
+    """
+
+    migration_strength: float
```
anmarques, line 82: I don't really like the term "migration_strength". How about "smoothing_strength"?

I know "alpha" is completely arbitrary, but at least it matches the paper's definition. Can we keep it as an alias? Also we should reference it in the docstring.

bfineran (dismissed):

src/sparseml/modifiers/smoothquant/base.py
Satrat: done!

src/sparseml/modifiers/smoothquant/base.py
```diff
@@ -0,0 +1,172 @@
...
+    :param ignore: list of layers to ignore, even if they match a regex in mappings
+    :param logarithmic_equalization: Whether to use a logarithmic scale for smoothing
+    :param num_calibration_steps: number of samples to use for calibration, or None to
+    use the whole dataset
+    """
+
+    smoothing_strength: float = Field(..., alias="alpha")
+    mappings: List[Tuple]
+    ignore: Optional[List[str]] = None
+    logarithmic_equalization: Optional[bool] = False
```
anmarques, line 87: It seems that the logarithmic_equalization argument is not being used in this implementation. That's fine. I actually suggest that you remove this argument from SmoothQuantModifier.

The only difference between smoothquant and logarithmic equalization is how the smoothing scale is computed. I suggest that you move the smoothing scale computation into a separate method, and then we can implement LogarithmicEqualizationModifier inheriting from SmoothQuantModifier and just changing this method.

src/sparseml/modifiers/smoothquant/pytorch.py
```diff
@@ -0,0 +1,217 @@
...
+        _LOGGER.info("Smoothing activation scales...")
+        for mapping in self.resolved_mappings_:
+            activation_scales = (  # get dynamic range for each activation channel
+                self.scales_[mapping.smooth_name].max_channel_vals
+                - self.scales_[mapping.smooth_name].min_channel_vals
+            )
+            smooth_layer = mapping.smooth_layer
+            balance_layers = mapping.balance_layers
+
+            # get the channel-wise dynamic range for each layer to be balanced
```
anmarques, line 193: This section between lines 192 and 205 is the computation of smoothing scales I referred to in the comment above. I suggest moving this to a separate method so we can override it when implementing LogarithmicEqualizationModifier.

src/sparseml/modifiers/smoothquant/base.py
anmarques: BTW, LogarithmicEqualizationModifier can be implemented in a different PR

anmarques (changes requested):

src/sparseml/modifiers/smoothquant/base.py
```diff
@@ -0,0 +1,172 @@
...
+    :param smoothing_strength: alpha, intensity of smoothing to perform (0-1 range)
+    :param mappings: list activation layers to smooth, and the which layers to offset
+    the smoothing to for each activation
+    :param ignore: list of layers to ignore, even if they match a regex in mappings
+    :param logarithmic_equalization: Whether to use a logarithmic scale for smoothing
+    :param num_calibration_steps: number of samples to use for calibration, or None to
+    use the whole dataset
+    """
+
+    smoothing_strength: float = Field(..., alias="alpha")
```
rahul-tuli, line 84: is the `...` here on purpose?

src/sparseml/modifiers/smoothquant/base.py
```diff
@@ -0,0 +1,172 @@
...
+    Implements the SmoothQuant algorithm from https://arxiv.org/abs/2211.10438. This
+    modifier performs a channel-wise smoothing of outliers in activations, making them
+    easier to quantize by reducing the dynamic range. The smoothing is offset by
+    applying the inverse operation to the next layer of weights, making the weights
+    slightly more difficult to quantize.
+
+    Because this modifier manipulates the weights of the model, it can only be used in
+    in one-shot and not during training. Activation ranges are determined by running a
+    small set of calibration data through the model.
+
```
rahul-tuli, line 74: Could we add in a small example yaml in the docstring?

src/sparseml/modifiers/smoothquant/base.py
rahul-tuli: Also adding an alias would mean 
- it accepts both smoothing_strength and alpha in recipes?, or
- it accepts just alpha?

src/sparseml/modifiers/smoothquant/base.py
```diff
@@ -0,0 +1,172 @@
...
+                " to be None or -1, got {}".format(self.end)
+            )
+        if self.start and self.start != -1:
+            raise ValueError(
+                "SmoothQuantModifier can only be applied during one-shot. Expected "
+                "start to be None or -1, got {}".format(self.start)
+            )
+
+        self.ignore = [] if not self.ignore else self.ignore
+        self.resolved_mappings_ = self._resolve_mappings(state.model)
```
rahul-tuli, line 115: resolved_mappings will be a list here? should we update the typing?

src/sparseml/modifiers/smoothquant/base.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def on_finalize(self, state: State, **kwargs) -> bool:
+        """
+        Clean up by clearing the scale and mapping data
+
+        :param state: unused
+        :return: True
+        """
+        self.scales_.clear()
+        self.resolved_mappings_.clear()
```
rahul-tuli, line 170: what about `None`?

src/sparseml/modifiers/smoothquant/pytorch.py
```diff
@@ -0,0 +1,217 @@
...
+                        )
+                        if balance_layer:
+                            balance_layers.append(balance_layer)
+                    # each mapping can contain multiple layers to balance, but only
+                    # one layer to smooth
+                    mapping = SmoothQuantMapping(
+                        layer_name, smooth_layer, balance_layers
+                    )
+                    resolved_mappings.append(mapping)
+        return resolved_mappings
```
rahul-tuli, line 108: Is this the same as the one in base class but wrapped in torch.no_grad? if that's the case we could avoid duplicacy by doing something like `func_to_wrap = wrapper(func_to_wrap)`?

rahul-tuli (dismissed):

src/sparseml/modifiers/smoothquant/pytorch.py
Satrat: Ah good catch, this function doesn't actually need torch.no_grad() to run, I fully ported it over to base.py and forgot to delete to pytorch one

src/sparseml/modifiers/smoothquant/base.py
Satrat: I updated this because it should actually be validation_alias, we will accept either alpha or smoothing_strength. Alpha is the term used in the smooth quant paper but its not a very descriptive name

src/sparseml/core/model/pytorch.py
```diff
@@ -94,3 +95,13 @@ def set_param(self, target: str, param: Parameter):
         :param param: the parameter to set
         """
         return set_param(target, param, self.model)
+
+    def get_matching_layer(
+        self, target: str, name_to_match: str, model: Module
+    ) -> Optional[Tuple[str, Module]]:
+        """
+        :param target: layer name to target when searching model
```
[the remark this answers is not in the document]
Satrat: Yep good call, the actual function mentions that target should be a regex but it makes sense to have it here too

anmarques (approved):

bfineran (approved):

PR #1766, opened 2023-10-16
Add kvcache config for Mistral

mgoin:
```
git clone https://huggingface.co/HuggingFaceH4/zephyr-7b-alpha
sparseml.transformers.export_onnx --model_path ./zephyr-7b-alpha --task text-generation --sequence_length 512 --trust_remote_code
cp deployment/model.onnx deployment/model-orig.onnx
python ~/onnx_kv_inject.py --input-file deployment/model-orig.onnx --output-file deployment/model.onnx
```

```
>>> from deepsparse import TextGeneration
>>> model = TextGeneration(model="deployment")
>>> out = model("Once upon a time,", max_new_tokens=100)
>>> out.generations[0].text
'there was a young woman named Lily. She was a kind and gentle soul, with a heart full of love and compassion. Lily had always been fascinated by the natural world, and she spent most of her free time exploring the forests and fields around her home.\n\nOne day, as she was wandering through the woods, Lily stumbled upon a small clearing. In the center of the clearing, she saw a beautiful butterfly fluttering its wings. The butterfly was unlike any she had'
```

HF Baseline :) 
Screenshot 2023-10-16 at 2 29 51 PM

### DIFF
--- a/src/sparseml/exporters/transforms/kv_cache/configs.py
+++ b/src/sparseml/exporters/transforms/kv_cache/configs.py
@@ -138,6 +138,11 @@ LLAMA_CONFIG = KeyValueCacheConfig(
     multiply_batch_by_num_att_heads=False,
 )
 
+# Mistral has a new config/model definition "MistralForCausalLM" but is heavily based off Llama2.
+# It contains these additions to Llama2-7b:
+# * Sliding Window Attention - Trained with 8k context length and fixed cache size, with a theoretical attention span of 128K tokens
+# * GQA (Grouped Query Attention) - allowing faster inference and lower cache size.
+# * Byte-fallback BPE tokenizer - ensures that characters are never mapped to out of vocabulary tokens.
 MISTRAL_CONFIG = KeyValueCacheConfig(
     model_name="mistral",
     additional_transforms=AdditionalTransformsLLAMA,
--- a/src/sparseml/exporters/transforms/kv_cache/configs.py
+++ b/src/sparseml/exporters/transforms/kv_cache/configs.py
@@ -138,6 +138,16 @@ LLAMA_CONFIG = KeyValueCacheConfig(
     multiply_batch_by_num_att_heads=False,
 )
 
+MISTRAL_CONFIG = KeyValueCacheConfig(
+    model_name="mistral",
+    additional_transforms=AdditionalTransformsLLAMA,
+    key_num_attention_heads="num_attention_heads",
+    key_num_embedding_hidden_size="hidden_size",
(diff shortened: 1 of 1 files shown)


dbogunowicz (approved):

dsikka (changes requested):
Can you add an explanation/comment explaining why this is needed and how this differs than the other llama models? 

Also, will this work for the 70b model? In that case, can we update the name of the config?

src/sparseml/exporters/transforms/kv_cache/configs.py
```diff
@@ -138,6 +138,16 @@ class Config:
     multiply_batch_by_num_att_heads=False,
 )
 
+MISTRAL_CONFIG = KeyValueCacheConfig(
+    model_name="mistral",
+    additional_transforms=AdditionalTransformsLLAMA,
```
dsikka, line 143: A warning will show up every time about the number of updated nodes if they aren't the standard, just an fyi

mgoin:
> @dsikka it should work for 70b model, but I am not sure whether @mgoin has the capacity to test it. I'd take a leap of faith and merge it regardless.

Mistral actually uses a "completely" new model/config class so the config.json actually holds a ["MistralForCausalLM"](https://huggingface.co/docs/transformers/main/en/model_doc/mistral#transformers.MistralForCausalLM) architecture. [Here is an example config](https://huggingface.co/mistralai/Mistral-7B-v0.1/blob/main/config.json). There is only one pretrained Mistral arch model and it is 7b.

Based on how we're currently architecting this, I assume we need to make a new config entry for every new `"model_type"` we see in `config.json` files. This makes the Mistral models not exactly connected to the Llama model support for GQA, aka the 70b model y'all are talking about. So it is disconnected, but I will put a comment that Mistral is very connected to Llama

src/sparseml/exporters/transforms/kv_cache/configs.py
mgoin: Yeah, we should fix this for all Llama models

dbogunowicz: @dsikka could you elaborate?

dsikka:
@mgoin from what we've seen though, it seems like the config you added could also be used for GQA?

mgoin:
@dsikka yes, it should work for Llama models with GQA if the name is changed to "llama". My position is that Llama GQA support is a different issue than Mistral support, but if y'all want to add it into this diff I'm good with that. I'm not sure how to structure the change without affecting non-GQA Llama models

PR #1769, opened 2023-10-17
ONNX Export for Refactored Modifiers

Satrat:
* Updated QuantizationModifier to apply structure before initialization
* New version of export script that is compatible with the new modifier and session framework.The main change is replacing Managers with SparseSessions

### Example
```
sparseml.transformers.export_onnx_refactor --task text-generation --model_path path/to/model/and/recipe
```
or
```
python src/transformers/sparsification/obcq/export.py  --task text-generation --model_path path/to/model/and/recipe
```

### DIFF
--- a/src/sparseml/transformers/sparsification/obcq/export.py
+++ b/src/sparseml/transformers/sparsification/obcq/export.py
@@ -70,6 +70,7 @@ import logging
 import os
 import shutil
 from typing import Any, Dict, List, Optional, Union
+import torch
 
 from torch.nn import Module
 from transformers import AutoConfig, AutoTokenizer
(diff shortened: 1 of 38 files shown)


src/sparseml/transformers/sparsification/obcq/export.py
```diff
@@ -0,0 +1,634 @@
...
+import os
+import shutil
+from typing import Any, Dict, List, Optional, Union
+
+import torch
+from torch.nn import Module
+from transformers import AutoConfig, AutoTokenizer
+from transformers.tokenization_utils_base import PaddingStrategy
+
+import sparseml.core.session as sml
```
bfineran, line 79: nit: we generally don't use import aliases, I think in this case it makes sense to alias, but maybe standardize to something a bit more explicit for readability (especially when splashing into other repos)

bfineran (approved):
LGTM - would potentially be good to add a way for the current default script `sparseml.transfomers.export_onnx` to route to this one if it detects one of the new recipes

mgoin (approved):

bfineran (approved):

PR #1773, opened 2023-10-18
[Bug][ConstantPruningModifier] Fix mask de register bug

rahul-tuli:
This PR fixes a bug where, buffers were being de-registered when mask was persistent (this should be reversed)
Also fixes the logic for de-registration; `unregister_buffer` raises an `AttributeError` on torch Modules
Also updates test to check buffer was actually de-registered

### DIFF
--- a/tests/sparseml/modifiers/pruning/constant/test_pytorch.py
+++ b/tests/sparseml/modifiers/pruning/constant/test_pytorch.py
@@ -68,7 +68,7 @@ def _test_optims():
 
     return [
         torch.optim.Adam,
-        torch.optim.SGD,
+        # torch.optim.SGD,
     ]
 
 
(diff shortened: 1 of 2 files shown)


tests/sparseml/modifiers/pruning/constant/test_pytorch.py
```diff
@@ -68,7 +68,7 @@ def _test_optims():
 
     return [
         torch.optim.Adam,
-        torch.optim.SGD,
+        # torch.optim.SGD,
```
rahul-tuli, line 71:
```suggestion
        torch.optim.SGD,
```

Satrat (approved):

bfineran (approved):

PR #1784, opened 2023-10-20
Enable IC runs w/ torchrun

anmarques:
The DDP fix implemented in this [PR](https://github.com/neuralmagic/sparseml/pull/1698) uses an older interface to pass the local rank for DDP that only works w/ torch.distributed.run.

This PR replaces this by passing the local rank via an environment variable, which is supported by both torch.distributed.run and the newer torchrun interfaces.


ohaijen (approved):
Thanks for the fix!

bfineran (approved):

PR #1789, opened 2023-10-24
[Test fix] Add stronger check for calibration_dataloader

rahul-tuli:
We are seeing some  weird behavior; There's a quantization test failing in GHA here: https://github.com/neuralmagic/sparseml/actions/runs/6619351710/job/17979710828?pr=1752#step:8:2243

This does not fail when run standalone;  the recent hunch is some previous test is missing proper teardown; which is affecting this test;

Best solution would be to figure out which test causes this and fixing that; this PR proposes a different solution, where a stronger check is placed on the calibration data loader

The failing test passes with this change

### DIFF
--- a/src/sparseml/modifiers/quantization/pytorch.py
+++ b/src/sparseml/modifiers/quantization/pytorch.py
@@ -14,7 +14,7 @@
 
 import logging
 from itertools import cycle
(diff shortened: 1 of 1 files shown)


rahul-tuli:
Closing cause we figured out the real culprit; and added proper teardown in #1790

PR #1791, opened 2023-10-25
[Feature Branch] Native Text Generation Finetuning

Satrat:
Defines a `SessionManagerMixIn` for HuggingFace's `Trainer` class in order to support sparseML recipes during finetuning.

**Feature Branches**
* Model loading/saving with checkpoint support: https://github.com/neuralmagic/sparseml/pull/1795 
* Additional dataset support (open_platypus and c4): https://github.com/neuralmagic/sparseml/pull/1803
* Doc strings and unit/integration tests: https://github.com/neuralmagic/sparseml/pull/1832
* Distillation: https://github.com/neuralmagic/sparseml/pull/1865

**Note:** Running finetuning with FSDP requires pytorch 2.0.1

### Generating a model for testing
Run SparseGPT to get a 50% pruned tiny llama model for testing
```
python src/sparseml/transformers/sparsification/obcq/obcq.py Xenova/llama2.c-stories15M open_platypus --recipe tiny_recipe.yaml --save True
```

tiny_recipe.yaml:
```
test_stage:
  obcq_modifiers:
    QuantizationModifier:
      ignore:
      - LlamaRotaryEmbedding
      - LlamaRMSNorm
      - SiLUActivation
      - model.layers.0.mlp.down_proj
      - model.layers.1.mlp.down_proj
      - model.layers.2.mlp.down_proj
      - model.layers.3.mlp.down_proj
      - model.layers.4.mlp.down_proj
      - model.layers.5.mlp.down_proj
      post_oneshot_calibration: false
      scheme_overrides:
        Embedding:
          input_activations: null
          weights:
            num_bits: 8
            symmetric: false
    SparseGPTModifier:
      sparsity: 0.5
      block_size: 128
      sequential_update: false
      quantize: true
      percdamp: 0.01
      prunen: 0
      prunem: 0
      targets:
      - model.layers.0
      - model.layers.1
      - model.layers.2
      - model.layers.3
      - model.layers.4
      - model.layers.5
      target_ids:
      - attention_mask
      - position_ids

```

### Running Finetuning Example
Test script for running epoch of finetuning on wikitext, using a constant pruning modifier to maintain sparsity
```
python test_trainer.py
```
with FSDP
```
accelerate launch --config_file integrations/huggingface-transformers/finetuning/example_fsdp_config.yaml test_trainer.py
```


fsdp_config.yaml
```diff
@@ -0,0 +1,24 @@
+compute_environment: LOCAL_MACHINE
```
bfineran, line 1: is this meant to be committed to main? if so we should move to inside of the src dir (and will need to add to the `MANIFEST` file)

src/sparseml/transformers/finetune/helpers.py
```diff
@@ -0,0 +1,152 @@
...
+)
+from sparseml.transformers.utils import SparseAutoModel
+
+
+__all__ = ["apply_recipe_structure_to_model", "reload_model_state"]
+
+_LOGGER = logging.getLogger(__name__)
+
+
+def apply_recipe_structure_to_model(model: Module, recipe_path: str, model_path: str):
```
bfineran, line 36: if this function is generic, we should move it out of the transformers dir

src/sparseml/transformers/finetune/session_mixin.py
```diff
@@ -0,0 +1,461 @@
...
+    def initialize_session(self, epoch: float, checkpoint: Optional[str]):
+        """
+        Initialize the SparseSession from the specified epoch, evaluates the recipe
+        and initialized the modifiers for the training session
+
+        :param epoch: Epoch to initialize session from, usually 0 unless loading
+        from a checkpoint
+        :param checkpoint: Optional checkpoint to initialize from to continue training
+        """
+        orig_state_dict = self.model.state_dict()
```
bfineran, line 117: would be good if we can move the model loading functionality out of a trainer mixin so that it can be easily re-used for stuff like export

src/sparseml/transformers/finetune/text_generation.py
```diff
@@ -0,0 +1,343 @@
...
+    # Summary on each process
+    _LOGGER.warning(
+        f"Process rank: {training_args.local_rank}, device: {training_args.device}, "
+        f"n_gpu: {training_args.n_gpu}, "
+        f"distributed training: {bool(training_args.local_rank != -1)}, "
+        f"16-bits training: {training_args.fp16}"
+    )
+    _LOGGER.info(f"Training/evaluation parameters {training_args}")
+
+    # Detecting last checkpoint.
```
bfineran, line 143: let's move this to a helper fn if we can

src/sparseml/transformers/finetune/text_generation.py
```diff
@@ -0,0 +1,343 @@
...
+    }
+    # this calls from_pretrained under the hood so should be FSDP safe
+    model = SparseAutoModel.text_generation_from_pretrained(
+        model_name_or_path=model_args.model_name_or_path,
+        model_type="model",
+        **model_kwargs,
+    )
+
+    # initialize structure of input model from recipe if needed
+    model_path = model_args.model_name_or_path
```
bfineran, line 194: what is the relationship between this initialization and that in the mixin

src/sparseml/transformers/finetune/text_generation.py
```diff
@@ -0,0 +1,343 @@
...
+        tokenizer_src,
+        cache_dir=model_args.cache_dir,
+        use_fast=True,
+        revision=model_args.model_revision,
+        use_auth_token=True if model_args.use_auth_token else None,
+    )
+
+    # Load datasets
+    # TODO: will any of this cause problems with FSDP?
+    splits = data_args.splits
```
bfineran, line 218: would also be good to move this to a helper fn unless it requires too many loose vars to be passed around

bfineran (commented):
looks great @Satrat, code is very well split and easy to follow

fsdp_config.yaml
Satrat: This is an example config file, users would have to update it based on their # of GPUs and they may also want to change other settings. It probably makes most sense in an example directory with a README explaining how to change and use it, I'll move it

src/sparseml/transformers/finetune/text_generation.py
Satrat: This is the same var passed to the mixin, it gets logged a couple times so I reassigned it for brevity. I'll update the mixin init to all use this var to make it more clear

bfineran (approved):

src/sparseml/pytorch/model_load/helpers.py
```diff
@@ -0,0 +1,173 @@
...
+
+    # reload the state dict for the model now that architecture matches expected
+    if reload_model_state(model, model_path, orig_state_dict):
+        _LOGGER.info(
+            "Reloaded model state after SparseML recipe structure modifications "
+            f"from {model_path}"
+        )
+
+
+def reload_model_state(
```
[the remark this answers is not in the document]
Satrat: ah good catch, removing!

PR #1796, opened 2023-10-27
Only apply DDP after manager is applied and weights are loaded

anmarques:
Before this PR, DDP was applied when model was first created, before recipe is applied. This creates problems when the recipe specifies parameter names, since DDP will wrap the model and alter the parameter names.

**Test plan:**

The following command was used as test case (it fails w/o the PR)

```
torchrun --nproc_per_node 4 --no_python \
    --recipe zoo:cv/classification/resnet_v1-50/pytorch/sparseml/imagenet/pruned90_quant-none \
    --amp \
    --arch-key resnet50 \
    --dataset-path /network/datasets/imagenet \
    --batch-size 256 \
    --output-dir resnet50-imagenet-pruned90_quant-none
```

### DIFF
--- a/src/sparseml/pytorch/torchvision/train.py
+++ b/src/sparseml/pytorch/torchvision/train.py
@@ -401,30 +401,27 @@ def main(args):
 
     _LOGGER.info("Creating model")
     local_rank = int(os.environ["LOCAL_RANK"]) if args.distributed else None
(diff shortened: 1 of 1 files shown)


ohaijen (approved):
Thanks for fixing this!

bfineran (approved):

PR #1798, opened 2023-10-28
Add mistral to obcq

mgoin:
Adds mistral support for the obcq script using the latest [nm-transformers](https://github.com/neuralmagic/transformers). This diff experiments with loading with the simple AutoModel interface and finding the model arch name from the config file's `model_type`

Here is a sample recipe I used for zephyr-beta
```yaml
test_stage:
  obcq_modifiers:
    QuantizationModifier:
      ignore:
        - MistralRotaryEmbedding
        - MistralRMSNorm
        - SiLUActivation
        - model.layers.1.mlp.down_proj
        - model.layers.31.mlp.down_proj
        - model.layers.30.mlp.down_proj
        - model.layers.30.mlp.gate_proj
        - model.layers.30.mlp.up_proj
      post_oneshot_calibration: True
      scheme_overrides:
        Embedding:
          input_activations: null
          weights:
            num_bits: 8
            symmetric: False
    SparseGPTModifier:
      sparsity: 0.5
      block_size: 128
      sequential_update: False
      quantize: True
      percdamp: 0.01
      prunen: 0
      prunem: 0
      targets: [
        "model.layers.0",
        "model.layers.1",
        "model.layers.2",
        "model.layers.3",
        "model.layers.4",
        "model.layers.5",
        "model.layers.6",
        "model.layers.7",
        "model.layers.8",
        "model.layers.9",
        "model.layers.10",
        "model.layers.11",
        "model.layers.12",
        "model.layers.13",
        "model.layers.14",
        "model.layers.15",
        "model.layers.16",
        "model.layers.17",
        "model.layers.18",
        "model.layers.19",
        "model.layers.20",
        "model.layers.21",
        "model.layers.22",
        "model.layers.23",
        "model.layers.24",
        "model.layers.25",
        "model.layers.26",
        "model.layers.27",
        "model.layers.28",
        "model.layers.29",
        "model.layers.30",
        "model.layers.31",
      ]
```

### DIFF
--- a/src/sparseml/transformers/sparsification/obcq/obcq.py
+++ b/src/sparseml/transformers/sparsification/obcq/obcq.py
@@ -36,7 +36,7 @@ __all__ = ["one_shot"]
 
 _LOGGER = logging.getLogger(__name__)
 SUPPORTED_DATASETS = ["wikitext2", "ptb", "c4", "open_platypus"]
-SUPPORTED_MODELS = ["opt", "llama"]
+SUPPORTED_MODELS = ["opt", "llama", "mistral"]
 
 
 def one_shot(
@@ -76,6 +76,9 @@ def one_shot(
         model_loader_fn = SparseCasualLM.opt_model_from_pretrained
         forward_fn = opt_forward
     elif "llama" in model_path.lower():
+        model_loader_fn = SparseCasualLM.llama_model_from_pretrained
+        forward_fn = llama_forward
+    elif "mistral" or "zephyr" in model_path.lower():
         model_loader_fn = SparseCasualLM.llama_model_from_pretrained
         forward_fn = llama_forward
     else:
--- a/src/sparseml/transformers/sparsification/obcq/obcq.py
+++ b/src/sparseml/transformers/sparsification/obcq/obcq.py
@@ -19,6 +19,7 @@ from pathlib import Path
 from typing import Optional
 
 from torch.nn import Module
+from transformers import AutoConfig
 
 import sparseml.core.session as session_manager
 from sparseml.core.framework import Framework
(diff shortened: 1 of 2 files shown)


Satrat (dismissed):

src/sparseml/transformers/sparsification/obcq/obcq.py
```diff
@@ -78,6 +78,9 @@ def one_shot(
     elif "llama" in model_path.lower():
         model_loader_fn = SparseCasualLM.llama_model_from_pretrained
         forward_fn = llama_forward
+    elif "mistral" or "zephyr" in model_path.lower():
```
bfineran, line 81: I don't think this evaluates the way you intend since the portion after the `or` is evaluated separately. as a bool would probably always evaluate to `True`

```python
>>> "mistral" or "zephyr" in "mistral"
'mistral'
```

mgoin: this is a good catch! i'm actually going to change this to use the model_type in the config to be more robust

src/sparseml/transformers/sparsification/obcq/obcq.py
```diff
@@ -78,6 +78,9 @@ def one_shot(
...
+        model_loader_fn = SparseCasualLM.llama_model_from_pretrained
+        forward_fn = llama_forward
```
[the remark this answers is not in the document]
mgoin: I'd actually like to get rid of these custom forward_fns and default to a plain inference with device_map auto to take care of cpu offloading. I'll leave that for a future PR. For now I'll just put mistral in a new block for the automodel path

bfineran (approved):

PR #1800, opened 2023-10-30
[Bug Fix] Reducing Memory Requirements for Sequential OBCQ

rsnm2:
Previously, the OBCQ script would move the entire input model to the GPU during initialization. This PR removes this and instead loads the model layer by layer. 

CAVEAT: This fix is only for the SparseGPTModifier. For the  quantization and smoothquant modifiers, the model is still fully loaded onto the device in order to run calibration. Do we want to add additional options to run calibration on the CPU (would also need to fix the fp16 issues) while keeping OBCQ on the GPU, or update the quantization modifiers themselves to be able to run layer by layer (probably more work)?

### Testing

test_recipe.yaml
```
test_stage:
  obcq_modifiers:
    SparseGPTModifier:
      sparsity: 0.5
      block_size: 128
      sequential_update: True
      quantize: False
      percdamp: 0.01
      mask_structure: "0:0"
      targets: [
        "re:model.layers.\\d*$"
      ]
```

To run:
```
python src/sparseml/transformers/sparsification/obcq/obcq.py /path/to/llama open-platypus --recipe test_recipe.yaml --eval wikitext
```

Memory requirements with change (from nvidia-smi): ~9.6GB
Before Change: >16GB

### DIFF
--- a/src/sparseml/core/state.py
+++ b/src/sparseml/core/state.py
@@ -177,10 +177,6 @@ class State:
 
         if "device" in kwargs:
             self.hardware.device = kwargs["device"]
-            print("\n\n\n------------------------------------------------------------")
-            print("SKIPPING MODEL.to(device)")
-            print("------------------------------------------------------------\n\n\n")
-            # self.model.model.to(self.hardware.device)
 
         if (
             start is not None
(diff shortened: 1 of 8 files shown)


rsnm2:
note  --- if running OBQ, this line is will cause OOM 

https://github.com/neuralmagic/sparseml/blob/24a61a979819862d85023ddb21b71932907d84e0/src/sparseml/modifiers/quantization/pytorch.py#L98

bfineran (approved):

mgoin (approved):

PR #1803, opened 2023-10-31
Additional Datasets for Finetuning

Satrat:
Adding support for C4 and Open Platypus and ability to specify additional kwargs to `datasets.load_dataset`. Also added support for custom dataset splits. For instance, Open Platypus only has a `train` split defined, but now new splits can be created like this:
```
data_args.splits = {
    "train": "train[:90%]",
    "test": "train[90%:]"
}
```

### Testing
Custom splits and additional kwargs for C4 can be tested with. Also tested with Open Platypus
```
python test_trainer.py
```

### TODO
* refactor OBCQ script to use this  same dataset registry (maybe in separate PR)

### DIFF
--- a/src/sparseml/transformers/finetune/data/__init__.py
+++ b/src/sparseml/transformers/finetune/data/__init__.py
@@ -15,6 +15,6 @@
 # flake8: noqa
 
 from .base import TextGenerationDataset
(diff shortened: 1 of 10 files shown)


bfineran (approved):
LGTM but looks like it needs a rebase

PR #1804, opened 2023-10-31
[OBCQ Recipe UX Change] Remove target_ids

rahul-tuli:
This PR removes target_ids param from OBCQ modifier; now this param is not needed and will be inferred automatically

Test plan: Run obcq on both llama, and opt-1.3b; and assert we do not drop in perplexity

Expected perplexity: 
 -  opt - b/w 18-20
-  llama - b/w 7-8

Test commands:

- [X] OPT
```bash
python3 src/sparseml/transformers/sparsification/obcq/obcq.py faceb
ook/opt-1.3b c4 --recipe src/sparseml/transformers/sparsification/obcq/example.yaml --eval wikitext2 
```

Output: (Truncated)
```bash
2023-10-31 10:55:11 sparseml.modifiers.obcq.utils.helpers INFO     Perplexity: 18.037699
```

- [X] LLama-7B

```bash
python src/sparseml/transformers/sparsification/obcq/obcq.py /home/rahul/llama/training open_platypus --recipe src/sparseml/transformers/sparsification/obcq/example_llama.yaml --eval wikitext2
```

Output: (Truncated)
```bash
2023-11-01 09:12:28 sparseml.modifiers.obcq.utils.helpers INFO     Perplexity: 7.168426
```

Changes in existing recipes: Look at example.yaml/example_llama.yaml but TLDR; just remove target_ids field altogether

### DIFF
--- a/src/sparseml/modifiers/obcq/pytorch.py
+++ b/src/sparseml/modifiers/obcq/pytorch.py
@@ -167,7 +167,6 @@ class SparseGPTModifierPyTorch(SparseGPTModifier):
         nsamples: int = None,
         dev: str = "cuda:0",
         layer_prefix: Optional[str] = None,
-        # target_ids: List[str] = None,
     ) -> Dict:
         """
         Runs calibration data through the bottom part of the network (everything up
--- a/src/sparseml/modifiers/obcq/pytorch.py
+++ b/src/sparseml/modifiers/obcq/pytorch.py
@@ -107,7 +107,6 @@ class SparseGPTModifierPyTorch(SparseGPTModifier):
         # decoder layer. Also return attention_mask as part of kwargs
         extras = self.compress_bottom(
             dev=self.device_,
-            target_ids=self.target_ids,
             layer_prefix=self.layer_prefix_,
             **accum_kwargs,
         )
@@ -167,8 +166,8 @@ class SparseGPTModifierPyTorch(SparseGPTModifier):
         dataloader: List = None,
         nsamples: int = None,
         dev: str = "cuda:0",
-        target_ids: List[str] = None,
         layer_prefix: Optional[str] = None,
+        # target_ids: List[str] = None,
     ) -> Dict:
         """
         Runs calibration data through the bottom part of the network (everything up
@@ -183,7 +182,7 @@ class SparseGPTModifierPyTorch(SparseGPTModifier):
         """
         layer_prefix = layer_prefix or self.layer_prefix_
         cached_inputs = cache_attention_inputs(
-            self.model, dataloader, dev, nsamples, target_ids, layer_prefix
+            self.model, dataloader, dev, nsamples, layer_prefix
         )
 
         outputs = cached_inputs.pop("inputs")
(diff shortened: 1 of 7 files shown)


src/sparseml/modifiers/obcq/utils/helpers.py
```diff
@@ -22,22 +23,33 @@
...
+            self.target_keys = self._get_target_keys(kwargs.keys())
+
         for key in self.target_keys:
             self.cache[key].append(kwargs[key])
         raise ValueError
 
     def get_cache(self):
         return self.cache
 
+    def _get_target_keys(self, input_keys):
```
anmarques, line 45: I suggest creating a list of _DEFAULT_TARGET_KEYS = ["attention_mask", "position_ids"] at the top of helpers.py to make this more explicit in case this needs to change in the future.

src/sparseml/modifiers/obcq/utils/helpers.py
```diff
@@ -22,22 +23,33 @@
...
+        target_keys = []
+        if "attention_mask" in input_keys:
+            target_keys.append("attention_mask")
+        if "position_ids" in input_keys:
+            target_keys.append("position_ids")
```
dsikka, line 50: nit: hardcoded constants - use enum or define constants

src/sparseml/modifiers/obcq/utils/helpers.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import logging
+from collections import defaultdict
```
dsikka, line 16: in general would it be possible to add docstrings/typing to this file? just so that it's easier to follow. not for this PR but would be nice to see

Satrat:
LGTM pending the other PR comments, I agree having a list of default ids would be useful

src/sparseml/modifiers/obcq/utils/helpers.py
rahul-tuli: Sure, will add it in a different PR, created a ticket to track request https://app.asana.com/0/1201735099598270/1205939915090873/f

src/sparseml/modifiers/obcq/pytorch.py
```diff
@@ -183,7 +181,7 @@ def compress_bottom(
         """
         layer_prefix = layer_prefix or self.layer_prefix_
         cached_inputs = cache_attention_inputs(
-            self.model, dataloader, dev, nsamples, target_ids, layer_prefix
+            self.model, dataloader, dev, nsamples, layer_prefix
```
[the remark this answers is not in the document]
rahul-tuli: Agree!

src/sparseml/modifiers/obcq/utils/helpers.py
rahul-tuli: Went with the route recommended by @anmarques; and declared `_DEFAULT_TARGET_IDS` at the top

I didn't want to spend too much time on this for the following reasons:
- right now we do not support targeting `LMhead` for pruning (which we should)
- LMHead has slightly different inputs, so the logic here would need to change eventually to support that
- This UX change is just for the meantime

dbogunowicz (approved):

Satrat (approved):

PR #1813, opened 2023-11-03
Add Logarithmic Activation Equalization for one-shot quantization

anmarques:
This PR implements a modifier for the Logarithmic Activation Equalization algorithm, introduced in thie paper [FPTQ: Fine-grained Post-Training Quantization for Large Language Models](https://arxiv.org/abs/2308.15987).

The algorithm is a variant of SmoothQuant, with a different method for computation of the smoothing scales. So this implementation basically inherits from SmoothQuantModifier and modifies the _calculate_smoothing_scales method of the PyTorch version.

The PR also adds a couple of changes to SmoothQuantModifier:
- Fix how the min/max values are computed
- Add default value for the smoothing strength. The parameter is not needed in LAE, so by adding a default value we switch it from a required to an optional parameter from the user point of view. This simplifies the inheritance logic.
- Replaced "SmoothQuantModifier" from logging messages by self.__class__.__name__ so changes the modifier name for inherited classes

### DIFF
--- a/src/sparseml/modifiers/logarithmic_equalization/base.py
+++ b/src/sparseml/modifiers/logarithmic_equalization/base.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
 from sparseml.modifiers.smoothquant import SmoothQuantModifier
 
 __all__ = ["LogarithmicEqualizationModifier"]
@@ -31,12 +30,11 @@ class LogarithmicEqualizationModifier(SmoothQuantModifier):
      small set of calibration data through the model.
 
      This algorithm is very similar to SmoothQuant, changing only how the smoothing scales
-     are computed.
+     are computed. This modifier inherits most functionality from the SmoothQuantModifier.
 
     example recipe:
      ```yaml
      LogarithmicEqualizationModifier:
-       smoothing_strength: 0.5
        mappings: [
          [["re:.*q_proj", "re:.*k_proj", "re:.*v_proj"], "re:.*self_attn_layer_norm"],
          [["re:.*fc1"], "re:.*final_layer_norm"]
@@ -50,6 +48,3 @@ class LogarithmicEqualizationModifier(SmoothQuantModifier):
      :param num_calibration_steps: number of samples to use for calibration, or None to
      use the whole dataset
     """
-
-    smoothing_strength: None # Not used in logarithmic equalization
-
--- a/src/sparseml/modifiers/logarithmic_equalization/base.py
+++ b/src/sparseml/modifiers/logarithmic_equalization/base.py
@@ -1,0 +1,55 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from sparseml.modifiers.smoothquant import SmoothQuantModifier
+
+__all__ = ["LogarithmicEqualizationModifier"]
+
+
+class LogarithmicEqualizationModifier(SmoothQuantModifier):
+    """
+     Implements the Logarithmic Equalization Algorithm from https://arxiv.org/abs/2308.15987.
+     This modifier performs a channel-wise smoothing of outliers in activations, making them
+     easier to quantize by reducing the dynamic range. The smoothing is offset by
+     applying the inverse operation to the next layer of weights, making the weights
+     slightly more difficult to quantize.
+
+     Because this modifier manipulates the weights of the model, it can only be used
+     in one-shot and not during training. Activation ranges are determined by running a
+     small set of calibration data through the model.
+
+     This algorithm is very similar to SmoothQuant, changing only how the smoothing scales
+     are computed.
+
+    example recipe:
+     ```yaml
+     LogarithmicEqualizationModifier:
+       smoothing_strength: 0.5
+       mappings: [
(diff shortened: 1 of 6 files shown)


bfineran (dismissed):

src/sparseml/modifiers/smoothquant/pytorch.py
```diff
@@ -189,4 +190,5 @@ def _calculate_smoothing_scales(
         scales = activation_scales.pow(self.smoothing_strength) / weight_scales.pow(
             1 - self.smoothing_strength
         )
+        scales = torch.where(weight_scales > 0.0, scales, activation_scales)
```
mgoin, line 193: what if weight_scales == 0? why consider the negative case at all if we are using the absolute val

src/sparseml/modifiers/logarithmic_equalization/pytorch.py
```diff
@@ -0,0 +1,52 @@
...
+        range of the activation and the following weights
+
+        :param balance_layers: layers to offset activation smoothing to
+        :param activation_scales: channel-wise dynamic range of activation to smooth
+        :return: channel-wise scales to use for smoothing activation
+        """
+        # calculate the amount of smoothing to apply
+        # s_j = max(|X_j|) / log2( 2 + max(|X_j|) )
+        # where j is the input channel
+        scales = activation_scales / torch.log2(2 + activation_scales)
```
mgoin, line 51: nit: strange for me to include max and abs in the comment but not in the code snippet (assumption being this is handled before calling the function)

anmarques: The max comes from observing statistics over multiple samples. But IMO the formula is a bit more intuitive w/ the max rather than referring to something that is only defined in code.

src/sparseml/modifiers/smoothquant/pytorch.py
mgoin: i read this backwards.. thanks 😅

mgoin (dismissed):
thanks for clarity

src/sparseml/modifiers/logarithmic_equalization/base.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+     Implements the Logarithmic Equalization Algorithm from
+     https://arxiv.org/abs/2308.15987.
+     This modifier performs a channel-wise smoothing of outliers in activations,
+     making them easier to quantize by reducing the dynamic range. The smoothing is
+     offset by applying the inverse operation to the next layer of weights, making
+     the weights slightly more difficult to quantize.
+
+     Because this modifier manipulates the weights of the model, it can only be
+     used in one-shot and not during training. Activation ranges are determined by
```
ohaijen, line 31: It's not clear to me why this modifier cannot be used during training. Since its application is mathematically neutral, it seems like it can also be applied between training steps (though it may require adding modules to the network, and also  may not be particularly helpful in any way).

Can probably resolve this by changing "can only be used" -> "should only be used".

src/sparseml/modifiers/logarithmic_equalization/base.py
```diff
@@ -0,0 +1,53 @@
...
+     LogarithmicEqualizationModifier:
+       mappings: [
+         [["re:.*q_proj", "re:.*k_proj", "re:.*v_proj"], "re:.*self_attn_layer_norm"],
+         [["re:.*fc1"], "re:.*final_layer_norm"]
+       ]
+       ignore: ["model.decoder.final_layer_norm"]
+     ```
+
+     :param mappings: list activation layers to smooth, and which layers to offset
+     the smoothing to for each activation
```
ohaijen, line 49: I found the regex approach confusing here. I guess it's smart enough to know that if it matches "re:.*k_proj", it should offset the layer matching "re:.*self_attn_layer_norm" in the corresponding resblock? If so, can this be clarified in the comment?

Also, regarding the "ignore" parameter just below. Does it match on the first value of a mapping, the second value, or both?

src/sparseml/modifiers/logarithmic_equalization/pytorch.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def _calculate_smoothing_scales(
+        self, balance_layers: List[Module], activation_scales: torch.Tensor
+    ) -> List[float]:
+        """
+        Calculate how much smoothing to apply to each channel based on the dynamic
+        range of the activation and the following weights
+
+        :param balance_layers: layers to offset activation smoothing to
+        :param activation_scales: channel-wise dynamic range of activation to smooth
```
ohaijen, line 45: nits: activation -> activations; "range of activation"->"ranges of activations".

src/sparseml/modifiers/smoothquant/base.py
```diff
@@ -111,13 +111,13 @@ def on_initialize(self, state: State, **kwargs) -> bool:
         """
         if self.end and self.end != -1:
             raise ValueError(
-                "SmoothQuantModifier can only be applied during one-shot. Expected end"
-                " to be None or -1, got {}".format(self.end)
+                f"{self.__class__.__name__} can only be applied during one-shot. "
+                f" Expected end to be None or -1, got {self.end}"
             )
         if self.start and self.start != -1:
```
ohaijen, line 117: This condition evaluates to False if self.start == 0, which is probably not what you want.

ohaijen (changes requested):
Hi Alex, mostly some nits. One small error on line 117 of src/sparseml/modifiers/smoothquant/base.py that you'll probably want to fix.

src/sparseml/modifiers/logarithmic_equalization/base.py
anmarques: Fixed

src/sparseml/modifiers/logarithmic_equalization/base.py
anmarques: Added more explanation on how these should be used

src/sparseml/modifiers/logarithmic_equalization/pytorch.py
anmarques: Fixed

src/sparseml/modifiers/smoothquant/base.py
anmarques: This guarantees it only applies as one-shot

ohaijen (approved):
Thanks for making the changes!

Satrat (approved):

PR #1814, opened 2023-11-03
Remove copies of the model for `sparseml.transformers.export_onnx`

mgoin:
Not ready for prime time, but it does work in making LLM export much more memory efficient and faster


bfineran (commented):
I think this could be "safe" to land in our current flows, however since the `self._module` is technically kept around, it might be better to add a decorator or something to the export functions that sets the device to cpu and mode to eval before entering a function then restoring the previous state when exiting

jeanniefinks:

PR #1821, opened 2023-11-10
Adding evolcodealpaca dataset

abhinavnmagic:
The implementation was tested by running quantization on mistral-7b model.


bfineran (commented):
looks like it needs quality fixes - good after

mgoin (approved):

PR #1826, opened 2023-11-13
Change Transformers export default sequence length to max_position_embeddings

mgoin:
## Usage

```
sparseml.transformers.export_onnx --model_path ./TinyLlama-1.1B-Chat-v0.3 --task text-generation --trust_remote_code 
2023-11-13 18:56:23 sparseml.transformers.export INFO     Attempting onnx export for model at /home/mgoin/models/TinyLlama-1.1B-Chat-v0.3 for task text-generation
2023-11-13 18:56:23 sparseml.transformers.export INFO     Using default sequence length of 2048
Special tokens have been added in the vocabulary, make sure the associated word embeddings are fine-tuned or trained.
```

## Reason

Unfortunately I think it is very important to export LLMs with the largest sequence length possible.

For instance I was exporting llama models with a sequence length of 512 for convenience
`sparseml.transformers.export_onnx --model_path ./llama2.c-stories15M --task text-generation --sequence_length 512 `
However due to the PyTorch export and constant folding it seems to produce a cached, fixed rotary embedding of the sequence length used. See it as 512 in this screenshot. The ONNX can still run and compile with a sequence length larger than 512, but the output is unstable and quickly starts to repeat

Screenshot 2023-11-08 at 7 09 36 PM

### DIFF
--- a/src/sparseml/transformers/export.py
+++ b/src/sparseml/transformers/export.py
@@ -297,7 +297,9 @@ def export_transformer_to_onnx(
     )
 
     if sequence_length is None:
-        _LOGGER.info(f"Using default sequence length of {config.max_position_embeddings}")
+        _LOGGER.info(
+            f"Using default sequence length of {config.max_position_embeddings}"
+        )
         sequence_length = config.max_position_embeddings
 
     tokenizer = AutoTokenizer.from_pretrained(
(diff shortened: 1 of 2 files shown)


bfineran (approved):
LGTM - might want to have a fallback default in case max_position_embeddings is never set, but it may just be the case that HF has a default already

rahul-tuli (approved):

mgoin:
> max_position_embeddings

The config class itself has a default value for this, so I'm inclined to leave it be. We could raise ValueError but that will be reflected downstream

PR #1827, opened 2023-11-14
[Bug Fix] Add Precision Argument for Disabling bf16

Satrat:
Asana Ticket: https://app.asana.com/0/1205229323407165/1205846437135486/f

The OBCQ script would load models as "auto", but for quantization we cannot support bfloat16 models. This PR adds a new argument to the script for casting to half or full precision. As expected casting to full precision really slows things down, quantization takes ~2x longer

### Testing 
Without precision cast(use "auto" as before) on a bfloat16 model:
```
src/sparseml/transformers/sparsification/obcq/obcq.py selfrag/selfrag_llama2_7b open_platypus --recipe src/sparseml/transformers/sparsification/obcq/example_llama.yaml --eval wikitext2 --precision auto
```
Fails at quantization with `fake_quantize_tensor_cachemask_kernel_types" not implemented for 'BFloat16'`

With precision cast, we can use torch.float16 or torch.float32:
```
src/sparseml/transformers/sparsification/obcq/obcq.py selfrag/selfrag_llama2_7b open_platypus --recipe src/sparseml/transformers/sparsification/obcq/example_llama.yaml --eval wikitext2 --precision half
```
And quantization/OBCQ will run as expected

### DIFF
--- a/src/sparseml/core/state.py
+++ b/src/sparseml/core/state.py
@@ -177,7 +177,6 @@ class State:
 
         if "device" in kwargs:
             self.hardware.device = kwargs["device"]
-            self.model.model.to(self.hardware.device)
 
         if (
             start is not None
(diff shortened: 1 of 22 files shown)


bfineran (dismissed):
LGTM. For export we'll need to make sure weights are exported as 32 I believe

rahul-tuli (approved):

PR #1830, opened 2023-11-14
Pass Device to SmoothQuant

Satrat:
Fix for a bug that caused smoothquant the run on the CPU, the device was not being passed to the calibration function properly


bfineran (approved):

mgoin (approved):
this fixed my issue, thanks!

PR #1833, opened 2023-11-15
[BugFix] Call `model.deployment_directory_path` explicitly in failing test

rahul-tuli:


dbogunowicz (approved):
That's it? Only one test breaks as a result of this change? Phew!
Before we land this, let's discuss this in the standup ok?

rahul-tuli:
> That's it? Only one test breaks as a result of this change? Phew!
> Before we land this, let's discuss this in the standup ok?

2 tests because of the nice parameterization, but yes!

Let's discuss more during stand up

bfineran (approved):

PR #1838, opened 2023-11-16
Upgrade PyTorch to 2.1

Satrat:
Upgrading PyTorch to 2.1, based on discussion here: https://app.asana.com/0/1205229323407165/1205806737893107/f 

This required upgrading torchvision to 0.16.0 to maintain compatibility. Tested one_shot and finetuning scripts run as expected with the upgrade, and that all pytorch tests pass locally


setup.py
```diff
@@ -63,17 +63,17 @@
 
 _onnxruntime_deps = ["onnxruntime>=1.0.0"]
 _clip_deps = ["open_clip_torch==2.20.0"]
-supported_torch_version = "torch>=1.7.0,<=2.0"
+supported_torch_version = "torch>=1.7.0,<=2.1"
```
mgoin, line 66: could we do < rather than <= so we could pick up hotfixes? it should work even if 2.2 doesn't exist

setup.py
```diff
@@ -63,17 +63,17 @@
 
 _onnxruntime_deps = ["onnxruntime>=1.0.0"]
 _clip_deps = ["open_clip_torch==2.20.0"]
-supported_torch_version = "torch>=1.7.0,<=2.0"
+supported_torch_version = "torch>=1.7.0,<2.1"
```
mgoin, line 66:
```suggestion
supported_torch_version = "torch>=1.7.0,<2.2"
```

setup.py
```diff
@@ -63,17 +63,17 @@
...
 _pytorch_deps = [
     supported_torch_version,
     "gputils",
 ]
 _pytorch_all_deps = _pytorch_deps + [
-    "torchvision>=0.3.0,<=0.15.1",
+    "torchvision>=0.3.0,<0.16",
```
mgoin, line 72:
```suggestion
    "torchvision>=0.3.0,<0.17",
```

setup.py
```diff
@@ -63,17 +63,17 @@
...
     "torchaudio<=2.0.1",
 ]
 _pytorch_vision_deps = _pytorch_deps + [
-    "torchvision>=0.3.0,<=0.15.1",
+    "torchvision>=0.3.0,<0.16",
```
mgoin, line 76:
```suggestion
    "torchvision>=0.3.0,<0.17",
```

mgoin (commented):
hey @Satrat here is what i meant by my proposal for `<` on a higher version (that doesn't have to exist yet but so we can use 2.1.x as they come out)

Satrat:
> hey @Satrat here is what i meant by my proposal for `<` on a higher version that doesn't have to exist yet, but so we can use 2.1.x as they come out

Ah my bad, fixed now!

bfineran (dismissed):

mgoin (approved):

PR #1858, opened 2023-11-28
[Export Refactor] Feature Branch

dbogunowicz:
## Feature Description

Feature branch that implements new exporting framework. This PR roughly outlines the new design as outlined in the diagram below:
image
## Features

The first order of business is to migrate Image Classification export to use the new framework. The related PRs are:
- [x] #1878 
- [x] #1880
- [x] #1883 
- [x] #1884 
- [x] #1888 
- [x] #1889 
- [x] #1890
- [x] Test the new Image Classification export end to end (#1898)

The second work item is to implement Transformers export:
- [x] #1908 
- [x] #1909 
- [x] #1910 

After the aforementioned smaller PRs land, the next order of affairs would be to:
- [ ] Add some basic documentation
- [ ] Remove the part of the codebase responsible for legacy export and swap the outward-facing pathways to use the new export
- [x] Fix any pending tests

Also add few missing, potentially slightly orthogonal features:
- [x] Make sure that all `create_model` functions properly create sparse models (use @Satrat 's new modifiers), also apply one shot etc.
- [ ] Validate `validate_correctness` function (some sparsezoo edits are required)
- [ ] Properly implement and validate separating the exported onnx model into multiple files
- [x] Use `device` argument when creating a model
- [ ] (Potentially) set `is_tar` as an export argument (to decide later)

**CLI Example**
```bash
sparseml.export PATH/TO/SPARSIFIED/LLM \
  --task text-generation
```


src/sparseml/export.py
```diff
@@ -0,0 +1,185 @@
...
+
+from sparseml.pytorch.opset import TORCH_DEFAULT_ONNX_OPSET
+from sparsezoo.utils.registry import RegistryMixin
+
+
+_LOGGER = logging.getLogger(__name__)
+AVAILABLE_DEPLOYMENT_TARGETS = ["deepsparse", "onnxruntime"]
+
+
+class SparseMLIntegrationRegistry(RegistryMixin):
```
bfineran, line 29: note - we can use mixin directly on the registered class if we want to (as an additional parent class)

src/sparseml/transformers/sparsification/obcq/obcq.py
```diff
@@ -90,25 +88,29 @@ def one_shot(
...
-        model_path, sequence_length=sequence_length, torch_dtype=torch_dtype
+    # create session and initialize a sparse model
+    session_manager.create_session()
+    model = initialize_sparse_model(
+        model_path=model_path,
+        task="text-generation",
+        sequence_length=sequence_length,
+        torch_dtype=torch_dtype,
+        config=config,
+        recipe=recipe_file,
```
Satrat, line 112: I think this is the reason for the failing tests. We're initializing the model with the recipe we want to apply to it, but it should be initialized with `input_recipe_path` instead, defined on deleted lines 130-137. `recipe_file` shouldn't get applied to the model until we call the `apply` function

Satrat (changes requested):

src/sparseml/export/export.py
```diff
@@ -46,8 +46,9 @@ def export(
     single_graph_file: bool = True,
     num_export_samples: int = 0,
     batch_size: int = 1,
+    recipe: Optional[Union[Path, str]] = None,
     deployment_directory_name: str = "deployment",
-    device: str = "auto",
+    device: str = "cpu",
```
Satrat: Why do we change this default from auto to cpu? on line 118 it looks like we're expecting auto as a default

src/sparseml/transformers/utils/initializers.py
```diff
@@ -206,20 +206,53 @@ def resolve_recipe_application(
...
+        return _resolve_recipe_file(recipe, model_path)
+    elif isinstance(recipe, str):
+        # recipe is a string containing the recipe
+        _LOGGER.debug(
+            "Applying the recipe string directly to the model, without "
+            "checking for a potential existing recipe in the model_path."
         )
+        return recipe
+
+    _LOGGER.info(
```
Satrat, line 245: Could we add another log statement on success that prints out the path to the recipe we are loading?

Satrat (commented):
Latest commits LGTM, just had some small comments. Lets also be sure to test this on llama 7b before landing to make sure we're getting the same perplexity results as before

Satrat (approved):

Satrat (dismissed):
LGTM, just have one unresolved comment about adding an additional log message. Also looks like we need to resolve a merge conflict with the GHA file

bfineran (approved):

PR #1864, opened 2023-11-29
Fix flaky Top-KAST test by imposing numerical stability (smaller lr, fewer epochs of training)

ohaijen:
Fix flaky Top-Kast test by imposing numerical stability.

### DIFF
--- a/tests/sparseml/pytorch/sparsification/pruning/test_modifier_pruning_topkast.py
+++ b/tests/sparseml/pytorch/sparsification/pruning/test_modifier_pruning_topkast.py
@@ -179,7 +179,7 @@ class TestTopKASTPruningModifier(ScheduledModifierTest):
                 assert torch.allclose(
                     modifier._module_masks._params[i][backward_mask],
                     layer_weights_pre._params[i][backward_mask]
-                    * (1 - 0.0002 * 0.25 * 1 / modifier._forward_sparsity),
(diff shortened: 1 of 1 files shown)


tests/sparseml/pytorch/sparsification/pruning/test_modifier_pruning_topkast.py
```diff
@@ -125,7 +125,6 @@ def _test_compression_sparsity_applied():
             assert not modifier.update_ready(epoch, test_steps_per_epoch)
             _test_compression_sparsity_applied()
 
-    @pytest.mark.skip(reason="This test is flaky; fix is in progress")
     @flaky(max_runs=3, min_passes=2)
```
dbogunowicz, line 128: Nit: now when the tests are potentially not flaky anymore, consider making the "flakiness" more lenient, potentially removing the `@flaky` completely

dbogunowicz (approved):
Approved, pending GHA testing results.

tests/sparseml/pytorch/sparsification/pruning/test_modifier_pruning_topkast.py
ohaijen: Hi Damian, I considered this. But I left it in, because it seems pretty harmless and the test still relies on randomness (in the network init), so it actually could rarely fail.

anmarques (approved):

PR #1874, opened 2023-12-04
Allow buffers to set modifier's device

abhinavnmagic:
While applying modifiers, we utilize the module's device to set the correct device for the additional modules/buffers like fake quantization modules.
Presently, we default to cpu in case the module doesn't have any parameters. This PR adds the capability to utilize the buffer's device in case there are no parameters in the module. The change is based on PyTorch's latest implementation from which the `get_device` function is adapted from.

### DIFF
--- a/src/sparseml/modifiers/quantization/utils/quantize.py
+++ b/src/sparseml/modifiers/quantization/utils/quantize.py
@@ -234,11 +234,12 @@ def add_output_activation_observers(module: Module):
 
     def get_device(m):
         # infers device based on either paramteres or buffers
-        unique_devices = {p.device for p in m.parameters()} | \
-                            {p.device for p in m.buffers()}
+        unique_devices = {p.device for p in m.parameters()} | {
+            p.device for p in m.buffers()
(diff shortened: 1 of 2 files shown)


src/sparseml/modifiers/quantization/utils/quantize.py
```diff
@@ -231,11 +231,15 @@ def add_output_activation_observers(module: Module):
...
-        # default to CPU if module has no parameters
-        device = "cpu"
+
+    def get_device(m):
+        # infers device based on either paramteres or buffers
+        unique_devices = {p.device for p in m.parameters()} | {
+            p.device for p in m.buffers()
+        }
+
+        assert len(unique_devices) <= 1
```
bfineran, line 241: need to raise a specific error here - generally we try to avoid asserts that users could hit

abhinavnmagic: Sure, will update the PR.

jeanniefinks:

PR #1878, opened 2023-12-06
[Export Refactor][Image Classification] `create_model` function

dbogunowicz:


src/sparseml/pytorch/image_classification/utils/helpers.py
```diff
@@ -344,17 +345,24 @@ def get_dataset_and_dataloader(
 # Model creation Helpers
 def create_model(
```
dbogunowicz, line 346: I am folding alternative pathway for inferencing `num_classes` if the user does not wish to provide it themselves. `num_classes` needs to be specified to create a model.

dbogunowicz: (see: `src/sparseml/pytorch/image_classification/export.py`)

src/sparseml/pytorch/image_classification/utils/helpers.py
```diff
@@ -643,3 +684,17 @@ def _download_model_from_zoo_using_recipe(
         Model(recipe_stub),
         recipe_name=recipe_type,
     )
+
+
+def is_image_classification_model(source_path: Union[Path, str]) -> bool:
```
dbogunowicz, line 689: Note: Empirically I have discover that not many models in sparsezoo have `arch_key` baked into their checkpoint

src/sparseml/export/registry.py
```diff
@@ -0,0 +1,113 @@
...
+# TODO: Fold it into the functionalities of the RegistryMixin
+# when the `resolve` method is generically implemented
+def infer_integration(source_path: Union[Path, str]) -> str:
+    """
+    Infer the integration to use for exporting the model from the source_path.
+
+    :param source_path: The path to the PyTorch model to export.
+    :return: The name of the integration to use for exporting the model.
+    """
+    if is_image_classification_model(source_path):
```
bfineran, line 50: probably want to add hot import here as well to make sure that the IC helper functions actually gets registered

src/sparseml/export/registry.py
```diff
@@ -0,0 +1,113 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
```
bfineran, line 1: would rename this file to `integration_helper_functions`

bfineran: also maybe wouldn't keep under `export/` since ideally these helpers will be used in other places such as train/torch eval flows

src/sparseml/export/registry.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+class IntegrationHelperFunctions(RegistryMixin, BaseModel):
+    """
+    Registry that maps names to helper functions
+    for creation/export/manipulation of models for a specific
+    integration.
+    """
+
+    create_model: Optional[Callable] = Field(
```
bfineran, line 67: we need defaults here eventually

src/sparseml/export/registry.py
```diff
@@ -0,0 +1,113 @@
...
+        if v is not None:
+            v = cls.wrap_to_return_first_output(v)
+        return v
+
+    @staticmethod
+    def wrap_to_return_first_output(func: Callable) -> Callable:
+        return lambda *args, **kwargs: func(*args, **kwargs)[0]
+
+
+@IntegrationHelperFunctions.register(name=Integrations.image_classification.value)
```
bfineran, line 111: this should go in its own file

src/sparseml/export/registry.py
bfineran, line 67: For every function - we MUST have an expected signature (even if we're using kwargs to make a bit generic)

i.e. we define the standard that `create_model` should look something like `(model_path, **kwargs) -> Model`, `create_dummy_input(**kwargs) -> Union[List[Tensor], Tensor]`

bfineran (changes requested):

src/sparseml/export/registry.py
bfineran: we could also do `(model_path, **kwargs) -> Union[Model, Optioanl[kwargs]` so we could return an arch key, the thing thing we can't do is say arch key is always output 2

src/sparseml/pytorch/image_classification/utils/helpers.py
```diff
@@ -643,3 +684,17 @@ def _download_model_from_zoo_using_recipe(
...
+    """
+    :param source_path: The path to the model
+    :return: Whether the model is an image classification model or not
+    """
+    try:
+        checkpoint = torch.load(os.path.join(source_path, "model.pth"))
```
bfineran, line 695: this expects directory, we should also allow a file for IC

src/sparseml/pytorch/image_classification/integration_helper_functions.py
```diff
@@ -0,0 +1,43 @@
...
+from src.sparseml.integration_helper_functions import (
+    IntegrationHelperFunctions,
+    Integrations,
+)
+from src.sparseml.pytorch.image_classification.utils.helpers import (
+    create_model as create_image_classification_model,
+)
+
+
+def create_model(source_path: Union[Path, str], **kwargs) -> torch.nn.Module:
```
bfineran, line 30: what's the plan for arch key?

src/sparseml/integration_helper_functions.py
```diff
@@ -0,0 +1,92 @@
...
+    export_sample_inputs_outputs: Optional[Callable] = Field(
+        description="A function that exports input/output samples given "
+        "a (sparse) PyTorch model."
+    )
+    create_deployment_folder: Optional[Callable] = Field(
+        description="A function that creates a "
+        "deployment folder for the exporter ONNX model"
+        "with the appropriate structure."
+    )
+
```
bfineran, line 69: we should have validation here as well

bfineran (approved):

src/sparseml/pytorch/image_classification/integration_helper_functions.py
dbogunowicz: What do you mean? To create an IC model, the `arch_key` is an optional argument. Potentially `arch_key` can be supplied as a kwarg, but it is not necessary. The `create_model` function calls `ModelRegistry.create(...)` that does not demand receiving `arch_key`, by default it tries to read it out from the model checkpoint .

src/sparseml/integration_helper_functions.py
dbogunowicz: Let's have this discussion in the PR that pertains to this functionality ok?

PR #1880, opened 2023-12-06
[Export Refactor][Image Classification] `create_dummy_input` function

dbogunowicz:


src/sparseml/pytorch/image_classification/integration_helper_functions.py
```diff
@@ -27,17 +27,46 @@
...
+
+    :param validation_loader: The validation loader to get a batch from.
+        If None, a fake batch will be created
+    :param image_size: The image size to use for the dummy input. Defaults to 224
+    :return: The dummy input as a torch tensor
+    """
+
+    if not validation_loader:
+        # create fake data for export
+        val_loader = [[torch.randn(1, 3, image_size, image_size)]]
```
dbogunowicz, line 63:
```suggestion
        validation_loader = [[torch.randn(1, 3, image_size, image_size)]]
```

src/sparseml/export/export.py
```diff
@@ -98,9 +98,12 @@ def export(
         IntegrationHelperFunctions.load_from_registry(integration)
     )
 
-    model = helper_functions.create_model(source_path, device)
+    # for now, this code is not runnable, serves as a blueprint
+    model, auxiliary_items = helper_functions.create_model(
+        source_path, **kwargs  # noqa: F821
+    )  # type: ignore
```
dbogunowicz, line 104:
```suggestion
    ) 
```

src/sparseml/pytorch/image_classification/integration_helper_functions.py
```diff
@@ -27,17 +27,46 @@
...
-    model, *_ = create_image_classification_model(checkpoint_path=source_path, **kwargs)
-    return model
+    model, *_, validation_loader = create_image_classification_model(
+        checkpoint_path=source_path, **kwargs
+    )
+    return model, dict(validation_loader=validation_loader)
+
+
+def create_dummy_input(
+    validation_loader: Optional[torch.utils.data.DataLoader] = None,
```
bfineran, line 49: how do we expect this optional value to flow from the create model function? thinking we need to accept `**kwargs` in this signature to avoid edge cases for extra kwargs (ie arch key that we'll need)

bfineran (approved):

src/sparseml/pytorch/image_classification/integration_helper_functions.py
dbogunowicz:
```python

# def [helper_functions].create_model(...) -> Tuple[torch.nn.Module, Dict[str, Any]]

model: torch.nn.Module, auxiliary_items: Optional[Dict[str, Any]] = helper_functions.create_model(...)

# in case of IC, auxiliary_items is either dict(validation_loader=validation_loader) or dict(validation_loader=None), so this should be picked up by `create_dummy_input`
```

Maybe i'd call the variable more generically, i.e. `data_loader` instead of `validation_loader`

PR #1881, opened 2023-12-06
Misc Finetuning Checkpointing Fixes

Satrat:
### Summary of Changes
This PR contains a few bug fixes for finetuning, mostly related to checkpoints
* Adding logic to `kd_wrapper.py` to determine if `named_modules` should return the full module or just the student. This varys based on when it is called (see comments)
* Expand the `SessionMixIn.save_model` function to export the resulting model as a single bin, this allows it to be reloaded outside of FSDP. 
* Cleaning up checkpoint logic to use the input recipe if we're resuming from a checkpoint
* Only update the loss if a modifier returns a non-None loss value (the previous distillation merge had broken non-distillation loss calculation)

### Testing
Tested saving and reloading checkpoints works with and without FSDP

### Remaining Issues
Resuming an FSDP checkpoint outside of FSDP still doesn't work out of the box. You need to treat the checkpoint directory as a new input directory rather than a checkpoint (which throws off the epoch start)

### DIFF
--- a/src/sparseml/modifiers/distillation/output/pytorch.py
+++ b/src/sparseml/modifiers/distillation/output/pytorch.py
@@ -16,7 +16,7 @@ from typing import Any, Dict
 
 import torch
 from torch.nn import Module
-
+from torch.distributed.fsdp import FullyShardedDataParallel
 from sparseml.core import Event, EventType, State
 from sparseml.modifiers.distillation.output.base import OutputDistillationModifier
 from sparseml.modifiers.distillation.utils.pytorch import KDFactory, KDModuleWrapper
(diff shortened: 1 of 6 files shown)


src/sparseml/modifiers/distillation/output/pytorch.py
```diff
@@ -37,6 +38,8 @@ def on_initialize(self, state: State, **kwargs) -> bool:
             return False
 
         self.wrappers_ = {}
+        if kwargs.get("fsdp_active"):
+            self.fsdp_active_ = True
```
bfineran, line 42: what do we want to convey with the trailing `_` in the variable names? if private our repos generally standardize to putting the `_` first.

Satrat: Its meant to convey a private variable. Since the `Module` class inherits from pydantics `BaseModel`, any class variables led by an `_` are readonly. Not sure if there is a better way to get around this? I'm using the trailing _ in a lot of the modifiers to signify "private" variables

PR #1883, opened 2023-12-07
[Export Refactor][Image Classification] `export_model` function

dbogunowicz:


bfineran (changes requested):
why do we need a special export function for IC? We should be able to use the default, no? If not, we should rethink the flow - something like IC shouldn't require every single function to be defined IMO

dbogunowicz:
@bfineran no worries. I have placed this function under `image_classification`, but you are right. It is general enough to make it a default, general export function for most integrations (excluding the very specific ones like e.g. `transformers`)

PR #1884, opened 2023-12-07
[Export Refactor][Image Classification] `apply_optimizations` function

dbogunowicz:


tests/sparseml/pytorch/image_classification/test_integration_helper_functions.py
```diff
@@ -27,3 +27,4 @@ def test_integration_helper_functions():
     assert image_classification.create_model
     assert image_classification.create_dummy_input
     assert image_classification.export
+    assert image_classification.graph_optimizations is None
```
dbogunowicz, line 30: @bfineran is that right? For image classification, we do not have any graph optimizations available?

src/sparseml/export/helpers.py
```diff
@@ -0,0 +1,107 @@
...
+
+import onnx
+
+from sparsezoo.utils.onnx import save_onnx
+
+
+__all__ = ["apply_optimizations"]
+
+
+class GraphOptimizationOptions(Enum):
```
bfineran, line 28: what's the idea behind this? why an enum instead of bool for now? if we think specific integrations will have different options, then this should likely be handled by kwargs

dbogunowicz: Let's talk about this. When designing this functionality I was planning ahead for more complex use cases, such as the `transformer` export. Specifically for the case, where we allow the user to specify a set of optimizations - thin. This particular Enum is here, to implement the UI consistent with the requirements from Mark.

PR #1888, opened 2023-12-08
[Export Refactor][Image Classification] `export_sample_inputs_outputs` function

dbogunowicz:


src/sparseml/integration_helper_functions.py
```diff
@@ -32,6 +33,27 @@ class Integrations(Enum):
     image_classification = "image-classification"
 
 
+def create_sample_inputs_outputs(
```
bfineran, line 36: we need the model here to actually create the outputs. right now I think you have it confused for labels

bfineran (changes requested):

src/sparseml/integration_helper_functions.py
dbogunowicz: Oh, you are right.

bfineran (approved):

PR #1889, opened 2023-12-08
[Export Refactor][Image Classification] `create_deployment_folder` function

dbogunowicz:


src/sparseml/export/helpers.py
```diff
@@ -11,29 +11,104 @@
...
 
-class GraphOptimizationOptions(Enum):
+
+def create_deployment_folder(
+    source_path: Union[Path, str],
+    target_path: Union[Path, str],
+    deployment_directory_name: str = "deployment",
+    deployment_directory_structure: Optional[Dict] = None,
+    onnx_model_name: Optional[str] = None,
+) -> str:
```
bfineran, line 49: should allow for a list of files to be copied in (will make it easier for other integrations to just call in with their list of configs, tokenizers, etc)

dbogunowicz: I use here dictionary instead of list, but this may be an overkill. Let's discuss.

PR #1890, opened 2023-12-08
[Export Refactor][Image Classification] `validate_correctness` function

dbogunowicz:


bfineran (approved):
LGTM pending proof of passing test - however as mentioned in other PR, we should allow for custom validation fns as well (at least for deployment structure)

PR #1891, opened 2023-12-11
[One Shot] Default to AutoModel loading in OBCQ script

Satrat:
Previously the OBCQ script would return with an error if one of the supported/tested model types wasn't found in the name of the `--model_path` argument. 

With this PR, instead of erroring out we just return a warning to the user and attempt to load the model with `AutoModelForCausalLM` anway.

### Testing
```
src/sparseml/transformers/sparsification/obcq/obcq.py model_path=sahil2801/replit-code-instruct-glaive open_platypus --recipe tests/sparseml/transformers/obcq/test_tiny.yaml --save True
```

The model successfully loads with a warning. One-Shot does fail for this model, will address that in another PR since its MPT-specific

Asana ticket: https://app.asana.com/0/1205229323407165/1206141086244198/f

### DIFF
--- a/src/sparseml/transformers/sparsification/obcq/obcq.py
+++ b/src/sparseml/transformers/sparsification/obcq/obcq.py
@@ -94,7 +94,13 @@ def one_shot(
         model_loader_fn = SparseCausalLM.auto_model_from_pretrained
         forward_fn = llama_forward
     else:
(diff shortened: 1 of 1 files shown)


bfineran (approved):

rahul-tuli (approved):
Nice!

PR #1897, opened 2023-12-12
Support for Stacking Recipes

Satrat:
First PR towards alternating finetuning/one-shot support. Covers recipe and staging changes necessary to enable running recipes stage by stage between training and one-shot. Previously there was no way to differentiate between stages that had been previously applied and the active stage.

### Summary of Changes
* Adding an `applied` attribute to `StageModifiers`, used to differentiate which stages are initialized for structure only (already applied stages). If a stage has already been applied, we skip skip it during initialization and updates
* Keep track of `applied_stages` in the `RecipeContainer`, so that when we reload the recipe after a stage is added we remember which have already been run 
* Adding helper functions to the `QuantizationModifier` for determining if a module is already quantized. This allows us to stack quantization modifiers as long as they don't affect a common module. If attempting to quantize an already quantized module, we raise an exception
*  Update recipe serialization code to add index's to stages with the same name, so they aren't overwritten during recipe export

### Testing
Added new unit tests to `tests/sparseml/transformers/obcq/test_repeats.py`

* Test one-shot SparseGPT recipes are able to be stacked, even if their stage names clash
* Test we error out on clashing quantization modifiers
* Test we allow multiple quantization modifiers as long as they don't overlap


bfineran (approved):

rahul-tuli (approved):
LGTM!

PR #1902, opened 2023-12-13
[One Shot] Specify recipe_args on CLI 

Satrat:
Requested by research, adding a CLI parameter for passing recipe args

### Example Usage

test_recipe.yaml
```yaml
test_stage:
  obcq_modifiers:
    SparseGPTModifier:
      sparsity: eval(use_sparsity)
      block_size: 128
      sequential_update: False
      quantize: False
      targets: [
        "re:model.layers.\\d+$"
      ]
```

To run:
```bash
python src/sparseml/transformers/sparsification/obcq/obcq.py Xenova/llama2.c-stories15M open_platypus --recipe_args use_sparsity=0.7 --recipe test_recipe.yaml
```

### DIFF
--- a/src/sparseml/transformers/sparsification/obcq/obcq.py
+++ b/src/sparseml/transformers/sparsification/obcq/obcq.py
@@ -16,7 +16,7 @@ import argparse
 import logging
 import os
 from pathlib import Path
-from typing import Optional
+from typing import Dict, Optional
 
 import torch
 from torch.nn import Module
@@ -54,6 +54,7 @@ def one_shot(
     deploy_dir: Optional[str] = ".",
     recipe_file: Optional[str] = None,
     precision: str = "auto",
+    recipe_args: Optional[Dict] = None,
     eval_data: Optional[str] = None,
     do_save: Optional[bool] = False,
 ) -> Module:
@@ -68,6 +69,7 @@ def one_shot(
     :param deploy_dir: The output directory to save the model to
     :param recipe_file: recipe containing SparseGPT configuration
     :param precision: precision to load model as, either auto, half or full
+    :param recipe_args: additional arguments to use for recipe evaluation
     :param eval_data: dataset to use for perplexity evalaution, or none to skip
     :param do_save: whether to save the output model to disk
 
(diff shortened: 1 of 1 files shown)


bfineran (approved):

natuan (changes requested):
@Satrat I've tried out this PR, and while the overwriting of recipe args works as expected, the saved recipe still has the variable argument, i.e. without specific values. This I think will break other flows (e.g. evaluation). Could you please check that from your side?

Satrat:
> @Satrat I've tried out this PR, and while the overwriting of recipe args works as expected, the saved recipe still has the variable argument, i.e. without specific values. This I think will break other flows (e.g. evaluation). Could you please check that from your side?

Good catch! Yes we definitely want the output to reflect the evaluated args, I'll look into it further tomorrow

Satrat:
> > @Satrat I've tried out this PR, and while the overwriting of recipe args works as expected, the saved recipe still has the variable argument, i.e. without specific values. This I think will break other flows (e.g. evaluation). Could you please check that from your side?
> 
> Good catch! Yes we definitely want the output to reflect the evaluated args, I'll look into it further tomorrow

Just an update on this: I had hoped this would be a quick bug fix, but trying to fix this revealed the new modifier refactor doesn't handle evaluated args correctly during export, so its going to be a bit more involved to fix. I opened a ticket to track: https://app.asana.com/0/1203126676641557/1206186419092269/f and will leave this PR open until its resolved

dbogunowicz (dismissed):

bfineran (approved):

src/sparseml/transformers/sparsification/obcq/obcq.py
```diff
@@ -56,6 +56,7 @@ def one_shot(
     deploy_dir: Optional[str] = ".",
     recipe_file: Optional[str] = None,
     precision: str = "auto",
+    recipe_args: Optional[Dict] = None,
```
rahul-tuli, line 59:
```suggestion
    recipe_args: Optional[Dict[str, Any]] = None,
```

rahul-tuli: Might have to import Any as well

rahul-tuli (approved):
LGTM!

PR #1909, opened 2023-12-15
[Export Refactor] Export `transformers`

dbogunowicz:


src/sparseml/transformers/utils/load_task_dataset.py
```diff
@@ -0,0 +1,94 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
```
dbogunowicz, line 1: Copied over from the old export file.

src/sparseml/transformers/utils/load_task_model.py
```diff
@@ -0,0 +1,69 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
```
dbogunowicz, line 1: Copied over from the old export file.

src/sparseml/integration_helper_functions.py
```diff
@@ -32,10 +32,14 @@ class Integrations(Enum):
     """
 
     image_classification = "image-classification"
+    transformers = "transformers"
+    transformers_generative = "transformers-generative"
```
bfineran, line 36: for top level integration - can we keep it as `transformers` and imply generative based on task name? this seems a bit cumbersome and hidden from UX perspective

src/sparseml/integration_helper_functions.py
```diff
@@ -47,24 +51,46 @@ def resolve_integration(
...
         will attempt to infer it from the source_path.
+    :param task: Optional name of the task to use.
     :return: The name of the integration to use for exporting the model.
     """
 
     if integration is not None:
         integration = integration.replace("_", "-")
 
+    if task is not None:
+        task = task.replace("_", "-")
```
bfineran, line 62: shouldn't this be resolved at a more consolidated level?

src/sparseml/transformers/utils/initializers.py
```diff
@@ -0,0 +1,203 @@
...
+    "initialize_tokenizer",
+    "initialize_trainer",
+    "initialize_config",
+    "resolve_sequence_length",
+]
+
+_LOGGER = logging.getLogger(__name__)
+
+
+# TODO: Let's not take this for granted. We should be able to
```
bfineran, line 44: make sure to resolve all todos before merge

src/sparseml/transformers/utils/initializers.py
```diff
@@ -0,0 +1,203 @@
...
+# device argument
+@dataclass
+class ForceCPUTrainingArguments(TrainingArguments):
```
bfineran, line 47: what is the expected usage?

src/sparseml/transformers/utils/initializers.py
```diff
@@ -0,0 +1,203 @@
...
+    return model
+
+
+def initialize_trainer(
+    model: AutoModel,
+    model_path: Union[str, Path],
+    validation_dataset: Optional[Any] = None,
+) -> Trainer:
+    """
+    Initialize a trainer
```
bfineran, line 127: doc string incomplete - should mention if this applies recipe structure, etc

additionally, what is the plan to migrate post sparseml v2 update?

dbogunowicz: as mentioned in the follow-up PR, I will sync with @Satrat to learn about the most recent state of sparseml v2 and will update the logic. Let's not make it a blocker for landing this PR though.

src/sparseml/transformers/utils/initializers.py
dbogunowicz: This is a legacy object from the old export. Now when I think about this, it should go.

src/sparseml/integration_helper_functions.py
dbogunowicz: Let's discuss it today. 

I agree that it would be more appropriate to "fold" the `transformers_generative` under `transformers` in this hierarchy of argument. The contrary,however, still feels like a good implementation for one important reason: in our context, `transformers_generative` are quite distinct from the rest of `transformers`. They have their own optimizations, potentially recipe loading logic (need to look into it), different set of files. There is also high probability that the more we focus on LLMs, and the more popularity they gain, over time they will more-and-more become a separate concept from the "normal" transformers.

This is why I feel that the separate `IntegrationHelperFunctions` for `generative_transformers` makes sense, and from that follows the separate integration name.

src/sparseml/integration_helper_functions.py
dbogunowicz: Like in the `export.py`?

PR #1910, opened 2023-12-15
[Export Refactor] Export generative `transformers`

dbogunowicz:


src/sparseml/transformers/integration_helper_functions_generative.py
```diff
@@ -0,0 +1,44 @@
...
+)
+
+
+generative_transformers_graph_optimizations = {
+    "kv_cache_injection": apply_kv_cache_injection
+}
+
+
+@IntegrationHelperFunctions.register(name=Integrations.transformers_generative.value)
+class GenerativeTransformers(Transformers):
```
bfineran, line 38: the generative models - especially ones going through our current OBCQ/finetuning flow need a separate path to load the recipe from the new session framework right? this will be different from the existing automodel. 

Should sync with @Satrat on this if needed

bfineran (changes requested):

src/sparseml/transformers/integration_helper_functions_generative.py
dbogunowicz: We will discuss it during PP session

PR #1916, opened 2023-12-18
[Logger Refactor] Frequency Support

rahul-tuli:
This PR adds, Frequency Type support to `LoggerManager`

Key Changes include:
* Introduce simple FrequencyManager to manage the frequency of logging
* Add `FrequencyManagerFactory` to create FrequencyManagers
* Attach callbacks to `optimizer.step()` or `model forward function` based on Frequency Type
* Moved `logger.py` into it's own folder
* Refactor some epoch specific logging helpers to be more general

Supported Frequency types are (NOT CASE SENSITIVE) :
- [X] `epoch`
- [X] `optimizer_step`
- [X] `batch`

By default, `epoch` frequency type is selected with a `log_freq` of `0.1`; if this works for users they can omit specifying loggers with session.initialize altogether

For more fine-grained control, A `LoggerManager` instance should be created, with the expected log_frequency and frequency type

Example script to log model info after 10 optimizer steps:
```python
from sparseml.core.logger.logger import LoggerManager
import sparseml.core.session as sml
from sparseml.core.framework import Framework
import torchvision
from torchvision import transforms
import torch
from torch.utils.data import DataLoader
import datasets
import os
from torch.optim import Adam
from torch.nn import CrossEntropyLoss
from sparseml.core.event import EventType
from sparseml.core import PythonLogger
import logging

NUM_LABELS = 3
BATCH_SIZE = 32
NUM_EPOCHS = 1
LOG_FREQUENCY = 10
FREQUENCY_TYPE = "optimizer_step" # or "epoch" # or "batch"
recipe = "/home/rahul/projects/sparseml/local/recipe.yaml"
device = "cuda:0"

# set up SparseML session
sml.create_session()
session = sml.active_session()


# download model
model = torchvision.models.mobilenet_v2(weights=torchvision.models.MobileNet_V2_Weights.DEFAULT)
model.classifier[1] = torch.nn.Linear(model.classifier[1].in_features, NUM_LABELS)
model.to(device)


# download data
beans_dataset = datasets.load_dataset("beans")
train_folder, _ = os.path.split(beans_dataset["train"][0]["image_file_path"])
train_path, _ = os.path.split(train_folder)
val_folder, _ = os.path.split(beans_dataset["validation"][0]["image_file_path"])
val_path, _ = os.path.split(train_folder)


# dataloaders
imagenet_transform = transforms.Compose([
   transforms.Resize(size=256, interpolation=transforms.InterpolationMode.BILINEAR, max_size=None, antialias=None),
   transforms.CenterCrop(size=(224, 224)),
   transforms.ToTensor(),
   transforms.Normalize(mean=[0.485, 0.456, 0.406], std=[0.229, 0.224, 0.225])
])

train_dataset = torchvision.datasets.ImageFolder(
    root=train_path,
    transform=imagenet_transform
)
train_loader = DataLoader(train_dataset, BATCH_SIZE, shuffle=True, pin_memory=True, num_workers=16)

val_dataset = torchvision.datasets.ImageFolder(
    root=val_path,
    transform=imagenet_transform
)
val_loader = DataLoader(val_dataset, BATCH_SIZE, shuffle=False, pin_memory=True, num_workers=16)


# loss and optimizer
criterion = CrossEntropyLoss()
optimizer = Adam(model.parameters(), lr=8e-3)


recipe_args = {}
session_data = session.initialize(
    framework=Framework.pytorch,
    recipe=recipe,
    recipe_args=recipe_args,
    model=model,
    teacher_model=None,
    optimizer=optimizer,
    train_data=train_loader,
    val_data=val_loader,
    start=0.0,
    steps_per_epoch=len(train_loader),
    loggers=LoggerManager(frequency_type=FREQUENCY_TYPE, log_frequency=LOG_FREQUENCY),
)


# loop through batches
for epoch in range(NUM_EPOCHS):
    running_loss = 0.0
    total_correct = 0
    total_predictions = 0
    for step, (inputs, labels) in enumerate(session.state.data.train):
        inputs = inputs.to(device)
        labels = labels.to(device)
        session.state.optimizer.optimizer.zero_grad()
        session.event(event_type=EventType.BATCH_START, batch_data=(inputs, labels))

        outputs = session.state.model.model(inputs)
        loss = criterion(outputs, labels)
        loss.backward()
        session.event(event_type=EventType.LOSS_CALCULATED, loss=loss)

        session.event(event_type=EventType.OPTIM_PRE_STEP)
        session.state.optimizer.optimizer.step()
        session.event(event_type=EventType.OPTIM_POST_STEP)

        running_loss += loss.item()

        predictions = outputs.argmax(dim=1)
        total_correct += torch.sum(predictions == labels).item()
        total_predictions += inputs.size(0)

        session.event(event_type=EventType.BATCH_END)

    loss = running_loss / (step + 1.0)
    accuracy = total_correct / total_predictions
    print("Epoch: {} Loss: {} Accuracy: {}".format(epoch + 1, loss, accuracy))


# finalize session
session.finalize()
``` 

The recipe used:
```yaml
test_stage:
  pruning_modifiers:
    MagnitudePruningModifier:
      init_sparsity: 0.0
      final_sparsity: 0.5
      start: 0
      end: 2
      update_frequency: 0.5
      targets:        
          - features.0.0.weight
          - features.1.conv.0.0.weight
          - features.1.conv.1.weight
          - features.2.conv.0.0.weight
          - features.2.conv.1.0.weight
          - features.2.conv.2.weight
          - features.3.conv.0.0.weight
          - features.3.conv.1.0.weight
          - features.3.conv.2.weight
          - features.4.conv.0.0.weight
          - features.4.conv.1.0.weight
          - features.4.conv.2.weight
          - features.5.conv.0.0.weight
          - features.5.conv.1.0.weight
          - features.5.conv.2.weight
          - features.6.conv.0.0.weight
          - features.6.conv.1.0.weight
          - features.6.conv.2.weight
          - features.7.conv.0.0.weight
          - features.7.conv.1.0.weight
          - features.7.conv.2.weight
          - features.8.conv.0.0.weight
          - features.8.conv.1.0.weight
          - features.8.conv.2.weight
          - features.9.conv.0.0.weight
          - features.9.conv.1.0.weight
          - features.9.conv.2.weight
          - features.10.conv.0.0.weight
          - features.10.conv.1.0.weight
          - features.10.conv.2.weight
          - features.11.conv.0.0.weight
          - features.11.conv.1.0.weight
          - features.11.conv.2.weight
          - features.12.conv.0.0.weight
          - features.12.conv.1.0.weight
          - features.12.conv.2.weight
          - features.13.conv.0.0.weight
          - features.13.conv.1.0.weight
          - features.13.conv.2.weight
          - features.14.conv.0.0.weight
          - features.14.conv.1.0.weight
          - features.14.conv.2.weight
          - features.15.conv.0.0.weight
          - features.15.conv.1.0.weight
          - features.15.conv.2.weight
          - features.16.conv.0.0.weight
          - features.16.conv.1.0.weight
          - features.16.conv.2.weight
          - features.17.conv.0.0.weight
          - features.17.conv.1.0.weight
          - features.17.conv.2.weight
          - features.18.0.weight
          - classifier.1.weight
      leave_enabled: True
```

Output with above script will log model info 4 times; (the dataset used has 33 batches, and we log once for every 10 optimizer steps, we also only run it for 1 epoch)

Output:
```bash
Logging all SparseML modifier-level logs to sparse_logs/18-12-2023_20.20.46.log
2023-12-18 20:20:46 sparseml.core.logger.logger INFO     Logging all SparseML modifier-level logs to sparse_logs/18-12-2023_20.20.46.log
2023-12-18 20:20:47 sparseml.core.recipe.recipe INFO     Loading recipe from file /home/rahul/projects/sparseml/local/recipe.yaml
Epoch: 1 Loss: 0.7876473863919576 Accuracy: 0.7108317214700194
```

Quick scan through the log file confirms correct logging:
Screen Shot 2023-12-18 at 8 22 48 PM

The log file has been attached for reference:
[18-12-2023_20.20.46.log](https://github.com/neuralmagic/sparseml/files/13709991/18-12-2023_20.20.46.log)

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206205545989441
  - https://app.asana.com/0/0/1205848553025323

### DIFF
--- a/src/sparseml/core/lifecycle/session.py
+++ b/src/sparseml/core/lifecycle/session.py
@@ -17,7 +17,7 @@ from typing import Any, List, Optional
 
 from sparseml.core.event import EventType
 from sparseml.core.framework import Framework
-from sparseml.core.helpers import log_model_info
+from sparseml.core.helpers import log_model_info, should_log_model_info
 from sparseml.core.lifecycle.event import CallbacksEventLifecycle, EventLifecycle
 from sparseml.core.logger.utils import (
     BatchFrequencyManager,
@@ -121,11 +121,15 @@ class SparsificationLifecycle:
                 step_function()
                 current_step = optimizer.state[optimizer.param_groups[0]["params"][-1]][
                     "step"
-                ]
-                if self.state.loggers.log_ready(
-                    current_step, self.state._last_log_step
+                ].item()
+                if should_log_model_info(
+                    model=self.state.model,
+                    loggers=self.state.loggers,
+                    current_log_step=current_step,
+                    last_log_step=self.state._last_log_step,
                 ):
-                    log_model_info(state=self.state, epoch=current_step)
+                    log_model_info(state=self.state, current_log_step=current_step)
+                    self.state._last_log_step = current_step
 
             optimizer.step = _optimizer_step_fn
 
@@ -139,10 +143,18 @@ class SparsificationLifecycle:
 
             def _model_forward_fn(*args, **kwargs):
                 output = forward_function(*args, **kwargs)
-                if self.state.loggers.log_ready(
-                    self.event_lifecycle.current_index, self.state._last_log_step
+                current_step = self.event_lifecycle.current_index
+
+                if should_log_model_info(
+                    model=self.state.model,
+                    loggers=self.state.loggers,
+                    current_log_step=current_step,
+                    last_log_step=self.state._last_log_step,
                 ):
-                    log_model_info(state=self.state, epoch=self.state._last_log_step)
+                    log_model_info(
+                        state=self.state, current_log_step=self.state._last_log_step
+                    )
+                    self.state._last_log_step = current_step
                 return output
 
             model.forward = _model_forward_fn
--- a/src/sparseml/core/lifecycle/session.py
+++ b/src/sparseml/core/lifecycle/session.py
@@ -106,62 +106,6 @@ class SparsificationLifecycle:
 
         return mod_data
 
-    def _attach_logging_callbacks(self):
-        if (
-            isinstance(
-                self.state.loggers.frequency_manager, OptimizerStepFrequencyManager
-            )
-            and self.state.optimizer
-            and self.state.optimizer.optimizer
-        ):
-            optimizer = self.state.optimizer.optimizer
-            step_function = optimizer.step
-
-            def _optimizer_step_fn():
-                step_function()
-                current_step = optimizer.state[optimizer.param_groups[0]["params"][-1]][
-                    "step"
-                ].item()
-                if should_log_model_info(
-                    model=self.state.model,
-                    loggers=self.state.loggers,
-                    current_log_step=current_step,
-                    last_log_step=self.state._last_log_step,
-                ):
-                    log_model_info(state=self.state, current_log_step=current_step)
-                    self.state._last_log_step = current_step
-
-            optimizer.step = _optimizer_step_fn
-
-        elif (
-            isinstance(self.state.loggers.frequency_manager, BatchFrequencyManager)
-            and self.state.model
-            and self.state.model.model
-        ):
-            model = self.state.model.model
-            forward_function = model.forward
-
-            def _model_forward_fn(*args, **kwargs):
-                output = forward_function(*args, **kwargs)
-                current_step = self.state.loggers.epoch_to_step(
-                    epoch=self.event_lifecycle.current_index,
-                    steps_per_epoch=len(self.state.data.train),
-                )
-
-                if should_log_model_info(
-                    model=self.state.model,
-                    loggers=self.state.loggers,
-                    current_log_step=current_step,
-                    last_log_step=self.state._last_log_step,
-                ):
-                    log_model_info(
-                        state=self.state, current_log_step=self.state._last_log_step
-                    )
-                    self.state._last_log_step = current_step
-                return output
-
-            model.forward = _model_forward_fn
-
     def finalize(self, **kwargs) -> List[Any]:
         if not self.initialized_:
             raise ValueError("Cannot finalize before initializing")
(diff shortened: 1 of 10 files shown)


src/sparseml/core/session.py
```diff
@@ -282,18 +282,21 @@ def _log_model_info(self, event_type: EventType):
...
-            epoch=epoch,
-            last_log_epoch=self.state._last_log_epoch,
+        if (
+            should_log_model_info(
+                model=self.state.model,
+                loggers=self.state.loggers,
+                current_log_step=epoch,
+                last_log_step=self.state._last_log_step,
+            )
+            and self.state.loggers.frequency_manager.frequency_type == "epoch"
```
rahul-tuli, line 292: These will only activate when frequency type is epoch

src/sparseml/core/lifecycle/session.py
```diff
@@ -155,6 +162,66 @@ def event(self, event_type: EventType, **kwargs) -> List[Any]:
...
+
+            optimizer.step = _optimizer_step_fn
+
+        elif (
+            isinstance(self.state.loggers.frequency_manager, BatchFrequencyManager)
+            and self.state.model
+            and self.state.model.model
+        ):
+            model = self.state.model.model
+            forward_function = model.forward
```
Satrat, line 200: When possible we should try to avoid using PyTorch specific calls within core. A workaround would be to add a forward function to `ModifiableModel` and then call self.state.model.forward.

src/sparseml/core/logger/utils/frequency_managers.py
```diff
@@ -0,0 +1,209 @@
...
+            or current_log_step >= last_log_step + self._log_frequency
+        )
+
+    def validate_log_frequency(self, log_frequency: Union[float, int, None]):
+        """
+        Validates the log frequency is a float or int or None, raising
+        a ValueError if invalid
+
+        :param log_frequency: _description_
+        :raises ValueError: _description_
```
Satrat, line 97: missing description?

src/sparseml/core/logger/utils/frequency_managers.py
```diff
@@ -0,0 +1,209 @@
...
+        """
+        if not isinstance(log_frequency, (float, int, type(None))):
+            raise ValueError(
+                f"frequency {log_frequency} must be a float or int or None"
+                f" But found {type(log_frequency)} instead"
+            )
```
Satrat, line 103: Could we also check that its > 0?

src/sparseml/core/logger/utils/frequency_managers.py
```diff
@@ -0,0 +1,209 @@
...
+        Validates the log frequency is an int or None, raising a ValueError if invalid
+        Also validates that the frequency is >= 100 (to avoid over-logging) if not None
+
+        :param log_frequency: the log frequency to validate
+        :raises ValueError: if the log frequency is neither an int or None
+        :raises ValueError: if the log frequency is < 10
+        """
+        super().validate_log_frequency(log_frequency)
+        if log_frequency is not None and log_frequency < 10:
+            raise ValueError(f"frequency {log_frequency} must be >= 10")
```
Satrat, line 151: docstrings are inconsistent with the code, class definition says >=16, fn definition says <100, actual code says >=10. Also could you clarify why we don't allow < 10?

src/sparseml/core/logger/utils/frequency_managers.py
```diff
@@ -0,0 +1,209 @@
...
+        Validates the log frequency is an int or None, raising a ValueError if invalid
+        Also validates that the frequency is >= 4 (to avoid over-logging) if not None
+
+        :param log_frequency: the log frequency to validate
+        :raises ValueError: if the log frequency is neither an int or None
+        :raises ValueError: if the log frequency is < 4
+        """
+        super().validate_log_frequency(log_frequency)
+        if log_frequency is not None and log_frequency < 4:
+            raise ValueError(f"frequency {log_frequency} must be >= 4")
```
Satrat, line 173: Same comment as above

tests/sparseml/core/test_helpers.py
```diff
@@ -74,9 +78,9 @@ def state_mock():
 
 def test__log_epoch_invokes_log_scalar():
```
Satrat, line 79: nit: extra _ in fn definition?

Satrat (changes requested):
Nice work on the logger classes, had a few suggestions but nothing major. Also would be good to see some unit tests for each of the logging types(epoch, batch, optim)

src/sparseml/core/lifecycle/session.py
rahul-tuli: Good Point, Noted will make that change

src/sparseml/core/logger/utils/frequency_managers.py
rahul-tuli: Yes, this was a brain fart on my end

src/sparseml/core/logger/utils/frequency_managers.py
rahul-tuli: It is done to avoid over-logging; however I think we should remove this constraint all together and chalk it up to user error if a lower frequency is specified

Satrat: Agreed! if they really want logging every batch thats their problem :)

rahul-tuli:
Closing as requirements changed slightly and new changes are a part of #1931

PR #1860, opened 2023-11-28
[Feature Branch] Logger Framework

rahul-tuli:
This PR serves as the feature branch for the logger framework refactor, all PRs pertaining to loggers will be merged into this, and this branch will land into main, I have tried to use a stacked PR approach here:

- [X] Move existing logger manager into `sparseml.core` and Update existing modifiers to use logger manager from `sparseml.core`: This has been merged into main #1759 
- [x] Add model level logs #1783
  - [x] Add convenience methods for log_strings #1855 
     - [x] Output info logs to console and all logs to log-file #1861
 - [X] #1924 Moves logger into it's own package
 - [X] #1925 Adds log dirs to `.gitignore`
 - [x] #1927 Base functionality for `FrequencyManager` (w/o toggles)
 - [X] #1933 Prettier log format
 - [x] #1930 `FrequencyManager` toggles
   - [x]  #1931 Add frequency manager to LoggerManager
     - [x] #1932 Add system and metric logging wrappers (addressed review comments)
       - [x] #1920 Session based Logs  
           - [x] #1934 Log losses  
       - [x] #1919 TB logging 
         - [x] #1918 WANDB logging function
             - [x] #1936 Activate TB and WANDB logging 

Fix a failing test:
- [x] https://github.com/neuralmagic/sparseml/pull/1963
---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206212680976480

PR #1924, opened 2023-12-20
[Move] logger into it's own package

rahul-tuli:
improves structure and separation of concerns

This change is backwards compatible i.e the following import would still work:
```python
from sparseml.core import LoggerManager
```

based off of #1860


bfineran (approved):

Satrat (approved):
LGTM!

PR #1927, opened 2023-12-21
[Add] frequency_manager to check log "readiness"

rahul-tuli:
This PR adds a `FrequencyManager` class

Responsibilities of this class:
 - Ascertain if we should log at the current step or not

Based off of #1924

Testing: Added automated tests

### DIFF
--- a/src/sparseml/core/logger/utils/frequency_manager.py
+++ b/src/sparseml/core/logger/utils/frequency_manager.py
@@ -69,7 +69,9 @@ class FrequencyManager:
             >= float(format(self.last_log_step + self.log_frequency, ".4f"))
         )
 
-        if not (check_model_update and cadence_reached):
+        if not cadence_reached or not check_model_update:
+            # early return if cadence not reached or,
+            # model update check not requested
             return cadence_reached
 
         model_updated_since_last_log: bool = (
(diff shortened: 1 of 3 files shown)


src/sparseml/core/logger/utils/frequency_manager.py
```diff
@@ -0,0 +1,146 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from typing import Union
+
+
+__all__ = ["FrequencyManager"]
+
+LogStepType = Union[int, float, None]
```
bfineran, line 21: Why do we call out int? Is there any int specific logic?

bfineran (changes requested):
Shouldn't the frequency be set differently based on additional args? Ie we only care about model update if we're in on_change mode

src/sparseml/core/logger/utils/frequency_manager.py
rahul-tuli: Int will be needed for `optimizer_step` and `batch` frequency types; the support for that will come through in a different PR

rahul-tuli:
> Shouldn't the frequency be set differently based on additional args? Ie we only care about model update if we're in on_change mode

Support for different modes, like `on_change` and `exact` will come through in a different PR.

bfineran (approved):

dbogunowicz (approved):

PR #1929, opened 2023-12-21
[BugFix] Move to use virtualenv over venv

rahul-tuli:
The dockerfile failed to build because python3-venv does not work well with multiple python versions; using virtualenv fixes this by explicitly specifying the python version; `using -m also helps`

This should not have any implications on the final image size

Creating and activating a virtual env has also been moved into it's own `RUN` command


bfineran (approved):

andy-neuma (approved):
cool

PR #1930, opened 2023-12-26
[Logger Framework][Add] mode and frequency type to FrequencyManager

rahul-tuli:
This PR adds `frequency_type` ("epoch", "step", "batch") and `mode` ("exact", "on_change") to frequency manager:

```
The frequency type attribute controls what the frequency manager is tracking, e.g. if the frequency type
is "epoch", then the frequency manager will track the number of epochs that
have passed since the last log, if the frequency type is "step", then the
frequency manager will track the number of optimizer steps
```

```
 The logging mode to use, either "on_change" or "exact",
"on_change" will log when the model has been updated since the last log,
 "exact" will log at the given frequency regardless of model updates.
```

The frequency manager, now also keeps track of the last step when model 
was updated. Whenever updated the it is the responsibility of users to 
call `model_updated(step)` function.

These two arguments will provide more fine-grained control over the logs, and how to filter them to avoid over-logging

based off of #1860

### DIFF
--- a/tests/sparseml/core/logger/utils/test_frequency_manager.py
+++ b/tests/sparseml/core/logger/utils/test_frequency_manager.py
@@ -139,10 +139,9 @@ def test_log_ready(
     assert actual == expected
 
 
-def _test_cases__validate_log_frequency():
-    # "log_frequency, frequency_type, expectation"
-
-    return [
+@pytest.mark.parametrize(
+    "log_frequency, frequency_type, expectation",
+    [
         # epoch frequency type accepts floats
         (0.1, "epoch", does_not_raise()),
         # batch and step frequency types need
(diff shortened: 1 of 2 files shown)


tests/sparseml/core/logger/utils/test_frequency_manager.py
```diff
@@ -137,3 +137,29 @@ def test_log_ready(
     )
 
     assert actual == expected
+
+
+def _test_cases__validate_log_frequency():
+    # "log_frequency, frequency_type, expectation"
+
+    return [
```
bfineran, line 145: why not just include below with the pytest call?

bfineran (approved):

tests/sparseml/core/logger/utils/test_frequency_manager.py
rahul-tuli: No reason in particular, will move it below

PR #1931, opened 2023-12-27
[Use] frequency manager in logger

rahul-tuli:
This PR updates `LoggerManager` to use `FrequencyManager`

Few properties like `log_frequency` was moved
Add wrapper functions `model_updated` and `log_written` to `LoggerManager`

Based off of #1930

### DIFF
--- a/src/sparseml/core/logger/utils/frequency_manager.py
+++ b/src/sparseml/core/logger/utils/frequency_manager.py
@@ -18,14 +18,14 @@ from typing import Literal, Union
 
 __all__ = [
     "FrequencyManager",
-    "PossibleLoggingMode",
-    "PossibleFrequencyType",
+    "LoggingModeType",
+    "FrequencyType",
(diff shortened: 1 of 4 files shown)


src/sparseml/core/logger/utils/frequency_manager.py
```diff
@@ -16,7 +16,12 @@
 from typing import Literal, Union
 
 
-__all__ = ["FrequencyManager"]
+__all__ = [
+    "FrequencyManager",
+    "PossibleLoggingMode",
```
bfineran, line 21: why not just `LoggingMode`?

rahul-tuli: Updated to `LoggingModeType` to avoid confusion

bfineran (approved):

rahul-tuli:
Merging as it was approved by Ben!

PR #1934, opened 2024-01-03
[Logger Framework] Log Losses

rahul-tuli:
This PR adds support to log losses in the new sparsification framework;
Based off of #1920 

Example log file has been attached for GMP with `mobilenet_v2`

[latest-log-gmp-mobilenet-12-01-2024_07.58.30.log](https://github.com/neuralmagic/sparseml/files/13919345/latest-log-gmp-mobilenet-12-01-2024_07.58.30.log)

### DIFF
--- a/src/sparseml/core/session.py
+++ b/src/sparseml/core/session.py
@@ -65,6 +65,7 @@ class SparseSession:
 
     def __init__(self):
         self._lifecycle = SparsificationLifecycle()
+        self._loss_logged = False
 
     @property
     def lifecycle(self) -> SparsificationLifecycle:
@@ -220,6 +221,10 @@ class SparseSession:
         :param kwargs: additional kwargs to pass to the lifecycle's finalize method
         :return: the modified state of the session after finalizing
         """
+        # log losses on finalization
+        self._loss_logged = False
+        self._log_loss(event_type=EventType.LOSS_CALCULATED, loss=self.state.loss)
+
         mod_data = self._lifecycle.finalize(**kwargs)
 
         return ModifiedState(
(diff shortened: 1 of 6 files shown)


src/sparseml/core/session.py
```diff
@@ -261,39 +266,66 @@ def event(
...
 
+    def log(self, event_type: EventType, loss: Optional[Any] = None):
+        """
+        Log model and loss information for the current event type
+
+        :param event_type: the event type to log for
+        :param loss: the loss to log if any
+        """
+        self._log_model_info()
+        self._log_loss(event_type=event_type, loss=loss)
```
bfineran, line 291: should only call if loss is not none

src/sparseml/core/session.py
```diff
@@ -65,6 +65,7 @@ class SparseSession:
 
     def __init__(self):
         self._lifecycle = SparsificationLifecycle()
+        self._loss_logged = False
```
bfineran, line 68: let's do last logged step instead to decouple with log model info

src/sparseml/core/session.py
```diff
@@ -261,39 +266,66 @@ def event(
...
+
+    def _log_loss(self, event_type: EventType, loss: Any):
+        if event_type != EventType.LOSS_CALCULATED or self._loss_logged:
+            return
+
+        current_step = self._lifecycle.event_lifecycle.current_index
+        self.state.loggers.metric.log_scalars(
+            tag="Loss", values=loss, step=current_step
+        )
+        self._loss_logged = True
```
bfineran, line 328: i.e. 
```suggestion
        self._last_loss_logged = current_step
```

bfineran (approved):

PR #1940, opened 2024-01-05
Expand Finetuning Dataset Registry to Match One-Shot

Satrat:
Adding 4 new datasets to the `TextGenerationDataset` registry: gsm8k, ptb, ultrachat_200k, evolcodealpaca. These datasets can now be used for finetuning.

### Testing
Added a unit test for each dataset load

Note that evolcodealpaca seems to be broken on the hugging face side, this command fails, so I've skipped the test for now
```
from datasets import load_dataset

dataset = load_dataset("theblackcat102/evol-codealpaca-v1")
```

**Note:** In a future PR, I will deprecate the one-shot specific `TransformersDataset`, so both finetuning and oneshot use the same registry

### DIFF
--- a/src/sparseml/transformers/finetune/README.md
+++ b/src/sparseml/transformers/finetune/README.md
@@ -1,0 +1,76 @@
+# Sparse Finetuning
+
+## Launching from Console Scripts
+
+### with DataParallel (default)
+
+```bash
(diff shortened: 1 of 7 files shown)


rahul-tuli (approved):
LGTM!

bfineran (approved):

PR #1948, opened 2024-01-10
[Export refactor] final manual testing fixes

bfineran:
fixes from testing LLMs export from a fresh python 3.8 environment with minimal extras installed (just nm-transformers and torch)

### DIFF
--- a/src/sparseml/integration_helper_functions.py
+++ b/src/sparseml/integration_helper_functions.py
@@ -63,7 +63,7 @@ def resolve_integration(
         from sparseml.transformers.utils.helpers import is_transformer_model
     except ImportError:
         # unable to import integration, always return False
-        is_transformer_model = _null_is_model()
+        is_transformer_model = _null_is_model
 
     if (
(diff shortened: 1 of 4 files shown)


src/sparseml/integration_helper_functions.py
```diff
@@ -51,10 +51,19 @@ def resolve_integration(
...
-        is_image_classification_model,
-    )
-    from sparseml.transformers.utils.helpers import is_transformer_model
+    try:
+        from sparseml.pytorch.image_classification.utils.helpers import (
+            is_image_classification_model,
+        )
+    except ImportError:
+        # unable to import integration, always return False
+        is_image_classification_model = _null_is_model
```
Satrat, line 60: should this be _null_is_model() like on line 66?

Satrat (approved):

src/sparseml/integration_helper_functions.py
bfineran: 66 should be like this one, good catch

PR #1951, opened 2024-01-11
Skip tests that do not work on CentOS 7

dbarbuzzi:
This PR adds a conditional skip to a few tests that do not work on CentOS 7 due to external dependencies.

### DIFF
--- a/tests/sparseml/pytorch/datasets/classification/test_cifar.py
+++ b/tests/sparseml/pytorch/datasets/classification/test_cifar.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import os
+import platform
 
 import pytest
 from torch.utils.data import Dataset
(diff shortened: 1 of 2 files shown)


dbarbuzzi:
It looks like the `test-setup` check was broken with #1858 - the `run` command starting on line 60 of `https://github.com/neuralmagic/sparseml/pull/1858/files#diff-b105740d4a5ab1ea47925dd77d65be298e61913a5885693220e322871ba8bd4e` has imbalanced parentheses (two open, one close); it’s also two-line shorter than equivalent checks which are using the same "template", so perhaps those lines are missing from that block.

bfineran (approved):
nice, didn't know about this feature in param

dbarbuzzi:
Closing this as a solution was found to update the host environment.

PR #1957, opened 2024-01-12
Checkpoint Restarting for Alternating Flow

Satrat:


src/sparseml/transformers/finetune/runner.py
```diff
@@ -208,13 +214,18 @@ def predict(self):
...
         Run the recipe stage by stage, allowing for alternating between one-shot and
         finetuning flows. Optionally save the model output at the end of each stage
+
+        :param checkpoint: optional checkpoint to pick up a stage from
         """
 
         recipe_obj = Recipe.create_instance(self._training_args.recipe)
+        with self.trainer.accelerator.main_process_first():
+            completed_stages = get_completed_stages(self._model_args.model_name_or_path)
+        self.trainer.accelerator.wait_for_everyone()
```
dbogunowicz, line 228: just for more context, what does accelerator, as well as this method do?

dbogunowicz (approved):

src/sparseml/transformers/finetune/runner.py
Satrat: accelerator is the FSDP manager class (its a hugging face library). When we run training in FSDP mode it launches as separate processes, when doing something like writing to disk we don't want the processes to access something at the same time. main_process_first is a context manager that guarantees the code block runs on GPU0 first before the other GPUs. the wait_for_everyone_call waits for all GPUs to reach that line before continuing

bfineran (approved):

PR #1967, opened 2024-01-17
[Logger Refactor] Add timer in logger manager

rahul-tuli:
This PR adds a time context manager to LoggerManager

```python
from time import sleep
from sparseml.core import LoggerManager
from logging import INFO
logger_manager = LoggerManager()

with logger_manager.time('test', level=INFO):
    sleep(1)
```

Output:
```bash
Logging all SparseML modifier-level logs to sparse_logs/17-01-2024_08.49.07.log
2024-01-17 08:49:07 sparseml.core.logger.logger INFO     Logging all SparseML modifier-level logs to sparse_logs/17-01-2024_08.49.07.log
manager test: 1.000s

```

### DIFF
--- a/src/sparseml/core/logger/logger.py
+++ b/src/sparseml/core/logger/logger.py
@@ -21,6 +21,7 @@ import os
 import time
 import warnings
 from abc import ABC
+from contextlib import contextmanager
 from datetime import datetime
 from logging import CRITICAL, DEBUG, ERROR, INFO, WARN, Logger
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


Satrat (approved):
nice and clean!

src/sparseml/core/logger/logger.py
```diff
@@ -1123,6 +1124,25 @@ def log_string(
...
+
+        :param tag: identifying tag to log the values with
+        """
+
+        start = time.time()
+        yield
+        elapsed = time.time() - start
+        if not tag:
+            tag = f"{DEFAULT_TAG}_time"
+        self.log_string(tag=tag, string=f"{elapsed:.3f}s", *args, **kwargs)
```
bfineran, line 1144: We should be logging a float not a string, will allow for easier plotting and such

bfineran (changes requested):

dbogunowicz (approved):

PR #1971, opened 2024-01-18
[Export Refactor] Reduce coupling of the `export` module with transformers; make `SparseAutoModel` more autonomous

dbogunowicz:


bfineran (dismissed):

bfineran (dismissed):

src/sparseml/transformers/utils/helpers.py
```diff
@@ -75,6 +75,22 @@ class TaskNames(Enum):
...
+    past key value support. This means that the model initialized from
+    this config will not take past key values as input and will not output
+    past key values.
+    """
+    # not take past_key_values as input
+    config.is_decoder = True
+    # whether to use past key values an input
+    config.use_past = False
+    # whether to output past key values
+    config.use_cache = False
```
mgoin, line 90: is there potential to fail if these members don't exist?

mgoin (dismissed):

src/sparseml/transformers/utils/helpers.py
dbogunowicz: Have you seen an instance of a config of a CausalModel (text-generation) that did not have those attributes? They are pretty standard no?

mgoin: yeah i was just thinking from a possible null access perspective, but that doesn't matter for python members

mgoin (approved):

bfineran (approved):

PR #1972, opened 2024-01-18
[Export Refactor] Warn the user when exporting repeatedly 

dbogunowicz:
## Motivation
Exporting the same model over and over again can be a bit risky. While we are unlikely to tweak any of the files in the original training folder (files such as model.bin, model.pth, config.json etc), we may run into a few minor issues. Let's illustrate it with an example.

### Training folder before the export 
```bash
-> file_1
-> file_2
...
```
### Training folder after the export (assuming default arguments and sample export)
```bash
-> file_1
-> file_2
...
-> sample-inputs
-> sample-outputs
...
-> deployment
```

Now, if the user decides to export once again, in worst case scenario the following thing might happen
- the samples (inputs, outputs, ...) will get overwritten (if the user swaps the dataset used for export, this might silently swap samples from one dataset for samples from a second dataset)
- we might get slightly chaotic inside the `deployment` folder when it comes to `model.onnx` and potentially `model.data`. Users may decide to export once with external data, and then without external data etc.. This may potentially leave the user with a corrupted `model.data` or `model.onnx`.

## Feature Description:
- when samples get overwritten, print the verbose warning
- when the deployment folder gets overwritten, we print the verbose warning and purge the deployment folder before populating it with the newest data


tests/sparseml/export/transformers/test_transformers.py
```diff
@@ -28,7 +28,7 @@
     "stub, task",
     [
         ("zoo:obert-medium-squad_wikipedia_bookcorpus-pruned95_quantized", "qa"),
-        ("zoo:roberta-large-squad_v2_wikipedia_bookcorpus-base", "qa"),
+        # ("zoo:roberta-large-squad_v2_wikipedia_bookcorpus-base", "qa"),
```
dbogunowicz, line 31:
```suggestion
        ("zoo:roberta-large-squad_v2_wikipedia_bookcorpus-base", "qa"),
```

bfineran (approved):

bfineran (approved):

mgoin (approved):

PR #1973, opened 2024-01-18
[Eval Refactor] Perplexity Evaluation Sparseml

rahul-tuli:
This PR will add perplexity support to `SparseMLEvalRegistry`

Based off of #2001


src/sparseml/evaluation/integrations/perplexity.py
```diff
@@ -0,0 +1,155 @@
...
+from tqdm import tqdm
+
+from huggingface_hub import hf_hub_download
+from sparseml.evaluation.registry import SparseMLEvaluationRegistry
+from sparseml.transformers.integration_helper_functions import create_model
+
+
+@SparseMLEvaluationRegistry.register("perplexity")
+def perplexity_eval(
+    target,
```
dbogunowicz, line 30: why are we creating the model inside the evaluator function and not before? Seems wasteful to add model initialisation logic to every evaluator function. So I suggest `target: Module`.

src/sparseml/evaluation/integrations/perplexity.py
```diff
@@ -0,0 +1,155 @@
...
+    datasets: str = "wikitext",
+    batch_size: int = 1,
+    device: Optional[str] = None,
+):
+    nsamples = 100
```
dbogunowicz, line 35: Let's have nsamples as an argument (default to None)

src/sparseml/evaluation/integrations/perplexity.py
```diff
@@ -0,0 +1,155 @@
...
+    dataset_config_name = infer_dataset_config_name(datasets)
+    task = "text-generation"
+    try:
+        # download the recipe if any
+        recipe_path = hf_hub_download(repo_id=target, filename="recipe.yaml")
```
dbogunowicz, line 40: What about sparsezoo case?

src/sparseml/evaluation/integrations/perplexity.py
```diff
@@ -0,0 +1,155 @@
...
+        recipe_path = None
+
+    model, other_objects = create_model(
+        source_path=target,
+        device=device,
+        task=task,
+        recipe=recipe_path,
+    )
+
+    tokenizer = other_objects["tokenizer"]
```
dbogunowicz, line 51: Could we move the code for dataset processing elsewhere? Or reuse @Satrat 's dataset registries if possible

rahul-tuli: Done!

src/sparseml/evaluation/integrations/perplexity.py
rahul-tuli: This was done because we need the target_name or model_path to create the tokenizer, which would be not be available at the integration level if we pass in the module.

src/sparseml/evaluation/integrations/perplexity.py
```diff
@@ -0,0 +1,202 @@
...
+    datasets: str = "wikitext",
+    batch_size: int = 1,
+    device: Optional[str] = None,
+    nsamples: Optional[int] = None,
+) -> Result:
+    dataset_config_name = _infer_dataset_config_name(datasets)
+    task = "text-generation"
+    split = "test"
+    recipe_path = fetch_recipe_path(target)
+    model, other_objects = create_model(
```
bfineran, line 49: noting we'll need to update to the sparse automodel/tokenizer

src/sparseml/evaluation/integrations/perplexity.py
```diff
@@ -0,0 +1,202 @@
...
+    input_text = _load_perplexity_dataset(
+        dataset_name=datasets,
+        dataset_config_name=dataset_config_name,
+        split=split,
+        nsamples=nsamples,
+    )
+    add_start_token = True
+    max_length = None
+
+    # if batch_size > 1 (which generally leads to padding being required), and
```
bfineran, line 65: let's add a link to the source code this is adapted from

src/sparseml/evaluation/integrations/perplexity.py
```diff
@@ -0,0 +1,202 @@
...
+        assert torch.all(
+            torch.ge(attn_masks.sum(1), 2)
+        ), "When add_start_token=False, each input text must be at least two "
+        "tokens long. Run with add_start_token=True if inputting strings of "
+        "only one token, and remove all empty input strings."
+
+    ppls = []
+    loss_fct = CrossEntropyLoss(reduction="none")
+
+    for start_index in tqdm(range(0, len(encoded_texts), batch_size)):
```
bfineran, line 116: we should add add a few in-line comments for what is going on here

src/sparseml/evaluation/utils/helpers.py
```diff
@@ -0,0 +1,65 @@
...
+from pathlib import Path
+
+from huggingface_hub import hf_hub_download
+from sparsezoo import Model
+
+
+__all__ = ["fetch_recipe_path"]
+
+
+def fetch_recipe_path(target: str):
```
bfineran, line 25: can we move this to a common util path
cc @dbogunowicz @Satrat

tests/sparseml/evaluation/integrations/test_perplexity.py
```diff
@@ -0,0 +1,210 @@
...
+):
+    import numpy
+    import torch
+    import tqdm
+    from datasets import load_dataset
+    from torch.nn import CrossEntropyLoss
+    from transformers import AutoModelForCausalLM, AutoTokenizer
+
+    # _compute() is taken as is from
+    # https://github.com/huggingface/evaluate/metrics/perplexity/perplexity.py
```
bfineran, line 86: let's consider adding this as a dev dependency so we will have an up to date source of truth and clean up this file a bit

bfineran (approved):

src/sparseml/evaluation/integrations_config.yaml
```diff
@@ -14,3 +14,4 @@
 # my_eval_integration: integrations/my_eval_integration.py
 # my_eval_integration2: integrations/my_eval_integration2.py
 
+perplexity: integrations/perplexity.py
```
bfineran, line 17: would be good to have any sort of coverage for this to protect against files being moved since we're referencing relative path

rahul-tuli: Added as a part of #2043

PR #1974, opened 2024-01-18
Zero Out Padding in SparseGPT Calculation

Satrat:
When alternating between finetuning and oneshot, we want to zero out the padding tokens for SparseGPT so that they do not affect the hessian calculation, but we want to keep the padding tokens for other modifiers. To solve this, we add a padding mask column to the calibration data, and (optionally) apply it during the forward pass of calibration. This should also enable us to batch one-shot calculations

### Summary of Changes
* new helper function in `TextGenerationDataset` to calculate and store padding masks when tokenizing a dataset
* Updated finetuning script to store padding masks for calibration data
* Updated `run_calibration_forward` to optionally apply a padding mask to input_ids, this is only enabled for SparseGPT/Wanda modifiers
* (no changes were made to the old OBCQ script, we just don't pad the input at all there)

### Testing
* confirmed launching OBCQ from `obcq.py` and from the finetuning script `text_generation.py` still work as expected
* unit test for padding column

Asana ticket: https://app.asana.com/0/1201735099598270/1206212842563198/f


dbogunowicz (approved):

rahul-tuli (approved):
Nice!

bfineran (approved):

PR #1978, opened 2024-01-19
[Export Refactor] Keep `model-orig.onnx` when exporting LLMs with KV Cache Injection

dbogunowicz:


bfineran (dismissed):

src/sparseml/transformers/utils/helpers.py
```diff
@@ -58,14 +58,16 @@ class TaskNames(Enum):
...
+ONNX_MODEL_NAME_INTERMEDIATE = "model-orig.onnx"
 RECIPE_NAME = "recipe.yaml"
 MANDATORY_DEPLOYMENT_FILES = {
     ONNX_MODEL_NAME,
     "tokenizer_config.json",
     "config.json",
 }
-NLG_TOKENIZER_FILES = {"special_tokens_map.json", "vocab.json", "merges.txt"}
 OPTIONAL_DEPLOYMENT_FILES = {"tokenizer.json", "tokenizer.model"}
+NLG_MANDATORY_DEPLOYMENT_FILES = {"special_tokens_map.json", "vocab.json", "merges.txt"}
```
mgoin, line 69: with we assert somewhere if these files aren't available from the original model? i'm pretty sure i've seen several models without these files, such as llama chat https://huggingface.co/meta-llama/Llama-2-7b-chat-hf/tree/main there is no vocab or merges file

dbogunowicz: @mgoin this is outside of the scope of this PR. I moved some of those files from mandatory to optional in #1962

bfineran (approved):

mgoin (approved):

PR #1980, opened 2024-01-19
FSDP OneShot NaN/memory fix

Satrat:
Running forward passes of large models in FSDP mode was resulting in NaNs, even before any modifiers were applied. 

The root cause was a weird FSDP edge case where calling `summon_full_params` on a non-root module (for instance on model.layers.0) would set the sub-module as the root node, corrupting the model. A simple fix was to call a dummy forward pass of the FSDP model before starting one-shot, so that the root of the model gets initialized correctly. Basically FSDP doesn't seem to account for the use case where you would want to access model weights before running a forward pass ¯\_(ツ)_/¯

With this fix we also no longer have to hackily call a blank training run to initialize the model, we can just run accelerator.prepare(). Also adding some empty_cache() calls between stages to clear out memory

### Test Script
```python
  from sparseml.transformers.finetune.text_generation import run_general
  
  model = "../ml-experiments/nlg-text_generation/llama_pretrain-llama_7b-base/dense/training"
  dataset_name = "open_platypus"
  concatenate_data = False
  run_stages=True
  output_dir = "./output_finetune_multi"
  recipe = est_multi_recipe.yaml"
  num_train_epochs=0.5
  overwrite_output_dir = True
  splits = {
      "train": "train[:95%]",
      "calibration": "train[95%:100%]"
  }

  run_general(
      model_name_or_path=model,
      dataset_name=dataset_name,
      run_stages=run_stages,
      output_dir=output_dir,
      recipe=recipe,
      num_train_epochs=num_train_epochs,
      overwrite_output_dir=overwrite_output_dir,
      concatenate_data = concatenate_data,
      remove_unused_columns = False,
      splits = splits
    )
```

test_multi_recipe.yaml
```
test_first_oneshot_stage:
  obcq_modifiers:
    SparseGPTModifier:
      sparsity: 0.7
      block_size: 128
      sequential_update: False
      quantize: False
      percdamp: 0.01
      prunen: 0
      prunem: 0
      targets: [
        "model.layers.0"
      ]
test_train_stage:
  pruning_modifiers:
    ConstantPruningModifier:
      targets: [
        "re:.*self_attn.q_proj",
        "re:.*self_attn.k_proj",
        "re:.*self_attn.v_proj",
        "re:.*self_attn.o_proj",
        "re:.*mlp.gate_proj",
        "re:.*mlp.up_proj"
      ]
      start: 0
```


bfineran (approved):

bfineran:
great find

bfineran (dismissed):

bfineran (approved):

rahul-tuli (approved):

PR #1997, opened 2024-01-24
Fix for unexpected column

Satrat:
https://github.com/neuralmagic/sparseml/commit/2e264e04760ca66266ce7249ad4cb71c6ec41c1f introduced a bug for quantization modifiers, we need to remove the padding column if it is unused to avoid a `TypeError: LlamaForCausalLM.forward() got an unexpected keyword argument 'padding_mask'`


mwitiderrick (commented):
Worked

dbogunowicz (approved):

PR #2002, opened 2024-01-24
Update Finetuning README with Alternating Flow Instructions

Satrat:


src/sparseml/transformers/finetune/example_alternating_recipe.yaml
```diff
@@ -0,0 +1,54 @@
...
+  pruning_modifiers:
+    ConstantPruningModifier:
+      targets: [
+        "re:.*self_attn.q_proj",
+        "re:.*self_attn.k_proj",
+        "re:.*self_attn.v_proj",
+        "re:.*self_attn.o_proj",
+        "re:.*mlp.gate_proj",
+        "re:.*mlp.up_proj"
+      ]
```
mgoin, line 53: why can't we just use `_ALL_` here like we used previous for sparse transfer?

src/sparseml/transformers/finetune/example_alternating_recipe.yaml
```diff
@@ -0,0 +1,54 @@
...
+  run_type: oneshot
+  obcq_modifiers:
+    SparseGPTModifier:
+      sparsity: 0.7
+      block_size: 128
+      sequential_update: False
+      quantize: False
+      percdamp: 0.01
+      prunen: 0
+      prunem: 0
```
mgoin, line 38: i thought this was replaced by `mask_structure: "0:0"` - does this still work? that's fine for backwards-compatibility

src/sparseml/transformers/finetune/example_alternating_recipe.yaml
Satrat: Yeah good point, `__ALL__` should work but let me test it quickly first

src/sparseml/transformers/finetune/example_alternating_recipe.yaml
Satrat: Yeah we didn't remove the args, but mask_structure is the recommended format so I'll swap it

rahul-tuli (approved):

bfineran (approved):

PR #1991, opened 2024-01-23
[Export Refactor] Allow for passing kwargs in the CLI

dbogunowicz:
## Feature Description

Passing arbitrary arguments e.g.

```bash
sparseml.export "" --sequence_length 10 --other "kwarg" --integration transformers
```

will result in passing `kwargs = dict(other="kwarg", sequence_length=10)`. Confirmed through manual testing

PR #2004, opened 2024-01-24
Fix integration choices in export.py

mgoin:
Previously this would fail if `--integration` was defined because the only valid value was the unfortunate string "image-classification, transformers":
```
sparseml.export --integration transformers --task model obcq_deployment
Usage: sparseml.export [OPTIONS] SOURCE_PATH
Try 'sparseml.export --help' for help.

Error: Invalid value for '--integration': 'transformers' is not 'image-classification, transformers'.
```

### DIFF
--- a/src/sparseml/export/export.py
+++ b/src/sparseml/export/export.py
@@ -395,7 +395,7 @@ def export(
 )
 @click.option(
     "--integration",
-    type=click.Choice(["image-classification, transformers"]),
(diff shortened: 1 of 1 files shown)


dbogunowicz:
@mgoin I already have this fix in https://github.com/neuralmagic/sparseml/pull/1991/files . You can close this PR

PR #2005, opened 2024-01-25
Add deprecation warning to OBCQ export script

dbogunowicz:


mgoin:
We have a lot of examples referencing that export script, not to mention the CLI command registered to it in setup.py - let's setup a plan to deprecate with @mwitiderrick

Satrat (approved):

mgoin (approved):

PR #2011, opened 2024-01-25
Introduce a wrapper to create `SparseAutoTokenizer.from_pretrained`

dbogunowicz:
Adding `SparseAutoTokenizer` to augment `transformers.AutoTokenizer`

- [x] Replace existing `AutoTokenizer.from_pretrained` paths with the new `SparseAutoTokenizer`


src/sparseml/transformers/utils/sparse_tokenizer.py
```diff
@@ -0,0 +1,54 @@
...
+        :param pretrained_model_name_or_path: the name of or path to the model to load
+        :return tokenizer: the loaded tokenizer from pretrained
+        """
+        if pretrained_model_name_or_path.startswith("zoo:"):
+            # if SparseZoo stub is passed, fetch the deployment
+            # path of the SparseZoo model and replace
+            # pretrained_model_name_or_path with the deployment path
+            pretrained_model_name_or_path = Model(
+                pretrained_model_name_or_path
+            ).deployment.path
```
rahul-tuli, line 43: Won't deployment have onnx instead of PyTorch file?

dbogunowicz: We don't care about .bin or .pth files. All we care about is that the appropriate tokenizer files are present. And they should be - we have appropriate checks and logic for moving those files from the training to deployment folder on export.

src/sparseml/transformers/utils/sparse_tokenizer.py
```diff
@@ -0,0 +1,54 @@
...
+
+        elif pretrained_model_name_or_path.startswith("hf:"):
```
mgoin, line 45: currently we aren't using the "hf:" style to specify HF model id, most places are just passing the string as you would to `transformers` if it isn't a local file - could we continue to do this since we are trying to keep UX parity with `transformers`

dbogunowicz: This should not be a problem. All clear.

src/sparseml/transformers/__init__.py
```diff
@@ -55,3 +55,4 @@ def _check_transformers_install():
 _check_transformers_install()
 
 from .export import *
+from .utils import *
```
mgoin, line 58: nice

mgoin (approved):

Satrat (approved):

src/sparseml/transformers/utils/sparse_tokenizer.py
```diff
@@ -0,0 +1,45 @@
...
+        :param pretrained_model_name_or_path: the name of or path to the model to load
+        :return tokenizer: the loaded tokenizer from pretrained
+        """
+        if pretrained_model_name_or_path.startswith("zoo:"):
+            # if SparseZoo stub is passed, fetch the deployment
+            # path of the SparseZoo model and replace
+            # pretrained_model_name_or_path with the deployment path
+            pretrained_model_name_or_path = Model(
+                pretrained_model_name_or_path
+            ).deployment.path
```
bfineran, line 43: let's download just the `tokenizer.json` and other relevant files if we can (iterate through sparsezoo deployment dir)

bfineran (changes requested):

src/sparseml/transformers/utils/sparse_tokenizer.py
mgoin: oh yes please don't download the model weights

src/sparseml/experimental/sparsegpt/utils.py
```diff
@@ -240,9 +240,9 @@ def get_wikitext2(nsamples, seed, seqlen, model):
     traindata = load_dataset("wikitext", "wikitext-2-raw-v1", split="train")
     testdata = load_dataset("wikitext", "wikitext-2-raw-v1", split="test")
 
-    from transformers import AutoTokenizer
+    from sparseml.transformers import SparseAutoTokenizer
```
Satrat, line 243: We shouldn't be messing with any of the code in sparseml/experimental, its research code

Satrat (dismissed):
Change request is small! Could you also clarify why we need a sparse tokenizer? How is this different from the transformers tokenizer?

dbogunowicz:
@Satrat the goal is to create a pathway, that could be generally used for creating and using sparse torch models (think about it in the context of the upcoming endeavors discussed internally). Example:
```
from sparseml.transformers import SparseAutoModelForCausalLM, SparseAutoTokenizer
model = SparseAutoModelForCausalLM.from_pretrained("STUB_OR_PATH", device_map="auto")
tokenizer = SparseAutoTokenizer("STUB_OR_PATH")
inputs = tokenizer(["PROMPT"], return_tensors="pt")
generated_ids = model.generate(**inputs)
outputs = tokenizer.batch_decode(generated_ids, skip_special_tokens=True)
```

mgoin:
To be clear, this tokenizer is only needed right now if you are using a model from SparseZoo right?

dbogunowicz:
@mgoin in practice yes

bfineran (approved):

PR #2020, opened 2024-01-29
[Fix] Remove removed models from `test_resnets`

dbogunowicz:


tests/sparseml/pytorch/models/classification/test_resnet.py
```diff
@@ -64,16 +64,12 @@
         ("resnet50", True, False, resnet50),
         ("resnet50", "base", False, resnet50),
         ("resnet50_2xwidth", False, True, resnet50_2xwidth),
```
dbogunowicz, line 66: Why non-pretrained `resnet50_2xwidth` (case:  ` ("resnet50_2xwidth", False, True, resnet50_2xwidth)`) does not fail tests? Is this model available, contrary to its pre-trained counterparts? @anmarques @tdg5 
Same thing applies to  case:`("resnet101_2xwidth", False, True, resnet101_2xwidth)`

bfineran (approved):

PR #2028, opened 2024-01-31
[Fix] Fix type hinting for OrderedDict (python 3.8)

dbogunowicz:


Satrat (approved):

rahul-tuli (approved):
Lol. We ran into this error because we also import `OrderedDict` from collections?

dbarbuzzi:
> We ran into this error because we also import `OrderedDict` from collections?

More specifically, because in python3.8, the `OrderedDict` from `collections` cannot be subscripted for use as type-hinting (it could _probably_ be used without subscripting).

PR #2030, opened 2024-01-31
Distillation Updates

Satrat:
Updating distillation modifier with llama-recipes functionality, so we can reproduce research results. In the initial SparseML implementation we would pass the student output from the previous layer to the next teacher layer. Instead we need to pass the initial input through each of the student and teacher models independently.

### Main Change
Rather than just wrapping the modules we want to distill, we need to wrap the whole model and create buffers to store the intermediate output of each distillation layer. The new wrapper is defined in `model_wrapper.py`. Note that the intermediate outputs are stored on the cpu

### Summary of Smaller Changes
* Support for FSDP teacher
* updating kd_wrapper to only compare first output (hidden state), previously was comparing cache as well
* default distillation transform to identity (no transform)
* Fix loss propogation to trainer
* alter distill_scale to scale by the number of layers in the case of square_head

### Testing 
Validated train loss and winogrande accuracy metrics against llama-recipes implementation on single GPU tinyllama1.1B. Ran with multi-gpu as well but not a full test run
* llama-recipe: 0.5872 winogrande acc, 1.28 training loss
* sparseML: 0.5848 winogrande acc, 1.17 training loss (total loss of 1.77 - distill loss 0.6)

**TODO:** Run multi-gpu llama7b test


bfineran (dismissed):

bfineran (approved):

bfineran (approved):

bfineran (approved):

bfineran:
failing with unrelated oom issues, merging

PR #2077, opened 2024-02-14
Fix Export Recipe Bug

Satrat:
When multiple exports that require recipe structure applications are performed in the same python session, the recipes get stacked in the SparseSession and are applied multiple times. 

The fix is to create a new session at the start of export, then clear it afterwards, so we aren't reusing recipes

Asana ticket: https://app.asana.com/0/1206109050183159/1206584891540686/f

### Testing
This example would previously fail on the second export:
```
from sparseml import export

for i in range(3):
    output_dir = "./test_export_" + str(i)
    export("zoo:codellama-7b-evolcodealpaca_codellama_pretrain-pruned60_quantized", task="text-generation", integration='transformers', target_path=output_dir)
```

Now it successfully completes all the exports

### DIFF
--- a/src/sparseml/export/export.py
+++ b/src/sparseml/export/export.py
@@ -70,6 +70,7 @@ from typing import Any, List, Optional, Union
 import numpy
 
 import click
+import sparseml.core.session as session_manager
 from sparseml.export.helpers import (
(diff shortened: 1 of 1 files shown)


bfineran (approved):
Nice catch

PR #2085, opened 2024-02-15
FakeQuantize Wrapper for Multi-GPU support

Satrat:
Creates a wrapper around the `FakeQuantize` forward pass to address two shortcomings of the pytorch implementation
* **no bfloat16 support,** so we cast to float32 before the forward pass and back to bfloat16 afterwards
* **no device_map=auto support,** so we move the scale, zero_point and observer min/max's to the correct device before the forward pass, and move them back afterwards

### Testing
see `test_fake_quant_wrapper` in `test_obcq.py` This example would previously fail to run due to the device map and model precision

### Profiling
Ran OBCQ with SmoothQuant to 50% sparsity on `TinyLlama/TinyLlama-1.1B-intermediate-step-1431k-3T. Sequence length 512 and 128 calibration samples with batch size 1
* bfloat16, device_map=auto: 3min 0.577581sec
* float32, device_map=auto:  3min 36.697597sec
* float32, device_map=cuda:0: 3min 48.144411sec

No difference in runtime running with float32 and device_map=cuda:0 on this branch compared to main, so the addition of `FakeQuantizeWrapper` has very minimal overhead


src/sparseml/modifiers/quantization/utils/fake_quant_wrapper.py
```diff
@@ -0,0 +1,77 @@
...
+        Overrides the forward pass by converting the weight to a compatible dtype and
+        moving configuration parameters to the device of the weight.
+
+        All changes are reverted before returning.
+        """
+        if self.fake_quant_enabled[0] == 0:
+            return super().forward(X)
+
+        # this assumes all params are on the same device
+        og_params_device = self.zero_point.device
```
rahul-tuli, line 39: love the name!

rahul-tuli (dismissed):

bfineran (approved):

rahul-tuli (approved):

mgoin (dismissed):
works for my smoke test, thanks a lot!

bfineran (approved):

PR #2091, opened 2024-02-16
Allow datasets smaller than calibration default

Satrat:
If the calibration dataset is smaller than the requested `num_calibration_samples` running oneshot would fail. With this PR, we instead throw a warning and proceed to calibrate with the full small dataset

### Testing 
This would previously fail, now throws a warning and runs with a truncated number of calibration samples
```
from sparseml.transformers import oneshot 

recipe_str = "tests/sparseml/transformers/obcq/test_tiny2.yaml"
model = "Xenova/llama2.c-stories15M"
device = "cuda:0"
dataset_name = "open_platypus"
concatenate_data = False
num_calibration_samples = 500
output_dir = "oneshot_out"
splits = {"calibration": "train[:1%]"}

oneshot(
    model_name_or_path=model,
    dataset_name=dataset_name,
    output_dir=output_dir,
    num_calibration_samples=num_calibration_samples,
    recipe=recipe_str,
    concatenate_data=concatenate_data,
    splits=splits,
    oneshot_device=device,
)
```

### DIFF
--- a/src/sparseml/transformers/finetune/data/data_helpers.py
+++ b/src/sparseml/transformers/finetune/data/data_helpers.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
 import os
 from typing import Any, Callable, Dict, List, Optional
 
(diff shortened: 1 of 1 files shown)


mgoin (approved):
Excellent, thank you for the improvement!

bfineran (approved):

PR #2092, opened 2024-02-16
Reset SparseSession Between Runs

Satrat:
When running sparseML commands such as `sparseml.transformers.apply/train/oneshot` in a Jupyter notebook, running multiple commands in a row would fail because the SparseSession object was persisting between runs.

To fix this we call `session_manager.create_session()` at the beginning of each run to clear data from the previous run. Also made this a cmd line arg in case the user wants the force the session to persist

### Testing 

Unit test `test_oneshot_then_finetune` previously required a session reset between sections, now that this is handled internally its no longer required.


bfineran (approved):

mgoin (approved):
this helps a lot, thanks!

PR #2095, opened 2024-02-19
Goin custom bug

horheynm:
Follow up to https://github.com/neuralmagic/sparseml/pull/2093

Automatically get rid of unwanted columns if no `remove_columns` were given

### DIFF
--- a/src/sparseml/transformers/finetune/data/custom.py
+++ b/src/sparseml/transformers/finetune/data/custom.py
@@ -47,12 +47,20 @@ class CustomDataset(TextGenerationDataset):
 
         raw_dataset: DatasetDict = super().get_raw_dataset()
 
+        if self.remove_columns is not None:
+            raw_dataset = self.map(
+                raw_dataset,
+                batched=False,
+                remove_columns=self.remove_columns,
+                num_proc=self.data_args.preprocessing_num_workers,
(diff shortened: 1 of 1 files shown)


src/sparseml/transformers/finetune/data/custom.py
```diff
@@ -66,3 +71,14 @@ def get_raw_dataset(self, *_ignore, **__ignore) -> DatasetDict:
             )
 
         return raw_dataset
+
+    def get_remove_columns_from_dataset(self, dataset) -> List[str]:
```
mgoin, line 75: Can you put a type for dataset?

Satrat (approved):

src/sparseml/transformers/finetune/data/custom.py
```diff
@@ -66,3 +71,14 @@ def get_raw_dataset(self, *_ignore, **__ignore) -> DatasetDict:
...
+        """Remove redandant columns from the dataset for processing"""
+        remove_columns = set()
+        for key in dataset.keys():
```
mgoin, line 78: Nit: you could use dataset.values() here since you don't use key for anything else

mgoin (approved):

Satrat (approved):

PR #2115, opened 2024-02-23
Custom datasets

horheynm:
## Summary
Bug fix for running custom dataset using accerate 
and use preprocessing func in module `src/sparseml/transformers/utils/preprocesing_funtions.py`, specified by string arg in parser.

```bash
#!/bin/bash
export CUDA_VISIBLE_DEVICES="0,1,2,3"
NPROC=$(($(echo $CUDA_VISIBLE_DEVICES | grep -o "," | wc -l)+1))
NPROC=1
ROOT=/home/george/sparseml/
DATASET=json
DATASET_PATH=/network/abhinav/datasets/evolcodealpaca
DATASET_CONFIG_NAME="default"
WORKERS=16
MAX_LEN=1024
SRC_MODEL_NAME=/network/abhinav/llama/models/Llama-2-7b-hf
SRC_MODEL=/network/abhinav/llama/models/Llama-2-7b-hf
RECIPE_DIR=/network/george/recipes #$ROOT/recipes/llama2_code
TEXT_COLUMN="text"

FSDP_CONFIG=$RECIPE_DIR/fsdp.yaml

LR=5e-5
WARM=0.1
EPOCHS=2
ID=$RANDOM
BATCH=16
GRAD_ACC=1
DST_MODEL_DIR=$HOME/models/llama2_coding/dense_finetuned
DST_MODEL_NAME=$SRC_MODEL_NAME@$DATASET@LR$LR@WARM$WARM@BS$BATCH$GrA$GRAD_ACC@EP$EPOCHS@ID$ID
DST_MODEL=$DST_MODEL_DIR/$DST_MODEL_NAME

PREPROCESSING_FUNC="evolved_codealpaca_dataset"


accelerate launch \
  --config_file $FSDP_CONFIG \
  --no_python sparseml.transformers.text_generation.train \
  --fsdp full_shard \
  --fsdp_config $RECIPE_DIR/extra_fsdp.json \
  --model $SRC_MODEL \
  --dataset $DATASET \
  --dataset_path $DATASET_PATH \
  --dataset_config_name $DATASET_CONFIG_NAME \
  --preprocessing_num_workers $WORKERS \
  --max_seq_length $MAX_LEN \
  --learning_rate $LR \
  --warmup_ratio $WARM \
  --per_device_train_batch_size $BATCH \
  --gradient_accumulation_steps $GRAD_ACC \
  --output_dir $DST_MODEL \
  --num_train_epochs $EPOCHS \
  --report_to wandb \
  --run_name $DST_MODEL_NAME \
  --precision bfloat16 \
  --bf16 True \
  --text_column $TEXT_COLUMN \
  --preprocessing_func $PREPROCESSING_FUNC

```


src/sparseml/transformers/finetune/data/data_helpers.py
```diff
@@ -215,4 +217,27 @@ def get_custom_datasets_from_path(path: str, ext: str = "json") -> Dict[str, str
                 if dir_dataset:
                     data_files[dir_name] = dir_dataset
 
+    return transform_dataset_keys(data_files)
+
+
+def transform_dataset_keys(data_files: Dict[str, Any]):
+    """Transform dict keys to `train`, `val` or `test`"""
```
Satrat, line 224: nit: clarify in docstrings that we only transform the keys if there is a match

src/sparseml/transformers/utils/preprocesing_funtions.py
```diff
@@ -0,0 +1,20 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+def evolved_codealpaca_dataset(data):
+    PROMPT_DICT = """[Instructions]:\n{instruction}\n\n[Response]:"""
+    data["text"] = PROMPT_DICT.format_map(data)
+
+    return data
```
Satrat, line 20: Is this preprocessing function only applicable to this one dataset? Why do we need to create this separate feature instead of just adding the dataset to the TextGenerationDataset registry so we can call it as a string?

src/sparseml/transformers/finetune/data/custom.py
```diff
@@ -12,13 +12,19 @@
...
 
+from sparseml.base import import_attr_from_path
 from sparseml.transformers.finetune.data import TextGenerationDataset
 
 
+PREPROCESSING_FUNCTION_REGISTRY_PATH = (
+    "sparseml.transformers.utils.preprocesing_funtions"
+)
+
+
```
Satrat, line 27: Instead of storing the registry as a constant variable, why not implement it as a RegistryMixIn? Then we can tag each helper function to the registry and avoid all that string parsing of the path

Satrat (commented):
I like the idea of a preprocessing function registry, but I'm not sure I understand the need for it in this case. It seems like this preprocessing function is for a specific dataset, is there a reason to not just add that dataset to the existing TextGenerationDataset registry and put the preprocessing function there?

EDIT: after talking with Abhinav, he is using a custom split of the evolcodealpaca dataset, so going through the dataset in our registry will not work. I think we should still refactor the preprocessing registry into a real RegistryMixIn, but I understand the need for it now

src/sparseml/transformers/utils/preprocesing_funtions.py
Satrat: EDIT: after talking with Abhinav, he is using a custom split of the evolcodealpaca dataset, so going through the dataset in our registry will not work. I think we should still refactor the preprocessing registry into a real RegistryMixIn, but I understand the need for it now

src/sparseml/transformers/finetune/data/custom.py
horheynm: Ok sounds good. RegistryMixIn for prepreprocessing func is created!

Satrat (dismissed):
Nice, this is a lot cleaner with the registry!

src/sparseml/transformers/finetune/data/custom.py
```diff
@@ -55,9 +58,16 @@ def get_raw_dataset(self, *_ignore, **__ignore) -> Union[DatasetDict, Dataset]:
             raw_dataset = super().get_raw_dataset()
 
         if self.preprocessing_func is not None:
+            func = (
+                self.preprocessing_func
+                if isinstance(self.preprocessing_func, Callable)
+                else PreproecessingFunctionRegistry.get_value_from_registry(
```
abhinavnmagic, line 64: Typo: PreprocessingFunctionRegistry

src/sparseml/transformers/utils/__init__.py
```diff
@@ -20,6 +20,7 @@
 from .helpers import *
 from .load_task_dataset import *
 from .metrics import *
+from .preprocesing_funtions import *
```
abhinavnmagic, line 23: Type: preprocessing_functions

abhinavnmagic (approved):

src/sparseml/transformers/finetune/data/custom.py
```diff
@@ -55,9 +58,16 @@ def get_raw_dataset(self, *_ignore, **__ignore) -> Union[DatasetDict, Dataset]:
             raw_dataset = super().get_raw_dataset()
 
         if self.preprocessing_func is not None:
+            func = (
+                self.preprocessing_func
+                if isinstance(self.preprocessing_func, Callable)
+                else PreprocessingFunctionRegistry.get_value_from_registry(
```
bfineran, line 64: besides registry, let's also enable support for a custom pathway like `"path/to/file.py:function_name"`

see https://github.com/neuralmagic/deepsparse/blob/4e791ee9843f87dd35a3c026bb7b63f5e20d6aef/src/deepsparse/legacy/loggers/build_logger.py#L69

src/sparseml/transformers/finetune/data/custom.py
```diff
@@ -55,9 +59,22 @@ def get_raw_dataset(self, *_ignore, **__ignore) -> Union[DatasetDict, Dataset]:
             raw_dataset = super().get_raw_dataset()
 
         if self.preprocessing_func is not None:
+
+            if isinstance(self.preprocessing_func, Callable):
```
rahul-tuli, line 63: nit: 
```suggestion
            if callable(self.preprocessing_func):
```

src/sparseml/transformers/finetune/data/custom.py
```diff
@@ -55,9 +59,22 @@ def get_raw_dataset(self, *_ignore, **__ignore) -> Union[DatasetDict, Dataset]:
...
+                func = self.preprocessing_func
+            else:
+                if ":" in self.preprocessing_func:
```
rahul-tuli, line 66: nit: Why not combine these to reduce indentation?

src/sparseml/transformers/finetune/data/data_args.py
```diff
@@ -55,8 +55,15 @@ class CustomDataTrainingArguments(DVCDatasetTrainingArguments):
         metadata={"help": "Column names to remove after preprocessing custom datasets"},
     )
 
-    preprocessing_func: Optional[Callable] = field(
-        default=None, metadata={"help": "The preprcessing function to apply"}
+    preprocessing_func: Union[None, str, Callable] = field(
+        default=None,
+        metadata={
+            "help": (
+                "The preprcessing function to apply ",
```
rahul-tuli, line 62:
```suggestion
                "The preprocessing function to apply ",
```

src/sparseml/transformers/finetune/data/custom.py
```diff
@@ -55,9 +59,22 @@ def get_raw_dataset(self, *_ignore, **__ignore) -> Union[DatasetDict, Dataset]:
...
+                    # load func_name from "/path/to/file.py:func_name"
+                    func = import_from_path(self.preprocessing_func)
```
rahul-tuli, line 68: https://github.com/neuralmagic/sparsezoo/blob/eabb28ca0b8d6a14089f5967c14457237daf04d1/src/sparsezoo/utils/registry.py#L258 Don't we already have support of this in the above line? If that would would we could combine the two if conditions maybe?

src/sparseml/transformers/finetune/data/data_helpers.py
```diff
@@ -218,4 +220,32 @@ def get_custom_datasets_from_path(path: str, ext: str = "json") -> Dict[str, str
                 if dir_dataset:
                     data_files[dir_name] = dir_dataset
 
+    return transform_dataset_keys(data_files)
+
+
+def transform_dataset_keys(data_files: Dict[str, Any]):
+    """
+    Transform dict keys to `train`, `val` or `test` for the given input dict
+    if matches exist with the existing keys
```
rahul-tuli, line 229: So if data-files contains keys : `train_sft, train_something_else` transformation would not take place?

but if we only have one of the two it would? Could we add that in the docstring?

src/sparseml/transformers/finetune/data/data_helpers.py
```diff
@@ -218,4 +220,32 @@ def get_custom_datasets_from_path(path: str, ext: str = "json") -> Dict[str, str
...
+    def transform_dataset_key(candidate: str):
+        for key in keys:
+            if candidate in key:
+                if key == candidate:
+                    return
+                val = data_files.pop(key)
+                data_files[candidate] = val
+
+    def do_transform(candidate: str):
+        return sum(candidate in key for key in keys) == 1
```
rahul-tuli, line 244: nit: Why not move these outside and just accept arguments instead of using closures? Won't that make them more testable?

src/sparseml/transformers/utils/preprocessing_functions.py
```diff
@@ -0,0 +1,27 @@
...
+
+from sparsezoo.utils.registry import RegistryMixin
+
+
+class PreprocessingFunctionRegistry(RegistryMixin):
+    ...
+
+
+@PreprocessingFunctionRegistry.register()
+def custom_evolved_codealpaca_dataset(data):
```
rahul-tuli, line 23: nice!

rahul-tuli, line 23: could we add typing for data? it's not clear what the function expects?

src/sparseml/transformers/finetune/data/data_helpers.py
```diff
@@ -218,4 +220,32 @@ def get_custom_datasets_from_path(path: str, ext: str = "json") -> Dict[str, str
...
+
+    def transform_dataset_key(candidate: str):
+        for key in keys:
+            if candidate in key:
+                if key == candidate:
+                    return
+                val = data_files.pop(key)
+                data_files[candidate] = val
+
+    def do_transform(candidate: str):
```
rahul-tuli, line 243: return type maybe?

rahul-tuli:
Do we know why the integration tests are failing though?

horheynm:
> Do we know why the integration tests are failing though?

Yeah OOM in GHA machine

Satrat (approved):

PR #2131, opened 2024-02-27
[Export API Refactor][Fix] Varying sample and model input names on correctness validation

dbogunowicz:


src/sparseml/export/validators.py
```diff
@@ -17,9 +17,12 @@
 import os.path
 from collections import OrderedDict
 from pathlib import Path
-from typing import Callable, List, Optional, Union
+from typing import Callable, List, Optional
+from typing import OrderedDict as OrderedDictType
+from typing import Union
```
dbogunowicz, line 22:
```suggestion
```

src/sparseml/export/validators.py
```diff
@@ -17,9 +17,12 @@
 import os.path
 from collections import OrderedDict
 from pathlib import Path
-from typing import Callable, List, Optional, Union
+from typing import Callable, List, Optional
```
dbogunowicz, line 20:
```suggestion
from typing import Callable, List, Optional, Union
```

bfineran (approved):

PR #2153, opened 2024-03-04
Add workflows for dev, nightly, and release Docker containers

dsikka:
# Summary
- Add a workflow which based on the conditions, will create a nightly, dev or release container
- So far, the workflows rely on a cuda-based Python 3.8 docker image from `nm-docker` https://github.com/neuralmagic/nm-docker/pull/1. This source image is a build arg and can be updated to be whatever python version is required by the workflow
- The following cases are covered:

## 1. Dev
- For dev, a container is created if a PR to main is opened or updated
- sparseml will be cloned using the specific branch and an editable install will be used to install all packages
- The container will be tagged with the PR number when pushed to GCR

## 2. Release
- When a PR is opened/updated to a release branch: in that case, sparseml will be installed using the specific version of the release

## 3. Nightly
- When a PR is merged into main, sparseml will install the latest nightly build
- A scheduled nightly cron job will trigger a nightly build and create a container with the most recent nightly build

# Testing
- Nightly and dev workflows have been tested and containers are currently being pushed to packages
- Release builds have not been tested 

# Next Steps:
- Update the names of the containers in GCR. So far using dummy test names but will need slight updating to consider the different workflows
- Find a non-destructive way to test release workflows


docker/containers/docker_dev/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+RUN python3.8 -m pip install --upgrade pip \
+    && python3.8 -m pip install --upgrade setuptools
+
+ARG CUDA
+ARG TORCH_VERSION
+ARG TORCHVISION_VERSION
+
+RUN python3.8 -m pip install torch==${TORCH_VERSION}+cu${CUDA} torchvision==${TORCHVISION_VERSION}+cu${CUDA} -f https://download.pytorch.org/whl/torch_stable.html \
+    && git clone https://github.com/neuralmagic/sparseml.git --depth 1 --single-branch -b ${BRANCH} \
```
horheynm, line 20: If i make edits to the code, can i push the change using http git clone here?

docker/containers/docker_nightly/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+RUN python3.8 -m pip install --upgrade pip \
+    && python3.8 -m pip install --upgrade setuptools
+
+ARG CUDA 
+ARG TORCH_VERSION
+ARG TORCHVISION_VERSION
+
+RUN python3.8 -m pip install torch==${TORCH_VERSION}+cu${CUDA} torchvision==${TORCHVISION_VERSION}+cu${CUDA} -f https://download.pytorch.org/whl/torch_stable.html \
+    && python3.8 -m pip install --no-cache-dir "sparseml-nightly[onnxruntime,torchvision,transformers,yolov5,ultralytics]"
```
horheynm, line 17: Are pushes to nightly blocked on previous nightly passing?

.github/workflows/build-container.yml
```diff
@@ -0,0 +1,78 @@
...
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build Dev Docker Container
+        if: ${{ env.DEV == 'true' }}
+        uses: docker/build-push-action@v4
+        with: 
+          context: ./docker/containers/docker_dev
+          build-args: |
+            BRANCH=${{github.head_ref}}
+          push: true
+          tags: ghcr.io/neuralmagic/test-sparseml-dev:${{ env.NAME }}
```
bfineran, line 62: can we add a todo to remove the `test-` if that's intended?

docker/containers/docker_dev/Dockerfile
```diff
@@ -0,0 +1,25 @@
+ARG SOURCE=ghcr.io/neuralmagic/cuda-python3.8
+# how were these versions selected before?
```
bfineran, line 2: likely latest supported at time of creation - or highest supported that worked out of the box with the CUDA image

docker/containers/docker_dev/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+ARG TORCH_VERSION=1.10.2
+ARG TORCHVISION_VERSION=0.11.3
```
bfineran, line 4: should bump these up if we can

bfineran, line 4: do you think we should be installing torchvision by default now? I don't think it's too heavy but worth weighing with the LLM focus

.github/workflows/build-container.yml
dsikka: yup. placeholders until we decide on names

docker/containers/docker_dev/Dockerfile
dsikka: wdym?

docker/containers/docker_dev/Dockerfile
dsikka: I think potentially start with and move as needed?

docker/containers/docker_nightly/Dockerfile
dsikka: there will be one follow-up PR such that containers aren't built if the wheels they need fail to build. So the dependency is on the wheel building.

docker/containers/docker_dev/Dockerfile
horheynm: Sorry for the late response. The question is for pushing code if that is intended - ssh need ssh keys, http needs username + password /token. 
SparseML is public, so i dont think ssh/http will be an issue actually.

bfineran (approved):

horheynm (approved):

PR #2154, opened 2024-03-05
Retire `flaky` in favour of `pytest-rerunfailures`

dbogunowicz:
Remove the dependency on `flake` (which is not actively maintained) in favour of `pytest` plugin. Also, allow for the newest `pytest` version


mgoin (approved):
Lgtm

bfineran (approved):

PR #2156, opened 2024-03-05
Improved Default Targets for SparseGPT

Satrat:
Previously if the `targets` field wasn't filled out, we would default to `ALL_TOKEN`. This default isn't functional because nested modules get run multiple times. A better default is to target all of the decoder layers. The path to these layers varies between models, but we can infer it utilizing `PretrainedModel._no_split_modules`

### Testing
```
from sparseml.core import Framework,ModifiableModel
from sparseml.modifiers.obcq.pytorch import SparseGPTModifierPyTorch

model = SparseAutoModelForCausalLM.from_pretrained("Xenova/llama2.c-stories15M")
modifiable_model = ModifiableModel(framework=Framework.pytorch, model=model)
targets = modifiable_model.get_no_split_params()
print(targets)

modifier = SparseGPTModifierPyTorch(sparsity=0.5)
modifier.targets = targets
modifier.model = modifiable_model
compressible_layers = modifier.compressible_layers()
print(compressible_layers.keys())
```

```
['LlamaDecoderLayer']
dict_keys(['model.layers.0', 'model.layers.1', 'model.layers.2', 'model.layers.3', 'model.layers.4', 'model.layers.5'])
```

### DIFF
--- a/src/sparseml/utils/pytorch/module.py
+++ b/src/sparseml/utils/pytorch/module.py
@@ -98,6 +98,17 @@ def match_targets(name: str, targets: Union[str, List[str]]) -> Tuple[bool, int]
     return False, -1
 
 
+def match_class(layer: Module, targets: Union[str, List[str]]) -> Tuple[bool, int]:
+    if isinstance(targets, str):
+        targets = [targets]
+
+    for index, target in enumerate(targets):
+        if layer.__class__.__name__ == target:
+            return True, index
+
+    return False, -1
+
+
 def get_default_params(layers: Dict[str, Module]) -> Dict[str, Parameter]:
     params = {}
     for name, layer in layers.items():
@@ -139,6 +150,11 @@ def match_layers_params(
         if match and not params:
             targets_found[match_index] = True
             resolved[name] = layer
+        else:
+            match, match_index = match_class(layer, targets)
+            if match:
+                targets_found[match_index] = True
+                resolved[name] = layer
 
         for param_name, param in layer.named_parameters():
             if "." in param_name:  # skip parameters of nested layers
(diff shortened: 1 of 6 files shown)


src/sparseml/utils/pytorch/module.py
```diff
@@ -96,6 +97,17 @@ def match_targets(name: str, targets: Union[str, List[str]]) -> Tuple[bool, int]
     return False, -1
 
 
+def match_class(layer: Module, targets: Union[str, List[str]]) -> Tuple[bool, int]:
+    if isinstance(targets, str):
+        targets = [targets]
+
+    for index, target in enumerate(targets):
+        if layer.__class__.__name__ == target:
+            return True, index
```
bfineran, line 106: why do we need a bool and index if returning -1 in the non-match case?

Satrat: I was just mimicking the style of the match_targets function above, it makes the if statement a bit cleaner on line 137 but I'm not opposed to using just the index

mgoin:
Amazing approach! I just tested this with Rahul's latest modifier application for `model_id = "Xenova/llama2.c-stories15M"` and `model_id = "facebook/opt-125m"` and it seemed to work perfectly. Thanks!

```python
from sparseml.transformers import (
    SparseAutoModelForCausalLM, SparseAutoTokenizer, oneshot
)
from sparseml.modifiers import SparseGPTModifier

model_id = "Xenova/llama2.c-stories15M"
model = SparseAutoModelForCausalLM.from_pretrained(model_id)
tokenizer = SparseAutoTokenizer.from_pretrained(model_id)
dataset = "open_platypus"

modifier = SparseGPTModifier(sparsity=0.5, sequential_update=True)

oneshot(
    model=model,
    tokenizer=tokenizer,
    dataset=dataset,
    recipe=modifier,
    output_dir="output",
)
```

rahul-tuli (approved):

bfineran (approved):

mgoin (approved):

robertgshaw2-neuralmagic:
nice!

PR #2177, opened 2024-03-12
SparseML Compression Pt 1: saving w/compression configs

Satrat:
This initial PR for safetensors compression sets up the `CompressionConfig` and `ModelCompressor` registries and implements bitmask compression on save. See the [corresponding internal docs PR](https://github.com/neuralmagic/internal-docs/pull/42) for design details

### To be implemented in follow-up PRs
* bitmask decompression
* saving/loading/inferring sparsity config from model
* SparseAutoModel load/save interface

### Example

```python
from sparseml.transformers.compression import BitmaskConfig, BitmaskCompressor
from safetensors import safe_open
import os

MODEL_PATH = "zoo:llama2-7b-gsm8k_llama2_pretrain-pruned50.oneshot"

model = SparseAutoModelForCausalLM.from_pretrained(MODEL_PATH)

sparsity_config = BitmaskConfig()
compressor = BitmaskCompressor(config=sparsity_config)

model_state_dict = model.state_dict()
sparse_state_dict = compressor.compress(model_state_dict)


model.save_pretrained(OUTPUT_PATH, safe_serialization=True, state_dict=sparse_state_dict)

safetensors_path = os.path.join(OUTPUT_PATH, "model-00001-of-00002.safetensors")
with safe_open(safetensors_path, framework="pt", device=0) as f:
    test_name = "model.layers.4.self_attn.k_proj.weight"
    bitmask = f.get_tensor(test_name + ".bitmask")
    shape = f.get_tensor(test_name + ".shape")
    values = f.get_tensor(test_name + ".compressed")
    row_offsets = f.get_tensor(test_name + ".row_offsets")
    print(f"bitmask: {bitmask}")
    print(f"shape: {shape}")
    print(f"values: {values}")
    print(f"row offsets: {row_offsets}")
```


src/sparseml/transformers/compression/config/sparse_bitmask.py
```diff
@@ -0,0 +1,36 @@
...
+    Configuration for storing a sparse model using
+    bitmask compression
+
+    :param global_sparsity: average sparsity of the entire model
+    :param sparsity_structure: structure of the sparsity, such as
+    "unstructured", "2:4", "8:16" etc
+    """
+
+    format: str = "sparse_bitmask"
+    global_sparsity: Optional[float] = None
```
dbogunowicz, line 35: How about defaulting to zero rather than None?

dbogunowicz (approved):
Couple of nitpicks. All in all great job, definitely well done on a bite-sized scope of this PR

setup.py
```diff
@@ -33,7 +33,7 @@
 _deps = [
     "setuptools<=59.5.0",
     "pyyaml>=5.0.0",
-    "numpy>=1.0.0",
+    "numpy>=1.17.0",
```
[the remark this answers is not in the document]
mgoin: this is required for the `bitorder` flag in [numpy.packbits](https://numpy.org/doc/stable/reference/generated/numpy.packbits.html)

bfineran (approved):

src/sparseml/transformers/compression/compressors/sparse_bitmask.py
```diff
@@ -0,0 +1,195 @@
...
+        :param model_state: state dict of uncompressed model
+        :return: compressed state dict
+        """
+        compressed_dict = {}
+        _LOGGER.info(
+            f"Compressing model with {len(model_state)} parameterized layers..."
+        )
+        for name, value in tqdm(model_state.items()):
+            bitmask_tensor = BitmaskTensor(value)
+            compressed_dict |= bitmask_tensor.dict(name_prefix=name)
```
mgoin, line 58: should we warn if there is a conflict/name-overlap?

src/sparseml/transformers/compression/compressors/sparse_bitmask.py
```diff
@@ -0,0 +1,195 @@
...
+    """
+    Owns compressions and decompression for a single bitmask compressed tensor.
+    Adapted from: https://github.com/mgoin/torch_bitmask/tree/main
+
+    :param tensor: Dense tensor to compress
+    """
+
+    def __init__(self, tensor: Tensor):
+        self.shape = tensor.shape
+        self.values, self.bitmasks, self.row_offsets = bitmask_compress(tensor.cpu())
```
mgoin, line 82: should we move these compressed tensors to the original device of tensor? currently they are always going to be on the cpu after this line, right?

mgoin (commented):
Nice job Sara, this looks good to me for saving. Just some comments on state dict conflicts and tensor devices

src/sparseml/transformers/compression/compressors/sparse_bitmask.py
Satrat: I think for compression this is desired. I'm assuming we want to keep the dense model in GPU memory, so rather than doubling the GPU memory we just offload to CPU. While its easy to load safetensor files tensor by tensor, I don't think there is a way to do this on save (unless I'm missing something)

Satrat: That being said though, I think it makes sense to at least save the dense tensor device so we can use it on decompression. Added that in to my latest commit

src/sparseml/utils/pytorch/utils.py
```diff
@@ -0,0 +1,43 @@
...
+
+import gc
+
+import torch
+
+
+__all__ = ["measure_cuda_memory"]
+
+
+class measure_cuda_memory:
```
mgoin, line 23: this seems unused?

Satrat: Yeah my bad, that was meant for the follow up PR. Just deleted

src/sparseml/transformers/compression/compressors/sparse_bitmask.py
mgoin: Okay that makes sense, thanks

src/sparseml/transformers/__init__.py
```diff
@@ -59,3 +59,4 @@ def _check_transformers_install():
 from .utils import *
 from .export import *
 from .finetune import *
+from .compression import *
```
dbogunowicz, line 62: perhaps sparseml.compression, but we can refactor it is a separate PR

dbogunowicz (approved):
🎅

dbogunowicz (approved):

bfineran (approved):

PR #2184, opened 2024-03-15
SparseML Compression Pt 2: Load compressed weights

Satrat:
This PR implements ModelCompressor.decompress(), which will decompress the weights in the safetensors file one by one. Also includes a bunch of helper functions for reading safetensors files and dealing with the compressed format. See the [corresponding internal docs PR](https://github.com/neuralmagic/internal-docs/pull/42) for design details

**Note:** https://github.com/neuralmagic/sparseml/pull/2177 needs to be merged first

### To be implemented in follow-up PR
* inferring sparsity config from model 
* SparseAutoModel save/load interface

### Example

Sample code for compressing a model with 50% sparsity(See PR https://github.com/neuralmagic/sparseml/pull/2177), then reloading the compressed weights as a dense model 
```python
from sparseml.transformers.compression import BitmaskConfig, BitmaskCompressor
from sparseml.utils.pytorch.utils import measure_cuda_memory
from tqdm import tqdm
import torch

MODEL_PATH = "zoo:llama2-7b-gsm8k_llama2_pretrain-pruned50.oneshot"

torch.cuda.set_device(0)
with measure_cuda_memory() as m:
    model = SparseAutoModelForCausalLM.from_pretrained(MODEL_PATH, device_map="cuda:0")
print(f"Load dense model peak GPU {m.overall_peak_memory / float(2**30):.4f} GB")

sparsity_config = BitmaskConfig()
compressor = BitmaskCompressor(config=sparsity_config)

# compresses the model using Bitmask compression
with measure_cuda_memory() as m:
    model_state_dict = model.state_dict()
    sparse_state_dict = compressor.compress(model_state_dict)

    # save the compressed model
    model.save_pretrained(
        OUTPUT_PATH, 
        safe_serialization=True, 
        state_dict=sparse_state_dict
    )

print(f"Save compressed model peak GPU {m.overall_peak_memory / float(2**30):.4f} GB")

# use the dense state dict to reload the model
torch.cuda.set_device(1)
with measure_cuda_memory() as m:
    model_again = SparseAutoModelForCausalLM.from_pretrained(
        OUTPUT_PATH, 
        device_map="cuda:1"
    )

    #returns iterator
    dense_state_dict = compressor.decompress(OUTPUT_PATH)
    for name, data in tqdm(dense_state_dict, desc="Decompressing model"):
        BitmaskCompressor.replace_layer(name, data, model_again)

print(f"Load compressed model peak GPU {m.overall_peak_memory / float(2**30):.4f} GB")
```

Load dense model peak GPU 25.2276 GB
Compressing model: 100%|████████████████████████████████████████████████████████████████████████████████████████| 291/291 [01:28<00:00,  3.29it/s]
Save compressed model peak GPU 25.2276 GB
Loading checkpoint shards: 100%|████████████████████████████████████████████████████████████████████████████████████| 2/2 [00:00<00:00,  6.27it/s]
Decompressing model: 291it [01:11,  4.08it/s]
Load compressed model peak GPU 25.7159 GB


dbogunowicz (approved):

src/sparseml/transformers/compression/compressors/sparse_bitmask.py
```diff
@@ -54,46 +61,77 @@ def compress(self, model_state: Dict[str, Tensor]) -> Dict[str, Tensor]:
...
         )
         for name, value in tqdm(model_state.items()):
-            bitmask_tensor = BitmaskTensor(value)
+            bitmask_tensor = BitmaskTensor.from_dense(value)
             compressed_dict |= bitmask_tensor.dict(name_prefix=name)
 
         return compressed_dict
 
-    def decompress(self, model_state: Dict[str, Tensor]) -> Dict[str, Tensor]:
+    def decompress(self, model_path: str) -> Dict[str, Tensor]:
```
dbogunowicz, line 69: could we have a variant of `decompress` method that yields weights keys and values? This would be much better and efficient fit for vLLM integration.

bfineran (dismissed):

src/sparseml/transformers/compression/README.md
```diff
@@ -47,21 +45,22 @@ model = SparseAutoModelForCausalLM.from_pretrained(MODEL_PATH)
...
-    row_offsets = f.get_tensor(test_name + ".row_offsets")
-    print(f"bitmask: {bitmask}")
-    print(f"shape: {shape}")
-    print(f"values: {values}")
-    print(f"row offsets: {row_offsets}")
+# use the dense state dict to reload the model
+model_again = model = SparseAutoModelForCausalLM.from_pretrained(
+    OUTPUT_PATH, 
+    state_dict=dense_state_dict
+)
```
mgoin, line 65: 1. Is `SparseAutoModelForCausalLM` required here? it seems like AutoModelForCausalLM would work as well since the state_dict has already been materialized
2. Will this temporarily double the memory usage since the weights are fully materialized in dense_state_dict and AutoModelForCausalLM will create an empty model before using state_dict?

mgoin (changes requested):
Could we add some memory usage tests? I want to make sure our peak memory during loading doesn't exceed the dense model memory as much as possible.

I have an example of how to measure current GPU memory usage in vLLM here https://github.com/vllm-project/vllm/pull/3120/files and we can measure peak by just removing the `torch.cuda.reset_peak_memory_stats(self.device)` before checking usage

src/sparseml/transformers/compression/compressors/sparse_bitmask.py
```diff
@@ -54,46 +61,77 @@ def compress(self, model_state: Dict[str, Tensor]) -> Dict[str, Tensor]:
...
-        Uncompresses a bitmask compressed state dict back to dense
+        Reads a bitmask compressed state dict located at model_path and decompresses it
+        back to a dense state dict. Weights are decompressed sequentially.
 
-        :param model_state: state dict of uncompressed model
+        :param model_path: path to compressed safetensors model
         :return: compressed state dict
         """
-        raise NotImplementedError()
+        weight_mappings = get_nested_weight_mappings(
```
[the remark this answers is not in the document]
Satrat: The decompression will fail if the expected keys aren't provided, we're assuming if decompress is called its because we've inferred this bitmask compression format from the config file (loading from the config file will be addressed in a follow up PR)

src/sparseml/transformers/compression/compressors/sparse_bitmask.py
Satrat: Yes we could definitely write a version of decompress that returns an iterator. I think thats best left as a followup though since the current implementation does work with vLLM

src/sparseml/transformers/compression/README.md
Satrat: 1. Yes AutoModelForCausalLM would work here as well
2. I didn't think passing in the state dict would create a second copy of the model, but will check this using the method you posted above

src/sparseml/transformers/compression/compressors/sparse_bitmask.py
Satrat: Going back on this actually @dbogunowicz, in order to reduce memory footprint I think it makes sense to just yield in this function. Pushing a change shortly

src/sparseml/transformers/compression/README.md
Satrat: @mgoin made some updates to the decompress function to return an iterator, so we can now load the model then insert the decompressed layers one by one. The updated example includes peak memory usage. I didn't add anything on memory usage into the unit tests because GHA runs on cpu only

bfineran (dismissed):

dbogunowicz (dismissed):

src/sparseml/transformers/compression/compressors/sparse_bitmask.py
```diff
@@ -64,45 +71,73 @@ def compress(self, model_state: Dict[str, Tensor]) -> Dict[str, Tensor]:
                         "be replaced."
                     )
             compressed_dict |= bitmask_dict
```
mgoin, line 73: this should be removed?

mgoin (commented):
LGTM thanks, just one line I think was missed

mgoin (approved):
thanks!

PR #2187, opened 2024-03-19
[Testing] Implementation of an updated testing framework

dsikka:
# Summary:

1. Update an existing unit test and an existing integration test, both for export such that they are using the `unittest.TestCase` base class and `parameterized` module
2. The integration tests are also updated to use config files to define the different combination of inputs that are meant to be tested for a particular integration test. Each config must include a `cadence` field, which dictates how frequently the test is run (set through an environment variables) and the `test_type` (sanity, regression, or smoke) which for now is a config variable but we can easily swap to just use decorators, as being done with the sample `smoke` test added
3. Introduce a base `CustomIntegrationTest` class which other integration tests can inherit and use to test custom scripts and custom test classes.
4. Add examples of decorators to mark `unit` , `custom`, and `integration` tests. Also introduce decorators to mark tests with certain package or compute environments requirements (e.g `requires_torch`, `requires_gpu`). These are described in a new `testing_utils.py` file

**For a summary of the changes, see `tests/docs/testing_framework.md`**

# Testing
- An example of a class which uses the `CustomIntegrationTest` class is the `TestGenerationExportIntegrationCustom` and it can be found in `test_generation_export.py`. This test class was tested using sample config/python files added under `tests/sparseml/export/transformers/generation_configs`
- An existing integration test was updated to use config files (as mentioned above) and was tested using the newly added
`tests/sparseml/export/transformers/generation_configs/tiny_stories.yaml`

# Using `unittest`
- The existing tests have been updated to use `unittest` package. We can definitely keep pytest for the existing tests if there are strong feelings although I do find it easier to read and understand and we can easily use the package `parameterized` to get the same functionality (or even expanded functionality for classes, compared to pytest fixtures).
- Custom testing should keep the unittest framework as it makes testing custom scripts/tests fairly easy/straightforward


tests/custom_test.py
```diff
@@ -0,0 +1,82 @@
...
+            self.skipTest("No custom class found. Testing test")
+        loader = unittest.TestLoader()
+        tests = loader.discover(self.custom_class_directory, pattern=config.script_path)
+        testRunner = unittest.runner.TextTestRunner()
+        output = testRunner.run(tests)
+        for out in output.errors:
+            raise Exception(output[-1])
+
+        for out in output.failures:
+            assert False
```
bfineran, line 81: can we include anything in the second half of this assert to get information on the failures?

tests/custom_test.py
```diff
@@ -0,0 +1,82 @@
...
+        assert True
```
bfineran, line 82: what is the purpose of this line?

tests/data.py
```diff
@@ -0,0 +1,31 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from dataclasses import dataclass
+from typing import List, Union
+
+
+# TODO: add enums
+# test type as decorators?
```
bfineran, line 20: definitely would like to add this, at least in a follow up

tests/docs/testing_framework.md
```diff
@@ -0,0 +1,97 @@
+# An Updated Testing Framework
```
bfineran, line 1: is this meant to be a migration doc, or long standing demonstrating the new framework? if the latter probably don't need as much framing around what changed

tests/sparseml/export/transformers/test_generation_export.py
```diff
@@ -0,0 +1,58 @@
...
+
+import pytest
+
+from huggingface_hub import snapshot_download
+from parameterized import parameterized_class
+from sparseml import export
+from tests.testing_utils import parse_params
+
+
+CONFIGS_DIRECTORY = "tests/sparseml/export/transformers/generation_configs"
```
bfineran, line 27: any reason we wouldn't want to use relative path here? hardcoding this would require us to always run from `sparseml/` not that this is not always the case though...

tests/docs/testing_framework.md
dsikka: More so a summary of the new framework and what we want to establish going forward in terms of testing structure.  Maybe makes more sense in internal docs?

tests/sparseml/export/transformers/test_generation_export.py
dsikka: yah I kept the path from `sparseml` as that is the directory uses to run all gha tests

tests/docs/testing_framework.md
```diff
@@ -0,0 +1,97 @@
...
+`cadence`: which dictates how often the test case runs (on commit, weekly, or nightly), 
+and the `script_path` which lists the name of the custom testing script or custom test
+class within the directory. 
+
+An example of an implementation of the `CustomIntegrationTest` can be found under
+`tests/examples`
+
+## Additional markers and decorators 
+
+- New markers are added in to mark tests as `unit`, `integration`, `smoke`, and `custom`
```
derekk-nm, line 61: I'm confused about the use of markers for what looks like a "test type", when there's also the need for a .yaml file that defines a `test_type`.  could these be conflicting?  are they used for different purposes?

tests/docs/testing_framework.md
```diff
@@ -0,0 +1,97 @@
...
+- An integration test should be written for every major/critical module 
+    - All arg combinations should be tested through a `smoke` check
+    (all may be overkill but ideally we're covering beyond the few important combinations)
+    - All **important** arg combinations should be covered through either a `sanity`
+    check or a `regression` check
+        - A small model should be tested through a `sanity` check  
+        - All other larger models should be tested through `regression` test types
+
+## End-to-end Testing Targets:
+- Tests cascading repositories (sparseml --> vLLM) but will become more prominent as are
```
derekk-nm, line 92: Perhaps add that there should be at least one use case scenario that emulates how a user is expected to use a particular feature being tested.

tests/testing_utils.py
```diff
@@ -0,0 +1,109 @@
...
+def is_torch_available():
+    try:
+        import torch  # noqa: F401
+
+        return True
+    except ImportError:
+        return False
+
+
+def is_gpu_avaialble():
```
derekk-nm, line 38: typo in the name.  should probably be `is_gpu_available`

tests/docs/testing_framework.md
dsikka: Not set in stone. We may just stick to the .yaml support or just use markers.

tests/docs/testing_framework.md
```diff
@@ -0,0 +1,98 @@
...
```
rahul-tuli, line 92:
```suggestion
- Tests cascading repositories (sparseml --> vLLM) but will become more prominent as our
```

rahul-tuli (commented):
I'm cool with using a mix of unit test and pytest, specially cause pytest can pick up unittest style tests. Good work on the documentation, really appreciate it.

rahul-tuli (approved):

rahul-tuli (approved):

bfineran (approved):

PR #2190, opened 2024-03-20
SparseML Compression Pt 3: SparseAutoModel interface & inferring params

Satrat:
This final PR for SparseML sparsity compression adds compression support to `save_pretrained` and `from_pretrained`. It also adds support for inferring global sparsity and sparsity structure params from the model. See the [corresponding internal docs PR](https://github.com/neuralmagic/internal-docs/pull/42) for design details, but there have been some minor changes that are reflected in [README.md](https://github.com/neuralmagic/sparseml/blob/a2c8ff20bd1099d3e335e0ccdee4f17bb5595f22/src/sparseml/transformers/compression/README.md)

### Callouts
* Overriding `model.save_pretrained()` is tricky, because the model is initialized from `SparseAutoModelForCausalLM.from_pretrained`, but the returned model is a child of `PreTrainedModel`(for instance `LlamaForCausalLM` inherits from `LlamaPreTrainedModel` inherits from `PreTrainedModel`) To override `PreTrainedModel.save_pretrained` we need to do a bit of "class instance surgery". See `compression_save.py` for implementation details
* Ended up not storing the sparsity config in the safetensors metadata, storing it in the HF config seemed sufficient and I felt it was going to complicate the code without much added gain
* How we're inferring the sparsity structure is definitely hacky. We determine it by looking at the applied modifiers and checking for a mask_structure attribute. I think this is sufficient for now, but open to other ideas here

### Example 
```python
from sparseml.utils.pytorch.utils import measure_cuda_memory
import torch

MODEL_PATH = "zoo:llama2-7b-open_platypus_orca_llama2_pretrain-pruned60"
RECIPE = "zoo:llama2-7b-open_platypus_orca_llama2_pretrain-pruned60"

torch.cuda.set_device(0)
with measure_cuda_memory() as m:
    model = SparseAutoModelForCausalLM.from_pretrained(MODEL_PATH, device_map="cuda:0")
print(f"Load dense model peak GPU {m.overall_peak_memory / float(2**30):.4f} GB")

sparsity_config = getattr(model,"sparsity_config", None)
print(f"Sparsity config before compression: {sparsity_config}")
with measure_cuda_memory() as m:
    model.save_compressed(OUTPUT_PATH)
print(f"Save compressed model peak GPU {m.overall_peak_memory / float(2**30):.4f} GB")

torch.cuda.set_device(1)
with measure_cuda_memory() as m:
    model_again = SparseAutoModelForCausalLM.from_pretrained(
        OUTPUT_PATH, device_map="cuda:1"
    )
print(f"Load compressed model peak GPU {m.overall_peak_memory / float(2**30):.4f} GB")
sparsity_config = getattr(model,"sparsity_config", None)
print(f"Sparsity config after compression: {sparsity_config}")

og_state_dict = model.state_dict()
reconstructed_state_dict = model_again.state_dict()
assert(len(og_state_dict) == len(reconstructed_state_dict))
for key in og_state_dict.keys():
    dense_tensor = og_state_dict[key]
    reconstructed_tensor = reconstructed_state_dict[key]
    assert torch.equal(dense_tensor.cpu(), reconstructed_tensor.cpu())
```

Loading checkpoint shards: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 6/6 [00:03<00:00,  1.55it/s]
Load dense model peak GPU 25.2276 GB
Sparsity config before compression: None
Compressing model: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 291/291 [01:22<00:00,  3.51it/s]
Save compressed model peak GPU 26.3272 GB
Loading checkpoint shards: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 2/2 [00:00<00:00,  7.87it/s]
Decompressing model: 291it [00:23, 12.31it/s]
Load compressed model peak GPU 25.7159 GB
Sparsity config after compression: format='sparse_bitmask' global_sparsity=57.66354988216862 sparsity_structure='unstructured'

### Testing
Unit tests are included for `from_pretrained`, `save_pretrained` and `save_compressed`. In addition I manually tested the following scenarios. See the README for instructions on turning on compression during finetuning/oneshot
* load dense -> apply SparseGPT -> save compressed -> reload compressed
* FSDP: load dense -> apply SparseGPT -> save compressed -> reload compressed
* load sparse -> sparse finetuning -> save compressed -> reload compressed
* FSDP: load sparse -> sparse finetuning -> save compressed -> reload compressed


src/sparseml/transformers/compression/README.md
```diff
@@ -25,58 +25,117 @@ needed for decompression in the compressed state_dict:
 {
     PARAM_NAME.compressed: compressed_tensor # 1d tensor
     PARAM_NAME.bitmask: value # 2d bitmask tensor (nrows x (ncols / 8))
-Satrat marked this conversation as resolved.
```
dbogunowicz, line 28: :P

src/sparseml/transformers/utils/sparse_model.py
```diff
@@ -13,35 +13,42 @@
...
+from sparseml.transformers.compression import CompressionConfig, ModelCompressor
 from sparseml.transformers.utils.helpers import resolve_recipe
 from sparseml.utils import download_zoo_training_dir
 from sparseml.utils.fsdp.context import main_process_first_context
 
 
-__all__ = ["SparseAutoModel", "SparseAutoModelForCausalLM", "get_shared_tokenizer_src"]
+_LOGGER = logging.getLogger(__name__)
 
+SPARSITY_CONFIG_NAME = "sparsity_config"
```
dbogunowicz, line 48: potentially move to helpers/utils, this is where all the const reside for now

dbogunowicz (approved):

src/sparseml/transformers/utils/sparse_model.py
Satrat: done!

mgoin (commented):
I want to still stand by the UX value of having `model.save_pretrained/save_compressed` since having those be static functions doesn't have a precident that I've seen. In fact I want to expand the example flow here to include `model.push_to_hub("my/model-id")` so we can ensure there is a nice end-to-end code example for pushing up compressed models

This still follows the trend with AutoGPTQ UX (although they don't exactly inherit from Transformers):
```python
model = AutoGPTQForCausalLM.from_pretrained(pretrained_model_dir, quantize_config)

model.quantize(examples)

model.save_quantized(quantized_model_dir)

model.push_to_hub(repo_id)
```

As a test I would like to see loading a compressed model from HF, both with SparseAutoModelForCausalLM as a success and with AutoModelForCausalLM as a failure (I'm curious what the error will look like and if there is anything we can do there)

mgoin (changes requested):
Forgot to request changes

src/sparseml/transformers/compression/README.md
```diff
@@ -30,52 +30,112 @@ needed for decompression in the compressed state_dict:
...
+sparsity config will still be included in `config.json` with format `dense_sparsity`
+
+```python
+from sparseml.transformers.utils import SparseAutoModelForCausalLM
+
+SparseAutoModelForCausalLM.save_pretrained(
+    model,
+    save_directory=output_dir
+)
+```
```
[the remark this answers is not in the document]
Satrat: Thats a good point on being expensive for large models, and the user can always explicitly pass in a DenseConfig() if they want to keep the model dense on disk while still reporting sparsity parameters. Pushing a commit now to default to no sparsity config by default and updating this example in the README

mgoin (commented):
I tried using this to compress an existing model on HF. I chose the TinyLlama here: https://huggingface.co/neuralmagic/TinyLlama-1.1B-Chat-v1.0-pruned2.4/tree/main

There are a few issues I noticed:
* It seems like float32 is always used to save, where we should preserve the existing dtype of the model
* `SparseAutoModelForCausalLM.save_compressed` seems to only produce the model weights and config file in the output directory, where we should be including other needed files like the tokenizer. This may be addressed by inheriting instead of using a static function

Code:
```python

MODEL_PATH = "neuralmagic/TinyLlama-1.1B-Chat-v1.0-pruned2.4"

model = SparseAutoModelForCausalLM.from_pretrained(MODEL_PATH, device_map="cuda:0")
SparseAutoModelForCausalLM.save_compressed(model, OUTPUT_PATH)
```

Output:
```
config.json: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 672/672 [00:00<00:00, 7.52MB/s]
pytorch_model.bin: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 2.20G/2.20G [01:51<00:00, 19.8MB/s]
/home/mgoin/venvs/nm/lib/python3.10/site-packages/torch/_utils.py:831: UserWarning: TypedStorage is deprecated. It will be removed in the future and UntypedStorage will be the only storage class. This should only matter to you if you are using storages directly.  To access UntypedStorage directly, use tensor.untyped_storage() instead of tensor.storage()
  return self.fget.__get__(instance, owner)()
generation_config.json: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 132/132 [00:00<00:00, 1.47MB/s]
2024-03-20 19:29:24 sparseml.transformers.utils.helpers INFO     model_path is a huggingface model id. Attempting to download recipe from https://huggingface.co/
2024-03-20 19:29:24 sparseml.transformers.utils.helpers INFO     Found recipe: recipe.yaml for model id: neuralmagic/TinyLlama-1.1B-Chat-v1.0-pruned2.4. Downloading...
recipe.yaml: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 237/237 [00:00<00:00, 2.64MB/s]
Logging all SparseML modifier-level logs to sparse_logs/20-03-2024_19.29.24.log
2024-03-20 19:29:24 sparseml.core.logger.logger INFO     Logging all SparseML modifier-level logs to sparse_logs/20-03-2024_19.29.24.log
2024-03-20 19:29:24 sparseml.core.recipe.recipe INFO     Loading recipe from file /home/mgoin/.cache/huggingface/hub/models--neuralmagic--TinyLlama-1.1B-Chat-v1.0-pruned2.4/snapshots/22ff818572f6fb2bd02110dd0b40c0169533c6da/recipe.yaml
manager stage: Model structure initialized
2024-03-20 19:29:24 sparseml.pytorch.model_load.helpers INFO     Applied an unstaged recipe to the model at neuralmagic/TinyLlama-1.1B-Chat-v1.0-pruned2.4
2024-03-20 19:29:24 sparseml.pytorch.model_load.helpers WARNING  Model state was not reloaded for SparseML: could not find model weights for neuralmagic/TinyLlama-1.1B-Chat-v1.0-pruned2.4
Compressing model: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 201/201 [00:04<00:00, 43.20it/s]
```

Looking at the output of the save_compressed shows a larger file size, but I think this is because it is saving as float32, even though the model is originally float16:
```
ll test_compress_output 
total 2.5G
-rw-r--r-- 1 mgoin mgoin  906 Mar 20 19:29 config.json
-rw-r--r-- 1 mgoin mgoin  124 Mar 20 19:29 generation_config.json
-rw-r--r-- 1 mgoin mgoin 2.5G Mar 20 19:29 model.safetensors
```

Here is the section of the config.json in that directory that mentions the sparsity level and dtype:
```
    "sparsity_config": {
        "format": "sparse_bitmask",
        "global_sparsity": 44.0587486922757,
        "sparsity_structure": "2:4"
    },
    "tie_word_embeddings": false,
    "torch_dtype": "float32",
```

Satrat:
> I tried using this to compress an existing model on HF. I chose the TinyLlama here: https://huggingface.co/neuralmagic/TinyLlama-1.1B-Chat-v1.0-pruned2.4/tree/main
> 
> There are a few issues I noticed:
> 
> * It seems like float32 is always used to save, where we should preserve the existing dtype of the model
> * `SparseAutoModelForCausalLM.save_compressed` seems to only produce the model weights and config file in the output directory, where we should be including other needed files like the tokenizer. This may be addressed by inheriting instead of using a static function
> 
> Code:
> 
> ```python
> from sparseml.transformers import SparseAutoModelForCausalLM
> 
> MODEL_PATH = "neuralmagic/TinyLlama-1.1B-Chat-v1.0-pruned2.4"
> OUTPUT_PATH = "./test_compress_output"
> 
> model = SparseAutoModelForCausalLM.from_pretrained(MODEL_PATH, device_map="cuda:0")
> SparseAutoModelForCausalLM.save_compressed(model, OUTPUT_PATH)
> ```
> 
> Output:
> 
> ```
> config.json: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 672/672 [00:00<00:00, 7.52MB/s]
> pytorch_model.bin: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 2.20G/2.20G [01:51<00:00, 19.8MB/s]
> /home/mgoin/venvs/nm/lib/python3.10/site-packages/torch/_utils.py:831: UserWarning: TypedStorage is deprecated. It will be removed in the future and UntypedStorage will be the only storage class. This should only matter to you if you are using storages directly.  To access UntypedStorage directly, use tensor.untyped_storage() instead of tensor.storage()
>   return self.fget.__get__(instance, owner)()
> generation_config.json: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 132/132 [00:00<00:00, 1.47MB/s]
> 2024-03-20 19:29:24 sparseml.transformers.utils.helpers INFO     model_path is a huggingface model id. Attempting to download recipe from https://huggingface.co/
> 2024-03-20 19:29:24 sparseml.transformers.utils.helpers INFO     Found recipe: recipe.yaml for model id: neuralmagic/TinyLlama-1.1B-Chat-v1.0-pruned2.4. Downloading...
> recipe.yaml: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 237/237 [00:00<00:00, 2.64MB/s]
> Logging all SparseML modifier-level logs to sparse_logs/20-03-2024_19.29.24.log
> 2024-03-20 19:29:24 sparseml.core.logger.logger INFO     Logging all SparseML modifier-level logs to sparse_logs/20-03-2024_19.29.24.log
> 2024-03-20 19:29:24 sparseml.core.recipe.recipe INFO     Loading recipe from file /home/mgoin/.cache/huggingface/hub/models--neuralmagic--TinyLlama-1.1B-Chat-v1.0-pruned2.4/snapshots/22ff818572f6fb2bd02110dd0b40c0169533c6da/recipe.yaml
> manager stage: Model structure initialized
> 2024-03-20 19:29:24 sparseml.pytorch.model_load.helpers INFO     Applied an unstaged recipe to the model at neuralmagic/TinyLlama-1.1B-Chat-v1.0-pruned2.4
> 2024-03-20 19:29:24 sparseml.pytorch.model_load.helpers WARNING  Model state was not reloaded for SparseML: could not find model weights for neuralmagic/TinyLlama-1.1B-Chat-v1.0-pruned2.4
> Compressing model: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 201/201 [00:04<00:00, 43.20it/s]
> ```
> 
> Looking at the output of the save_compressed shows a larger file size, but I think this is because it is saving as float32, even though the model is originally float16:
> 
> ```
> ll test_compress_output 
> total 2.5G
> -rw-r--r-- 1 mgoin mgoin  906 Mar 20 19:29 config.json
> -rw-r--r-- 1 mgoin mgoin  124 Mar 20 19:29 generation_config.json
> -rw-r--r-- 1 mgoin mgoin 2.5G Mar 20 19:29 model.safetensors
> ```
> 
> Here is the section of the config.json in that directory that mentions the sparsity level and dtype:
> 
> ```
>     "sparsity_config": {
>         "format": "sparse_bitmask",
>         "global_sparsity": 44.0587486922757,
>         "sparsity_structure": "2:4"
>     },
>     "tie_word_embeddings": false,
>     "torch_dtype": "float32",
> ```

`AutoModelForCausalLM` behaves the same for both of these points:

```python
from transformers import AutoModelForCausalLM

MODEL_PATH = "neuralmagic/TinyLlama-1.1B-Chat-v1.0-pruned2.4"
OUTPUT_PATH = "./test_compress_output_tiny_llama"

model = AutoModelForCausalLM.from_pretrained(MODEL_PATH, device_map="cuda:0")
model.save_pretrained(OUTPUT_PATH)
```

Output:
```
(.venv) sadkins@gpuserver6:/nm/drive0/sadkins/sparseml$ ls test_compress_output_tiny_llama/
config.json  generation_config.json  pytorch_model.bin
```
config.json shows `"torch_dtype": "float32",` and file size is 4GB

Adding in the `torch_dtype="auto"` argument to `from_pretrained` fixes the typing issue for both the sparse and non-sparse classes. For the tokenizer not being saved, isn't it expected that we would have to save it separately? 

```python
from sparseml.transformers import SparseAutoModelForCausalLM, SparseAutoTokenizer

MODEL_PATH = "neuralmagic/TinyLlama-1.1B-Chat-v1.0-pruned2.4"
OUTPUT_PATH = "./test_compress_output_tiny_llama"

model = SparseAutoModelForCausalLM.from_pretrained(MODEL_PATH, device_map="cuda:0", torch_dtype="auto")
tokenizer = SparseAutoTokenizer.from_pretrained(MODEL_PATH)
SparseAutoModelForCausalLM.save_compressed(model, OUTPUT_PATH)
tokenizer.save_pretrained(OUTPUT_PATH)
```

Output:
```
(.venv) sadkins@gpuserver6:/nm/drive0/sadkins/sparseml$ ls test_compress_output_tiny_llama/
config.json  generation_config.json  model.safetensors  special_tokens_map.json  tokenizer_config.json  tokenizer.json
```
config.json shows `"torch_dtype": "float16",` and file size is 1.28GB

We could default to `torch_dtype="auto"` rather than None in SparseModelForCausalLM, I agree its a more intuitive default, but it would change the behavior from the Hugging Face parent

src/sparseml/transformers/utils/sparse_model.py
```diff
@@ -93,6 +101,18 @@ def skip(*args, **kwargs):
...
+            sparsity_config = CompressionConfig.load_from_registry(
+                format, **sparsity_config
+            )
+            compressor = ModelCompressor.load_from_registry(
+                format, config=sparsity_config
+            )
+
         # temporarily set the log level to error, to ignore printing out long missing
         # and unexpected key error messages (these are EXPECTED for quantized models)
         logger = logging.getLogger("transformers.modeling_utils")
```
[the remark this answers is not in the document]
Satrat: Tried this out, but the runtime ended up being equivalent to what we currently have (probably because all of our keys are mismatched so no weights are loaded anyways) so I'm going to leave it out

src/sparseml/transformers/utils/sparse_model.py
```diff
@@ -112,6 +137,85 @@ def skip(*args, **kwargs):
             )
         return model
 
+    @staticmethod
+    def save_pretrained(
```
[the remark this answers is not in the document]
Satrat: Latest commit now has this update! PR description is updated

src/sparseml/transformers/compression/README.md
```diff
@@ -30,52 +30,131 @@ needed for decompression in the compressed state_dict:
...
+# should contain model.safetensors or model.safetensors.index.json
+model_path = "/PATH/TO/COMPRESSED_MODEL"
+
+model = SparseAutoModelForCausalLM.from_pretrained(
+    model_name_or_path=model_path,
+    **model_kwargs,
+)
+```
+
+Saving a compressed model with an explicitly provided compression config. The config
```
robertgshaw2-neuralmagic, line 63: Do we need this?

What is the advantage of enabling a user to provide a `SparsityConfig` as opposed to inferring it from the model attributes as per below?

This feels like a way that someone could trip themselves up

src/sparseml/transformers/compression/README.md
```diff
@@ -30,52 +30,131 @@ needed for decompression in the compressed state_dict:
...
+Saving a compressed model, inferring the config from the model attributes
+
+```python
+model.save_compressed(
+    model,
+    save_directory=output_dir,
+)
+
+# alternative
+model.save_pretrained(
```
robertgshaw2-neuralmagic, line 88: I think having multiple ways of doing something is confusing

So let's eliminate any external reference to `model.save_compressed` (if not too much problem to remove internally that would be great too) and standardize on:

```python
model.save_pretrained(
   save_directory="output_dir",
   save_compressed=True
)

File: src/sparseml/transformers/compression/README.md [hunk 8]
```diff
@@ -30,52 +30,131 @@ needed for decompression in the compressed state_dict:
...
+    save_directory=output_dir,
+    save_compressed=True
+)
+```
+
+Saving a model in the dense format, but still include a sparsity config in `config.json`
```
robertgshaw2-neuralmagic, line 94: Could this just be the default if the user does not provide `save_compressed=True`?

I am thinking of the following scenario:

- User prunes model in SparseML
- User exports model with `model.save_pretrained(save_directory=output_dir)`, forgetting to add a SparsityConfig
- User tries to load the model to use the sparse kernels in vllm. There is no sparsity config so vllm does not detect sparsity and it runs the dense kernels :(

I think this is very likely to happen in the case of new users.

In some regards, Im suggesting that every model exported from `SparseAutoModelForCausalLM` should have a `sparsity_config` associated with it

I think this will prevent a lot of issues down the road

An alternative would be to make `save_compressed=True` the default

src/sparseml/transformers/utils/sparse_model.py
```diff
@@ -26,22 +26,28 @@
...
 
-__all__ = ["SparseAutoModel", "SparseAutoModelForCausalLM", "get_shared_tokenizer_src"]
+_LOGGER = logging.getLogger(__name__)
 
 
-_LOGGER = logging.getLogger(__name__)
+__all__ = ["SparseAutoModel", "SparseAutoModelForCausalLM", "get_shared_tokenizer_src"]
 
 
 class SparseAutoModelForCausalLM(AutoModelForCausalLM):
```
robertgshaw2-neuralmagic, line 53: Do you think this should be implemented in `SparseAutoModelForCausalLM` or `SparseAutoModel`?

Just curious since if we expand to encoders / encoder-decoder / multi-modal, then all of this will have to be ported

Maybe something we do in a future PR

src/sparseml/transformers/compression/README.md
Satrat: This is mainly to future-proof, in the future we may have multiple types of sparse compression for instance. Currently, this allows the user to specify a DenseConfig() where we don't compress the model on disk but still store the global_sparsity and structure in the config

src/sparseml/transformers/compression/README.md
Satrat: It would be simple to remove internally, I'm not tied to keeping it but it was requested by @mgoin to add this in. GPTQ has a similar `model.save_quantized`

src/sparseml/transformers/compression/README.md
Satrat: I did originally have the default as this DenseConfig(), but as @mgoin pointed out it requires we read the whole model to calculate the global sparsity which can take a bit of time for large models. For 7b models its not that bad, I think its takes about a minute, but could become unruly for larger models

I don't love the idea of compressing by default, because it limits users to only being able to load the model within our ecosystem. I think we either accept the overhead of calculating global sparsity, or we could have the default be to set the dense sparsity config without filling in all the details

src/sparseml/transformers/utils/sparse_model.py
Satrat: `SparseAutoModel` has a lot of legacy code right now for old transformers flows, I do agree it makes sense to put this code somewhere more general but I think we should hold off until we're actually adding in additional model types

robertgshaw2-neuralmagic: This makes sense

src/sparseml/transformers/compression/README.md
robertgshaw2-neuralmagic: That makes sense on why compressing by default is an issue. Agree especially for flows where someone fine-tunes after exporting or needs to evaluate

I think it is worth taking the slow export times to ensure that every model exported from SparseML can be loaded properly into nm-vllm. Thoughts @mgoin ?

src/sparseml/transformers/compression/README.md
robertgshaw2-neuralmagic: @mgoin can you clarify why we need `save_compressed`?

I don't see what advantage there is to having multiple pathways to do the same thing

src/sparseml/transformers/compression/README.md
robertgshaw2-neuralmagic: That makes sense

src/sparseml/transformers/compression/README.md
mgoin: Yeah I'm glad you brought up the detection of sparsity in the model being important for algo selection, not just compression. I agree with your perspective and think we should calculate+store the sparsity in the config when using `SparseAutoModelForCausalLM`. Maybe a simple tqdm while calculating the sparsity would help the user know what's going on for a few minutes - it could also print beforehand that it is happening and note "it can be skipped by passing in X". Sorry for the back-and-forth @Satrat !

src/sparseml/transformers/compression/README.md
mgoin: It is simply that I like the AutoGPTQ `save_quantized` that is a thin wrapper around `save_pretrained` with the right arg. If you think it is confusing then I'm not deeply tied to it and it can be removed

src/sparseml/transformers/compression/utils/compress_save.py
```diff
@@ -0,0 +1,155 @@
...
+                )
+                sparsity_config_data = sparsity_config.dict()
+                config_file_path = os.path.join(save_directory, CONFIG_NAME)
+
+                # add the sparsity config to the model's config file
+                with open(config_file_path, "r") as config_file:
+                    config_data = json.load(config_file)
+                config_data[SPARSITY_CONFIG_NAME] = sparsity_config_data
+                with open(config_file_path, "w") as config_file:
+                    json.dump(config_data, config_file, indent=4, sort_keys=True)
```
mgoin, line 117: From checking a few config files, it looks like `indent=2` might be standard? Here's an example https://huggingface.co/neuralmagic/OpenHermes-2.5-Mistral-7B-pruned50/blob/main/config.json

mgoin:
@Satrat thanks for figuring out those dtype and saving confusions, I think you're all right there so nothing to change there. I'll look a bit more on HF saving/uploading flows to make sure I'm testing it right.

dbogunowicz (approved):

src/sparseml/transformers/compression/utils/compress_save.py
```diff
@@ -67,16 +71,23 @@ def save_pretrained_wrapper(
...
             model = model_ref()
 
             if sparsity_config is not None:
-                # if a sparsity config is provided, always save compressed
                 sparsity_config.fill_config_details(model)
-                save_compressed = True
-            elif save_compressed:
+            elif not skip_compression_stats:
                 # try to infer a sparsity config from the model if none is provided
+                _LOGGER.warning(
```
mgoin, line 85: Maybe this should just be an "info" level, since it will run by default all the time

src/sparseml/pytorch/model_load/helpers.py
```diff
@@ -233,6 +233,7 @@ def save_model_and_recipe(
     save_path: str,
     tokenizer: Optional[Any] = None,
     save_safetensors: bool = False,
```
mgoin, line 235: should we save_safetensors true by default now?

src/sparseml/transformers/compression/utils/compress_save.py
```diff
@@ -0,0 +1,130 @@
...
+                )
+
+            if sparsity_config is None:
+                # model is not sparse, save as dense
+                return original_save_pretrained.__get__(model, model_class)(
+                    save_directory, **kwargs
+                )
+
+            # if we've gotten to this point we have a config so we can run compression
+            kwargs["safe_serialization"] = True
```
mgoin, line 98: i guess this is where it doesn't matter about safetensors

mgoin (dismissed):

src/sparseml/transformers/compression/utils/compress_save.py
Satrat: Compression is only set up to work with safetensors files, so we force it if we're running compression

PR #2191, opened 2024-03-20
Fix Distill Teacher Eval Mode

Satrat:
We were previously having to set the teacher to eval mode on each forward pass in distillation. This ended up being caused by how we were wrapping the teacher model into the student. By overwriting the train() function in `KDModelWrapper` to only affect the student, we keep the teacher in eval_mode for the whole run.

This gives a minor speed boost but doesn't fully fix the runtime discrepancies between llama-recipes and SparseML. Running bf16 distillation on a llama1.1b model on GPU6(4 active GPUs, device batch size 8) we get:
* 1.6sec/batch on main
* 1.5sec/batch on this branch
* 1.1sec/batch on llama-recipes

### DIFF
--- a/integrations/huggingface-transformers/finetuning/example_fsdp_config.yaml
+++ b/integrations/huggingface-transformers/finetuning/example_fsdp_config.yaml
@@ -15,7 +15,7 @@ fsdp_config:
 machine_rank: 0
 main_training_function: main
 num_machines: 1
-num_processes: 2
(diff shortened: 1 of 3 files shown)


anmarques (dismissed):
Nice catch

bfineran (approved):

shubhra (approved):

PR #2192, opened 2024-03-20
[transformers] Prompt masking

xkazuka:
Applying character masks to a prompts in the format `[foo]some text here\n[bar]response here`, to mask characters owned by`[bar]`


bfineran (dismissed):

tests/sparseml/transformers/utils/test_helpers.py
```diff
@@ -166,3 +167,29 @@ def test_save_zoo_directory(tmp_path, stub):
...
+            "[Response]",
+            (
+                "111111111111111111111111"  # 24
+                "00000000000000000"  # 17
+                "1111111111111111111111111111"  # 28
+                "0000000000000"  # 13
+            ),
+        ),
+    ],
+)
```
Satrat, line 193: Could we add another test case here where the text doesn't start with a prompt?

src/sparseml/transformers/finetune/data/custom.py
```diff
@@ -108,5 +107,7 @@ def get_remove_columns_from_dataset(
             remove_columns.remove(self.text_column)
         if self.PROMPT_KEY in remove_columns:
             remove_columns.remove(self.PROMPT_KEY)
+        if "mask" in remove_columns:
```
Satrat, line 110: nit: could we make "mask" a constant like we do for self.PROMPT_KEY? It makes it easier to access the column in other parts of the code

src/sparseml/transformers/utils/helpers.py
```diff
@@ -554,3 +555,49 @@ def fetch_recipe_path(target: str):
...
+    :param string: The input string to be masked.
+    :param prompt: The prompt string to identify characters to keep visible.
+    :param censor: The censor string to identify characters to obscure.
+
+    Returns:
+        str: A string representing the mask where '1' indicates visible
+        characters and '0' indicates obscured characters.
+
+    """
+    mask = ["1"] * len(string)
```
Satrat, line 575: Is there a reason mask needs to be a list of chars? I think a boolean mask would be more intuitive

src/sparseml/transformers/utils/helpers.py
```diff
@@ -554,3 +555,49 @@ def fetch_recipe_path(target: str):
...
+
+def generate_mask(string: str, prompt: str, censor: str) -> str:
+    """
+    Generate a mask based on provided prompt and censor strings to obscure
+    characters in the input string.
+
+    Args:
+    :param string: The input string to be masked.
+    :param prompt: The prompt string to identify characters to keep visible.
+    :param censor: The censor string to identify characters to obscure.
```
Satrat, line 568: This is backwards, we want to obscure the prompt and keep the non-prompt tokens visible. I think "censor" is a confusing variable name here too, maybe call it response?

src/sparseml/transformers/finetune/data/base.py
```diff
@@ -156,16 +157,29 @@ def group_text_fn(data):
...
+            labels = data["input_ids"].copy()
+            if "offset_mapping" in data:
+                offset_mapping = data["offset_mapping"]
+                # get the character level mask
+                mask = data.get("mask")
+                if mask is not None:
+                    for i, (start, end) in enumerate(offset_mapping):
+                        # if any char is to be filtered
+                        if "0" in mask[start:end]:
+                            labels[i] = LABELS_MASK_VALUE
```
Satrat, line 169: I think this is backwards, "0" corresponds to chars that aren't the prompt right? We want to mask out character that are part of the prompt

Satrat (changes requested):
Calculation of the mask looks good but pretty sure its being applied backwards. Lets add a test case to test the whole pipeline so we catch this error(ie text -> tokenization -> text and confirm the labels are masked out only on the prompts.

Also, can you update the internal_docs repo with instructions on how to use this new feature on a custom dataset: https://github.com/neuralmagic/internal-docs/blob/main/teams/ml-engineering/sparseml/text_generation/custom_datasets.md

src/sparseml/transformers/utils/helpers.py
horheynm: The reason for not having list of bool is because we concatenate at the end. We can do list of bool and combine at the end to a string, or output list of bool, but i dont think its too far away from a programmers intuition to use "1" and "0".  I have details in docstrings and will be happy to edit if it can be impoved.

src/sparseml/transformers/utils/helpers.py
```diff
@@ -559,47 +559,53 @@ def fetch_recipe_path(target: str):
...
+    :param prompt: The prompt string to identify characters to obscure.
+    :param response: The response string to identify characters to keep visible.
 
     Returns:
         str: A string representing the mask where '1' indicates visible
         characters and '0' indicates obscured characters.
 
     """
+    if prompt is None:
+        prompt = ""
```
Satrat, line 580: nit: could skip this and do `prompt: str = ""` in the function definition instead

tests/sparseml/transformers/utils/test_helpers.py
```diff
@@ -170,26 +170,43 @@ def test_save_zoo_directory(tmp_path, stub):
...
 @pytest.mark.parametrize(
-    "string, prompt, censor, expected_mask",
+    "string, response, prompt, expected_mask",
     [
-        ("[foo]hello\n\n[bar]world", "[foo]", "[bar]", "1111111111110000000000"),
+        (
+            ("[foo]hello\n\n" "[bar]world"),
+            "[bar]",
+            "[foo]",
+            ("000000000000" "1111111111"),
```
Satrat, line 179: Do we support a situation where we don't start with a prompt? Ie a mask like "111100000111110000"? If so can we add another unit test case for that, or if not note this in the `generate_mask` function docstring?

tests/sparseml/transformers/utils/test_helpers.py
```diff
@@ -170,26 +170,43 @@ def test_save_zoo_directory(tmp_path, stub):
...
+            ("hello\n\n" "[bar]world"),
+            "[bar]",
+            None,
+            ("0000000" "1111111111"),
+        ),
     ],
 )
-def test_generate_mask(string, prompt, censor, expected_mask):
-    assert generate_mask(string, prompt, censor) == expected_mask
+def test_generate_mask(string, response, prompt, expected_mask):
```
Satrat, line 211: Could we also an in a more E2E test case? Ie loading, processing and tokenizing a few rows of a dataset

Satrat (changes requested):
Overall LGTM, just want to see a few more test cases (see comments)

Satrat (dismissed):

bfineran (dismissed):

jeanniefinks:

PR #2195, opened 2024-03-23
One shot user flow tweaks

robertgshaw2-redhat:
Enables:
- applying one shot to BERT (previously would fail b/c `labels` are not allowed to be passed to BERT)
- calling export on a model rather than on a file path (previously would fail during `helper_functions.create_data_loader` b/c `sequence_length` is calculated during `helper_functions.create_model`, which does not run if a `model` is passed to `export`

Also, there seems to be a stray import from sparsezoo which is failing

- Note: need to check whether the labels change impacts the LLMs (though I don't think the labels are used here)

### DIFF
--- a/src/sparseml/transformers/finetune/data/custom.py
+++ b/src/sparseml/transformers/finetune/data/custom.py
@@ -20,7 +20,7 @@ from sparseml.transformers.finetune.data import TextGenerationDataset
 from sparseml.transformers.utils.preprocessing_functions import (
     PreprocessingFunctionRegistry,
 )
-from sparsezoo.utils.helpers import import_from_path
+# from sparsezoo.utils.helpers import import_from_path
 
 
 @TextGenerationDataset.register(name="custom", alias=["json", "csv"])
--- a/src/sparseml/transformers/finetune/runner.py
+++ b/src/sparseml/transformers/finetune/runner.py
@@ -154,7 +154,6 @@ class StageRunner:
         tokenized_dataset = self.get_dataset_split("calibration")
         if "labels" in tokenized_dataset.column_names:
             tokenized_dataset = tokenized_dataset.remove_columns("labels")
-    
         calib_data = format_calibration_data(
             tokenized_dataset=tokenized_dataset,
             num_calibration_samples=self._data_args.num_calibration_samples,
--- a/src/sparseml/transformers/finetune/runner.py
+++ b/src/sparseml/transformers/finetune/runner.py
@@ -151,8 +151,12 @@ class StageRunner:
         """
         _LOGGER.info("*** One Shot ***")
 
+        tokenized_dataset = self.get_dataset_split("calibration")
+        if "labels" in tokenized_dataset.column_names:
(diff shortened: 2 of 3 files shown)


robertgshaw2-neuralmagic:
- Recipe:

```yaml
test_stage:
  obcq_modifiers:
    QuantizationModifier:
      ignore:
        - Tanh
        - GELUActivation
      post_oneshot_calibration: true
      scheme_overrides:
        # For the embeddings, only weight-quantization makes sense
        Embedding:
          input_activations: null
          weights:
            num_bits: 8
            symmetric: false
    SparseGPTModifier:
      sparsity: 0.0
      quantize: true
      targets: ["re:encoder.layer.\\d+$"]
```

- Script
```python
from sparseml.transformers import oneshot
from sparseml import export

from transformers import AutoModel, AutoTokenizer
from datasets import load_dataset

import argparse
parser = argparse.ArgumentParser()
parser.add_argument("--recipe-path", type=str)

MODEL_ID = "sentence-transformers/all-MiniLM-L6-v2"
DATASET_ID = "neural-bridge/rag-dataset-12000"
NUM_SAMPLES = 512

if __name__ == "__main__":
    args = parser.parse_args()

    model = AutoModel.from_pretrained(MODEL_ID)
    tokenizer = AutoTokenizer.from_pretrained(MODEL_ID)

    ds = load_dataset(DATASET_ID, split="train")
    ds = ds.shuffle(seed=42).select(range(NUM_SAMPLES))
    ds = ds.rename_column("context", "text")

    oneshot(
        model=model,
        tokenizer=tokenizer,
        dataset=ds,
        recipe=args.recipe_path,
        output_dir="./minilm-int8-pt",
        max_seq_length=model.config.max_position_embeddings,
        concatenate_data=False,
    )

    export(model=model, integration="transformers", task="mlm", target_path="example_export")
```

- To run:

```bash
python3 example.py --recipe-path recipe.yaml
```

src/sparseml/transformers/finetune/runner.py
```diff
@@ -151,8 +151,11 @@ def one_shot(self, stage: Optional[str] = None):
         """
         _LOGGER.info("*** One Shot ***")
 
+        tokenized_dataset = self.get_dataset_split("calibration")
+        if "labels" in tokenized_dataset.column_names:
+            tokenized_dataset = tokenized_dataset.remove_columns("labels")
```
mgoin, line 156: what is the issue here? if this is relatively safe to keep, then leave a comment on why this specific column needs to be removed

src/sparseml/transformers/finetune/data/custom.py
```diff
@@ -20,7 +20,7 @@
 from sparseml.transformers.utils.preprocessing_functions import (
     PreprocessingFunctionRegistry,
 )
-from sparsezoo.utils.helpers import import_from_path
+# from sparsezoo.utils.helpers import import_from_path
```
mgoin, line 23: what is the issue here?

robertgshaw2-neuralmagic:
Sorry - I didn't mean to put a PR on this ... I was just trying to hack something together. Will close

PR #2199, opened 2024-03-27
Forkless transformers [4.34.1]

dbogunowicz:


src/sparseml/export/validators.py
```diff
@@ -17,9 +17,12 @@
 import os.path
 from collections import OrderedDict
 from pathlib import Path
-from typing import Callable, List, Optional, Union
+from typing import Callable, List, Optional
```
mgoin, line 20: Nit: add back Union here

src/sparseml/transformers/sparsification/modification/base.py
```diff
@@ -0,0 +1,65 @@
...
+
+from sparseml.base import get_version
+
+
+_LOGGER = logging.getLogger(__name__)
+
+__all__ = ["check_transformers_version"]
+
+_TRANSFORMERS_MIN_VERSION = "4.36.0"
+_TRANSFORMERS_MAX_VERSION = "4.37.0"
```
mgoin, line 27: I'm okay with this as long as it is a warning, even if it is annoying to maintain. Speaking of which, we should probably change the torch max version to a warning rather than a failure

src/sparseml/transformers/sparsification/modification/modify_model.py
```diff
@@ -0,0 +1,70 @@
...
+    compatible with the SparseML library.
+    The model will be modified, if there exist a modification
+    function for the model in the registry of modifications.
+    Otherwise, the original model will be returned.
+
+    :param model: The original HuggingFace transformers model
+    :return: The potentially modified model
+    """
+    model_name = model.__class__.__name__
+    NM_DISABLE_MODEL_MODIFICATION = os.environ.get(
```
mgoin, line 40: Maybe should be NM_DISABLE_TRANSFORMERS_MODIFICATION

src/sparseml/transformers/sparsification/modification/modifying_llama.py
```diff
@@ -0,0 +1,251 @@
...
+class MatMulRightInput_PV(QuantizableIdentity):
+    ...
+
+
+class MatMulOutput_QK(QuantizableIdentity):
+    ...
+
+
+class MatMulOutput_PV(QuantizableIdentity):
+    ...
```
mgoin, line 93: Maybe these could be pulled into the common quantized modules file? Since some of these are duplicates in mistral

src/sparseml/transformers/utils/helpers.py
```diff
@@ -235,7 +235,8 @@ def resolve_sequence_length(config: AutoConfig) -> int:
 
 
 def resolve_recipe(
```
mgoin, line 237: Would you expect there to be any external usage of this function?

tests/sparseml/transformers/sparsification/modification/conftest.py
```diff
@@ -0,0 +1,157 @@
...
+from transformers import AutoConfig, AutoModel
+
+from accelerate import init_empty_weights
+from sparseml.transformers import SparseAutoConfig, SparseAutoModelForCausalLM
+from sparseml.transformers.sparsification.modification import modify_model
+
+
+@pytest.fixture
+def mistral_zoo_model():
+    stub = "zoo:mistral-7b-evolcodealpaca_mistral_pretrain-pruned50_quantized"
```
mgoin, line 27: Should we find smaller models to use here? Even if the weights aren't downloading, it is a lot of memory to allocate

tests/sparseml/transformers/sparsification/modification/test_modifying_llama.py
```diff
@@ -0,0 +1,69 @@
...
+        - MatMulOutput_PV
+      scheme_overrides:
+        Embedding:
+          input_activations: null
+          weights:
+            num_bits: 8
+            symmetric: False"""
+
+
+def test_modifying_llama(llama_model, shared_helper_functions):
```
mgoin, line 42: We should get to the point where we can test making a modified model and an unmodified model, and then run through them comparing outputs. This may help catch the forward passes changing between transformers versions. This may be too intensive for current hardware, but I think should absolutely be setup with the gpu runners - thoughts?

mgoin (commented):
Nice job! Appreciate the modification tests and just have a few comments

src/sparseml/transformers/sparsification/modification/base.py
dbogunowicz: @mgoin regarding the maintenance -> how bout we just parse the accepted transformers range from `setup.py` (simple regex on the `cat` of this file), instead of returning to this script and manually editing the range?

src/sparseml/transformers/utils/helpers.py
dbogunowicz: I would be surprised. It is not user-facing. Matter of fact, the only usage of it in the `src` happens within the `SparseAutoModelForCausalLM`

tests/sparseml/transformers/sparsification/modification/conftest.py
dbogunowicz: The crux is that this test tests sparsezoo models specifically, and we do not have mistral models smaller than 7b there (or in general afaik)

tests/sparseml/transformers/sparsification/modification/test_modifying_llama.py
dbogunowicz: Good idea. This is a big step toward robust detection of problems that arise from the transformers upgrade.

src/sparseml/transformers/sparsification/modification/base.py
dbogunowicz: @mgoin it's already emitting a warning

setup.py
```diff
@@ -79,8 +79,7 @@
     "opencv-python<=4.6.0.66",
 ]
 _transformers_deps = _pytorch_deps + [
-    f"{'nm-transformers' if is_release else 'nm-transformers-nightly'}"
-    f"~={version_nm_deps}",
+    "transformers<3.35 ",
```
dbogunowicz, line 82:
```suggestion
    "transformers<4.35.0",
```

src/sparseml/transformers/sparsification/modification/base.py
```diff
@@ -0,0 +1,65 @@
...
+from packaging import version
+
+from sparseml.base import get_version
+
+
+_LOGGER = logging.getLogger(__name__)
+
+__all__ = ["check_transformers_version"]
+
+_TRANSFORMERS_MIN_VERSION = "4.36.0"
```
dbogunowicz, line 26:
```suggestion
_TRANSFORMERS_MIN_VERSION = "4.34.1"
```

src/sparseml/transformers/sparsification/modification/base.py
dbogunowicz, line 27:
```suggestion
_TRANSFORMERS_MAX_VERSION = "4.35.0"
```

mgoin: Cool, thank you

mgoin (dismissed):

PR #2201, opened 2024-03-28
[BugFix] Loading channelwise quantized models

rahul-tuli:
Bugfix for the following ticket: https://app.asana.com/0/1205229323407165/1206936014133703/f

## Test plan:

The failing commands in the test ticket now work!

### Issue description

We did not incorporate a check for the safetensors file(s) during the state dictionary reload process. This was due to the early return within the function reload_model_state, which resulted in the initialize_channel_wise_scale_zp function not being invoked on the model. Consequently, this mismatched the shapes of zero points and scales, leading to the following error during inference: `RuntimeError: dimensions of scale and zero-point are not consistent with input tensor`

### DIFF
--- a/src/sparseml/pytorch/model_load/helpers.py
+++ b/src/sparseml/pytorch/model_load/helpers.py
@@ -22,6 +22,7 @@ import torch
 from torch.nn import Module
 
 import sparseml.core.session as session_manager
+from safetensors import safe_open
 from sparseml.core.framework import Framework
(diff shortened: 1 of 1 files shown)


Satrat (approved):
Good catch!

mgoin (approved):
LGTM, nice explanation

bfineran (approved):

PR #2204, opened 2024-03-29
Update transformer version to <4.40

dbogunowicz:
Changelog:
- support for `transformers<4.40`
- update the `forward` methods in the `modify...` scripts in `sparseml.transformers.sparsification.modification`, to be compatible with the latest transformers version
- renaming `SiLUActivation` to `SiLU` in sample recipes
- removing the deprecated `use_cuda_amp` argument and `ShardedDDPOption` class


setup.py
```diff
@@ -79,7 +79,7 @@
     "opencv-python<=4.6.0.66",
 ]
 _transformers_deps = _pytorch_deps + [
-    "transformers<4.35.0",
+    "transformers<4.40",
     "datasets<=2.14.6",
```
mgoin, line 83:
```suggestion
    "datasets<2.19",
```
or we could just put a minimum version and leave the ceiling open

bfineran (approved):

robertgshaw2-neuralmagic:
This can be a follow up PR

But - what do we think about having an alias internally for `SiLU` such that older recipes do not break?

mgoin (approved):

PR #2213, opened 2024-04-02
Fix Python 3.8 Compatibility 

Satrat:
Changing dictionary update from `|=` to `.update`

### DIFF
--- a/src/sparseml/export/validators.py
+++ b/src/sparseml/export/validators.py
@@ -17,8 +17,9 @@ import logging
 import os.path
 from collections import OrderedDict
 from pathlib import Path
-from typing import Callable, List, Optional, Union
+from typing import Callable, List, Optional
 from typing import OrderedDict as OrderedDictType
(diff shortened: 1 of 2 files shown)


src/sparseml/export/validators.py
```diff
@@ -17,8 +17,9 @@
 import os.path
 from collections import OrderedDict
 from pathlib import Path
-from typing import Callable, List, Optional, Union
+from typing import Callable, List, Optional
 from typing import OrderedDict as OrderedDictType
+from typing import Union
```
dbogunowicz, line 22: lets remove this file diff from the PR

dbogunowicz (approved):

src/sparseml/export/validators.py
Satrat: This just got in there from running `make style`, its needed for the quality checks to pass

mgoin (approved):

PR #2214, opened 2024-04-03
[Quantization] Add alias mapping for target types (fixes SiLU name change issue)

bfineran:
transformers recently updated usage of `SiLU` to the `SiLUActivation` class. When upgrading transformers this will cause backwards compatibility issues with existing recipes.

This PR proposes a fix by adding class alias support in quantization modifier name matching. Any reference to `SiLU` in a recipe will now also check for `SiLUActivation`

**test_plan:**
@dbogunowicz to revert test patches in https://github.com/neuralmagic/sparseml/pull/2204 to test this PR


bfineran:
@dbogunowicz to revert test patches in #2204 to test this PR

mgoin (dismissed):
This is a great solution, thanks

dbogunowicz (dismissed):
@bfineran LGTM; testing confirms that. Let's hodl off with landing this patch as proposed during our call and wait for the parent PR to land first.

dbogunowicz (approved):

PR #2217, opened 2024-04-04
[GHA] Update workflows to use new runners, push to new PyPi internal server

dsikka:
# Summary
- Updates the PyPi wheel builds and Docker Container builds such that container builds require the wheel build to be successful first 
- Updates to use the new runners with greater disk space
- Adds steps to test (base sparseml tests for now) and push the PyPi wheel to the internal PyPI server, after the wheel build is complete 
- Runs the docker container if the previous steps pass. Temporarily only builds dev containers and nightly/release container build steps are disabled as the automated push to the public PyPi server is still WIP

# Next Steps
- Need to add the automated step to push wheel to the public PyPI server. Once that's in, we can add the nightly and release docker container steps
- Possibly update the wheel build to not run if the `test-check.yml` workflow fails
- Potentially update the level of testing done on the built wheel; just running base tests for now


bfineran (dismissed):

bfineran:
landing now for initial audit

bfineran (approved):

PR #2223, opened 2024-04-05
Test to make sure `SiLU` and `SiLUActivation` are aliases

dbogunowicz:


src/sparseml/transformers/sparsification/modification/base.py
```diff
@@ -24,7 +24,7 @@
 __all__ = ["check_transformers_version"]
 
 _TRANSFORMERS_MIN_VERSION = "4.39.0"
-_TRANSFORMERS_MAX_VERSION = "4.39.2"
+_TRANSFORMERS_MAX_VERSION = "4.39.3"
```
mgoin, line 27: maybe if we just set a large hotfix number it'd be easier? i.e. 4.39.100

tests/sparseml/transformers/test_recipe_compatibility.py
```diff
@@ -0,0 +1,62 @@
...
+            num_bits: 8
+            symmetric: false"""
+
+
+def test_silu_alias_same_output(recipe, model_path):
+    model_ = SparseAutoModelForCausalLM.from_pretrained(
+        model_path, recipe=recipe.format(silu_activation="SiLU")
+    )
+    session_manager.create_session()
+    session_manager.active_session().reset()
```
mgoin, line 52: Is this needed to create two SparseAutoModelForCausalLM? We should remove this requirement for usability

dbogunowicz: yes, it is needed to reset the "recipe context" from the first model to create the second model. I am not sure how easy this would be to remove. cc: @Satrat

PR #2226, opened 2024-04-05
[BugFix] Model State Reload with Quantized Stubs in SparseAutoModelForCausalLM

rahul-tuli:
### Description

Identified a bug in the main branch where the model state fails to reload when using quantized stubs with `SparseAutoModelForCausalLM.from_pretrained(...)`. The issue was due to the `reload_model_state` method expecting weight files in a local directory, not accounting for remotely hosted model directories.

### Solution

Propose downloading the model directory before invoking `reload_model_state` to ensure weight files are available locally for model state reload.

### Testing

Tested with the following script, confirming the fix resolves the issue:

```python

model_path = "mgoin/llama2.c-stories15M-quant-pt"
m1 = SparseAutoModelForCausalLM.from_pretrained(model_path)
```

### Observations

**Before the Fix**:

Model state fails to reload due to missing local weight files, as shown in warnings and errors in the logs.

```bash
..
..
2024-04-05 17:33:33 sparseml.core.recipe.recipe INFO     Loading recipe from file /home/rahul/.cache/huggingface/hub/models--mgoin--llama2.c-stories15M-quant-pt/snapshots/aa70fc9dc46615b68f935fb5405ae7875b88b716/recipe.yaml
manager stage: Model structure initialized
2024-04-05 17:33:34 sparseml.pytorch.model_load.helpers INFO     Applied an unstaged recipe to the model at mgoin/llama2.c-stories15M-quant-pt
2024-04-05 17:33:34 sparseml.pytorch.model_load.helpers WARNING  Model state was not reloaded for SparseML: could not find model weights for mgoin/llama2.c-stories15M-quant-pt
```
**After the Fix**:

```bash
..
..
..
2024-04-05 21:17:16 sparseml.pytorch.model_load.helpers INFO     Reloaded model state after SparseML recipe structure modifications from /nm/drive0/rahul/.cache/huggingface/hub/models--mgoin--llama2.c-stories15M-quant-pt/snapshots/aa70fc9dc46615b68f935fb5405ae7875b88b716
```

Successfully reloaded model state with the fix.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204866054040952

### DIFF
--- a/src/sparseml/transformers/utils/helpers.py
+++ b/src/sparseml/transformers/utils/helpers.py
@@ -96,7 +96,7 @@ POSSIBLE_TOKENIZER_FILES = {
     "special_tokens_map.json",
     "tokenizer_config.json",
 }
-RELEVANT_HF_SUFFIXES = ["json", "md", "bin", "safetensors", "yaml", "yml"]
+RELEVANT_HF_SUFFIXES = ["json", "md", "bin", "safetensors", "yaml", "yml", "py"]
 
 
 def remove_past_key_value_support_from_config(config: AutoConfig) -> AutoConfig:
--- a/src/sparseml/transformers/utils/helpers.py
+++ b/src/sparseml/transformers/utils/helpers.py
@@ -34,8 +34,10 @@ from transformers import AutoConfig
 from transformers.trainer_utils import get_last_checkpoint
 from transformers.utils import PaddingStrategy
 
-from huggingface_hub import HUGGINGFACE_CO_URL_HOME, hf_hub_download
+from huggingface_hub import HUGGINGFACE_CO_URL_HOME, hf_hub_download, snapshot_download
 from sparseml.export.helpers import ONNX_MODEL_NAME
+from sparseml.utils import download_zoo_training_dir
+from sparseml.utils.fsdp.context import main_process_first_context
 from sparsezoo import Model, setup_model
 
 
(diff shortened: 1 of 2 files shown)


mgoin (commented):
The snapshot_download can pull down a lot of extra files since it downloads the whole folder. This might interact weirdly with the resolve_recipe call with specifically tries to download the recipe. This works, but I would like to be a bit more selective with the download

rahul-tuli:
> The snapshot_download can pull down a lot of extra files since it downloads the whole folder. This might interact weirdly with the resolve_recipe call with specifically tries to download the recipe. This works, but I would like to be a bit more selective with the download

Addressed in latest commit @mgoin

mgoin (dismissed):
Thanks a lot! It's additional complexity but I think good to have

src/sparseml/transformers/utils/helpers.py
```diff
@@ -92,6 +96,7 @@ class TaskNames(Enum):
     "special_tokens_map.json",
     "tokenizer_config.json",
 }
+RELEVANT_HF_SUFFIXES = ["json", "md", "bin", "safetensors", "yaml", "yml"]
```
mgoin, line 99: Actually we want to have py files as well if there is custom code to run

rahul-tuli: Added! much thanks for the sharp eye!

rahul-tuli:
> Thanks a lot! It's additional complexity but I think good to have

It was a great callout. Really appreciate it.

mgoin (approved):

PR #2228, opened 2024-04-05
Update OneShot Test Pathways

Satrat:
A lot of the oneshot tests were using the old pathway from before we merged finetuning and oneshot to a single script. Updating everything to use the latest pathways. A future PR will focus on refactoring tests into the new unit test framework

Asana ticket: https://app.asana.com/0/1201735099598270/1207004126198306/f


dbogunowicz (dismissed):
lgtm pending failing tests

dsikka (dismissed):

tests/sparseml/transformers/obcq/quant.yaml
```diff
@@ -0,0 +1,41 @@
...
+            num_bits: 8
+            symmetric: False
+    SparseGPTModifier:
+      sparsity: 0.0
+      block_size: 128
+      sequential_update: False
+      quantize: True
+      percdamp: 0.01
+      prunen: 0
+      prunem: 0
```
mgoin, line 33: these should be updated to mask_structure, unless this is intentional for back-compat

Satrat: good point, updated

bfineran (approved):

PR #2229, opened 2024-04-08
SparseML dependency on `compressed-tensors`

dbogunowicz:
GHA will be green after: https://github.com/neuralmagic/compressed-tensors/pulls lands


setup.py
```diff
@@ -55,6 +55,7 @@
     "GPUtil>=1.4.0",
     "protobuf>=3.12.2,<=3.20.3",
     "click>=7.1.2,!=8.0.0",  # latest version < 8.0 + blocked version with reported bug
+    "sparsetensors @ git+ssh://git@github.com/neuralmagic/sparsetensors.git",
```
bfineran, line 58: you can't build/publish a wheel with a dependency with git deps due to privacy restrictions

bfineran (changes requested):
biggest thing I'd like to talk through with you and @Satrat is moving the actual configs into sparsetensors. specifically having configs take in the sparsification structure as an input so we don't need to couple the analysis code with the config classes

Satrat:
@bfineran @dbogunowicz is there anything blocking this from being merged??

dbogunowicz:
@Satrat I'd say the readiness of the pypi package. Now since we have the basic pathways in place, I will get this PR and its counterpart in vLLM in order. But let's discuss the readiness status today.

rahul-tuli (changes requested):

tests/sparseml/transformers/test_clear_ml.py
```diff
@@ -11,20 +11,23 @@
...
 
+
 try:
     from clearml import Task
 
-except Exception as err:
-    clearml = None
+    is_clearml = True
+except Exception:
+    is_clearml = False
```
dbarbuzzi, line 25: Why not something like this, since the whole file appears to be for testing clearml — then you wouldn't need any special skip decorators on the test(s).

```python
try:
    from clearml import Task
except ImportError:
    pytest.skip("clearml not installed", allow_module_level=True)
```

Alternatively, [`pytest.importorskip`](https://docs.pytest.org/en/8.0.x/reference/reference.html#pytest-importorskip-ref) works at a module-level as well:

```python
clearml = pytest.importorskip(clearml, reason="clearml module is required")
# use clearml.Task in later usage, or do Task = clearml.Task after imports to put `Task` in the namespace directly
```

dbogunowicz: awesome, thanks!

dbogunowicz: those solutions do not work ad hoc for me. let's just get this in; i really would like to unblock team with this PR

dbarbuzzi: Strange, I’ve not had issues using either approach. Whatever works, though!

bfineran (dismissed):

Satrat (approved):

Satrat (approved):

rahul-tuli (approved):

bfineran (approved):

PR #2233, opened 2024-04-09
Refactor the quantization modification logic

dbogunowicz:
# Feature Description

The `sparseml.transformers.sparsification.modification` package is a set of modifications that are applied to some of the transformer models, to make them compatible with our quantization flows.

This PR moves the act of modification away from `SparseAutoModel`, to the `QuantizationModifier`. This means we only modify the model when necessary - when we apply the quantization structure. This helps us to avoid clashes with the new behavior in `transformers`, where the models by default get initialized with `SDPAttention`.

## Notable changes
- moving the high-level `modification` logic to `sparseml.modifiers.quantization`. Keeping only the `transformers`-specific logic in the original directory
- only the models that are being supported by the new, "post-refactor" quantization modifiers "participate" in the modification. If the user wishes to initialize one of the old models, the modification happens on the initialization of `SparseAutoModel`.

To my best knowledge, the failing tests are orthogonal to the contents of this PR.

## Legacy PR description

Keeping the original PR message (analysis of the problem, that forced me to go down the final path present in the PR) for posteriority, as it contains a lot of useful context:

As reported by @Satrat, after upgrading the transformers version we did not see the expected training speedups during e.g. sparse fine-tuning process. It turned out this was caused by the `modify_model(...)` function during the initialization of the `SparseAutoModelForCausalLM`.

Let's explain what was happening using `LLaMa` as an example. 

The model as of `transformers==4.39.1` can be initialized with three types of attention:
- `LlamaSdpaAttention` - the default in the current transformers version; it uses the CUDA optimized `torch.nn.functional.scaled_dot_product_attention` for quicker computation of attention
- `LlamaAttention` - the previous default before the transformers upgrade; this is the attention type that uses `torch.matmul` method and thus is being modified by us through the `modify_model(...)` method.
- `LlamaFlashAttention` - irrelevant in the context of this write-up.

The current, erroneous behavior, was the following:
1. We were initializing the model with the default, SDPA-attention
2. Because of the misuse of the `isinstance()` method, we were overriding this attention module's `forward` method, effectively replacing SDPA-attention's `forward` method with the original attention class' `forward` method.
3. This is what was slowing down the training of the LlaMa model -> unknowingly, we were no longer using the fast CUDA-optimized attention but traditional, torch-based attention computation.

This PR hardens the modification logic - it uses a more restrictive `type()` instead of `isinstance()` to pick the correct attention type to modify. Now there is no difference in iterations per second when sparse fine tuning with or without the `modify_model(...)` function.


tests/sparseml/transformers/sparsification/modification/conftest.py
```diff
@@ -27,7 +27,11 @@ def mistral_zoo_model():
     stub = "zoo:mistral-7b-evolcodealpaca_mistral_pretrain-pruned50_quantized"
     config = SparseAutoConfig.from_pretrained(stub)
     with init_empty_weights():
-        model = SparseAutoModelForCausalLM.from_config(config)
+        # attn_implementation="eager" needed so that the model uses the
+        # appropriate attention type (MistralAttention)
+        model = SparseAutoModelForCausalLM.from_config(
+            config, attn_implementation="eager"
```
Satrat, line 33: Is this something that needs to be applied for every model? Wondering if we should default to this attention implementation

tests/sparseml/transformers/sparsification/modification/conftest.py
```diff
@@ -116,28 +128,28 @@ def check_model_modified(
...
-            and isinstance(module, module_to_replace)
+            and module.__class__.__name__ is module_to_replace.__name__
         ]
 
         modified_modules_modified_model = [
             module
             for module in modified_model.modules()
             if func_to_validate_replacement(module)
-            and isinstance(module, module_to_replace)
+            and module.__class__.__name__ is module_to_replace.__name__
```
Satrat, line 138: wondering if we should make this a helper function because its so ugly lol

bfineran (dismissed):

src/sparseml/modifiers/quantization/pytorch.py
```diff
@@ -73,6 +74,9 @@ def __init__(self, **kwargs):
 
     def on_initialize_structure(self, state: State, **kwargs):
         module = state.model.model
+        # before the structure is modified to support quantization,
+        # we need to potentially modify the model architecture
+        module = modify_model(module)
```
Satrat, line 79: Has this been tested with FSDP?? I think that is out of scope of this PR but would be good to know if it works or not

src/sparseml/transformers/sparsification/sparse_model.py
```diff
@@ -144,6 +141,8 @@ class SparseAutoModel:
     Factory class for creating sparse models using transformers AutoModel classes
     """
 
+    from sparseml.modifiers.quantization.modification import modify_model
+
```
Satrat, line 145: Why do we need this here if modification is happening within the quantization modifier?

src/sparseml/modifiers/quantization/pytorch.py
Satrat: Also, do you think we need this on line 95 too? That is where we setup for quantization when the input model hasn't been quantized set

Satrat (commented):
LGTM, but two main thoughts: can we document this environment variable somewhere with its intended usage? And could we add a test script to this PR that demonstrates the speed issue being fixed?

src/sparseml/modifiers/quantization/pytorch.py
dbogunowicz: @Satrat could we write a unit test that covers the FSDP scenario? I am confused, what do you mean by line 95?

dbogunowicz: Oh I see: `self._enable_module_qat(module)`. It seems like it. When would the model go through line 95 and not `on_initialize_structure` method?

src/sparseml/transformers/sparsification/sparse_model.py
dbogunowicz: We'd need it for everything but `SparseAutoModelForCausalLM`. Remember that the quantization modifier is only relevant for LMMs, not "classical" transformer models.

src/sparseml/modifiers/quantization/pytorch.py
Satrat: Since FSDP requires multiple GPUs we don't have any integration tests set up for it. Theres documentation on generating an FSDP config here: https://github.com/neuralmagic/sparseml/blob/main/integrations/huggingface-transformers/finetuning/configure_fsdp.md, and you can use that config for the `tests/sparseml/transformers/obcq/test_obcq.py:test_obcq_tinystories` integration test with the `quant.yaml` recipe parameter

Satrat: As for the line 95 comment, `on_initialize_structure` is for initializing a recipe that has been previously applied to a model. So it would be used in the context of reloading a quantized model. `on_initialize` is called when you're quantizing a model for the first time

tests/sparseml/transformers/sparsification/modification/test_modifying_llama.py
```diff
@@ -39,31 +39,22 @@ def llama_recipe():
...
-    apply_recipe_structure_to_model(
-        model=modify_model(llama_zoo_model), model_path=None, recipe_path=llama_recipe
+def test_modify_with_quantization_recipe_sparsezoo(
+    llama_zoo_model, llama_recipe, shared_helper_functions
+):
+    # TODO: Improve that
+    shared_helper_functions.check_model_modified_causal(
+        llama_zoo_model,
+        recipe=llama_recipe,
+        modified_module=LlamaAttentionWithQuantizableMatmuls,
```
Satrat, line 59: Could we also add this check to `tests/sparseml/transformers/obcq/test_obcq.py:test_obcq_tinystories` for the recipe cases that have quantization? This would check that we are correctly modifying the model when applying oneshot

bfineran (approved):

Satrat (approved):

PR #2237, opened 2024-04-10
[OneShot][Testing] Expand Integration tests to run for llama-7b; add gpu/auto cases

dsikka:
# Summary
- Update obcq tests into separate integration tests
- Add/Update `test_sparsities --> test_obcq_sparsity.py`; now tests LLama-7b using gpu and "auto" using the same sparsity recipe as TInyStories
- Add/Update the original `test_obcq_tinystories -->  test_obcq_completion.py`; tests quantization, quantization + sparsity for TinyStories. Tests quantization and quantization + pruning with  Llama-7b gpu 
- All GPU tests set-up to run on a nightly cadence
- All other Integration tests are separated into their own files; still need to update/add configs for them for the different cases we'd like to cover 

# Remaining Question
- For the new `test_obcq_compeltion.py` (which was previously `test_obcq_tinystories`) we are just running one-shot without any metric calculation on the results; is there anything we can add? Perplexity seems to be very long running, even for TinyStories
- How should the remaining integration tests which were not updated to be tested for llama, be updated in terms of use cases?


bfineran (approved):
looks good for initial tests - should immediately follow up to add eval to some of the tests

tests/sparseml/transformers/obcq/test_obcq_completion.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def tearDown(self):
+        shutil.rmtree(self.output)
+
+
+@requires_torch
+@requires_gpu
+@pytest.mark.integration
+@parameterized_class(parse_params(GPU_CONFIGS_DIRECTORY))
+class TestOBCQCompletionGPU(unittest.TestCase):
```
Satrat, line 70: Is there a reason we need to break this out into a separate test? Just so it runs on a different cadence? Wondering if UnitTest supports inheriting from other tests, then we could avoid the repeated lines

Satrat (changes requested):
LGTM overall! Could we add in a GPU test for test repeats? Other than that I think this is sufficient coverage for the 7b models, assuming we can add a perplexity check

tests/sparseml/transformers/obcq/test_obcq_completion.py
dsikka: the @requires_gpu is test specific but we could do a bit of clean-up to prevent repeat code.

tests/sparseml/transformers/obcq/test_obcq_completion.py
```diff
@@ -0,0 +1,107 @@
...
+        oneshot(
+            model=self.model,
+            dataset=self.dataset,
+            oneshot_device=self.device,
+            recipe=self.recipe,
+            max_seq_length=128,
+            num_calibration_samples=32,
+            pad_to_max_length=False,
+            output_dir=self.output,
+        )
```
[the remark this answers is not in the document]
dsikka: I'm using this as my example in running perplexity: https://github.com/neuralmagic/sparseml/blob/main/tests/integrations/transformers/evaluation/test_perplexity.py

Seems like just passing `open_platypus` doesn't work (instead of wikitext). Do we have a perplexity example for a smaller dataset?

bfineran (approved):

tests/sparseml/transformers/obcq/test_obcq_completion.py
```diff
@@ -0,0 +1,158 @@
...
+        total_new_ppl = 0.0
+        for idx, sample in enumerate(dataloader):
+            if idx >= iter:
+                break
+            new_output = first_tiny_model(**(tensors_to_device(sample, self.device)))
+            new_ppl = torch.exp(new_output.loss)
+            total_new_ppl += new_ppl
+
+        avg_new_ppl = total_new_ppl / iter
+        self.assertLess(avg_new_ppl, 5000)
```
Satrat, line 111: Could we make the expected perplexity an entry in the config? I don't think it makes sense to hardcode since it differs depending on the model

tests/sparseml/transformers/obcq/test_obcq_completion.py
```diff
@@ -0,0 +1,158 @@
...
+
+        iter = 10
+        dataloader = self.labeled_dataloader(dataset, self.model)
+        total_new_ppl = 0.0
+        for idx, sample in enumerate(dataloader):
+            if idx >= iter:
+                break
+            new_output = first_tiny_model(**(tensors_to_device(sample, self.device)))
+            new_ppl = torch.exp(new_output.loss)
+            total_new_ppl += new_ppl
```
Satrat, line 108: I'd wrap this in a torch.no_grad

tests/sparseml/transformers/obcq/test_obcq_completion.py
Satrat: Does this run for the 7b model or is that in a different test? We may want to run with more calibration samples and more test samples for the 7b model so we can better track if the ppl degrades. 5,000 is way too high a threshold for 7b and should be below 20 at least

bfineran (dismissed):

Satrat (approved):

bfineran (approved):

PR #2259, opened 2024-04-30
Allow torch 2.3 and remove torch ceiling version restriction

mgoin:
vLLM now requires torch 2.3.0, so we should try to raise the restriction in SparseML. Going forward, I think we shouldn't be so restrictive to newer versions of pytorch and should raise a warning at most, not an explicit exception.


rahul-tuli (dismissed):

dbogunowicz (dismissed):

mgoin:
@bfineran @robertgshaw2-neuralmagic is this good with you going forward to not restrict pytorch?

bfineran (approved):

dbogunowicz:
@mgoin should we land this (after the GHA tests have passed?)

PR #2265, opened 2024-05-03
Fixing GHA Hangs

Satrat:
Fixing a few issues
* ClearML tests are hanging, commented them out for now and removing the dependency
* HF Trainer attempts to automatically initialize a clearml logger if clearml is installed. This isn't ideal, we don't always want to log everything and there is no easy way to turn it off. Updated our Trainer script to remove this functionality for now. This was also causing tests to hang
* Moved some tests that required transformers to the correct directory
* Changed some oneshot tests to use llama 15m instead of 1.1b, 1.1b is too big for the per-commit GHA runners
* updated the runners from aws to ubuntu because they weren't loading
* fixed a weird issue in an export unit test, it was failing to download the model when we passed in an instantiated model


.github/workflows/test-check.yaml
```diff
@@ -140,7 +140,7 @@ jobs:
       - name: "🔬 Running onnx tests"
         run: make test TARGETS=onnx
   pytorch-tests:
-    runs-on: aws-avx2-64G
```
dsikka, line 143: this runner seems to running fine with our other tests so it may no longer be an issue but if we can get away with just using an ubuntu image from github, that would be more ideal (given none of the tests require VPN)

tests/sparseml/pytorch/utils/test_logger.py
```diff
@@ -46,7 +45,7 @@
             or True
         ),
         *([WANDBLogger()] if WANDBLogger.available() else []),
-        *([ClearMLLogger()] if ClearMLLogger.available() else []),
+        # *([ClearMLLogger()] if ClearMLLogger.available() else []),
```
dbogunowicz, line 48: Definitely do not want to stall landing this PR, however, let's clear the ClearML issues once and for all, it has been slowing us back significantly.

dbogunowicz: So in the context of this PR I'd keep line 48 commented out, but leave a TODO comment so that there is a footprint indicating that we should fix it.

src/sparseml/transformers/sparsification/compressed_tensors_utils.py
```diff
@@ -92,6 +92,7 @@ def save_pretrained_wrapper(
                     "compressed state please use vLLMQuantizationModifier instead."
                 )
 
+                kwargs["safe_serialization"] = False
```
dbogunowicz, line 95: I think this is not needed anymore.

.github/workflows/test-check.yaml
dbogunowicz: Could we potentially check this before landing?

tests/sparseml/export/transformers/test_generative_transformers.py
```diff
@@ -42,13 +42,13 @@ def setup(self, tmp_path, stub, task):
 
         shutil.rmtree(tmp_path)
 
-    def test_export_initialized_model_no_source_path(self, setup):
+    def test_export_initialized_model_no_source_path(self, tmp_path, stub, task):
```
dbogunowicz, line 45: Thanks for that @Satrat

dbogunowicz (commented):
Great detective work

.github/workflows/test-check.yaml
Satrat: Yup, just reverted this change, lets wait and see if it runs on the aws runner

dsikka (approved):

dbogunowicz (approved):

PR #2268, opened 2024-05-06
[MOE Quantization] Update transformers version to 4.40.0

dbogunowicz:
We need to update the transformers version to support QWEN2-MOE model, see: https://github.com/huggingface/transformers/releases/tag/v4.40.0
_(it also, fits into our goal to be constantly matching the latest release)_

#### By default, untie the vocab embedding weights
transformers `4.40.X` has now difficulty saving post one-shot models. This applies to models that are 1) quantized using new (not vLLM) QuantizationModifiers 2) in the "fakequant" state. 
HF developers made changes to the internal logic of parsing "tied weights" (such as embedding and lm_head modules) on `save_pretrained`. Those decisions are buried quite deep in the transformer's codebase. My solution is to untie the weights for the one-shot models on initialization. Benefits: It requires minimal changes on our side, and does not influence the size of saved models on disk (safe-tensors untie weights as well). Downsides: The one-shot process might be slightly less performant because of doubling the memory required to store the embedding layer in CUDA.


mgoin (approved):
LGTM once the tests pass!
Export one seems simple
```
ERROR tests/sparseml/export/transformers/test_generative_transformers.py::TestEndToEndExport::test_export_happy_path[roneneldan/TinyStories-1M-text-generation] - FileNotFoundError: [Errno 2] No such file or directory: '/tmp/pytest-of-runner/pytest-0/test_export_happy_path_ronenel0/model/.huggingface/.gitignore.lock'
```

bfineran (approved):

dbogunowicz:
@mgoin the failure in export-tests looks transient, just fyi

rahul-tuli (dismissed):

mgoin (approved):

src/sparseml/transformers/finetune/text_generation.py
```diff
@@ -148,11 +148,13 @@ def intialize_model_from_path(
         cache_dir=model_args.cache_dir,
         revision=model_args.model_revision,
         use_auth_token=True if model_args.use_auth_token else None,
+        tie_word_embeddings=False,
```
bfineran, line 151: per @Satrat let's add a warning that this is disabled and can be turned on if using the new modifier

bfineran (approved):

tests/sparseml/transformers/compression/test_quantization.py
```diff
@@ -162,11 +162,13 @@ def test_quantization_scale_and_zp(self):
...
+        # 1.1b nightly tests
         # allow for error here due to implementation differences
-        for name, (o_scale, o_zp) in old_quant_inputs.items():
-            n_scale, n_zp = new_quant_inputs[name]
-            assert math.isclose(o_scale, n_scale, abs_tol=1e-2, rel_tol=1e-2)
-            assert abs(o_zp - n_zp) < 5
+        # for name, (o_scale, o_zp) in old_quant_inputs.items():
+        #    n_scale, n_zp = new_quant_inputs[name]
+        #    assert math.isclose(o_scale, n_scale, abs_tol=1e-2, rel_tol=1e-2)
+        #    assert abs(o_zp - n_zp) < 5
```
rahul-tuli, line 171: Could we remove the commented code before merging, ideally git history should be enough to go back and add this in for larger tests?

tests/sparseml/transformers/compression/test_quantization.py
```diff
@@ -162,11 +162,13 @@ def test_quantization_scale_and_zp(self):
             assert math.isclose(o_scale, n_scale, abs_tol=1e-3, rel_tol=1e-3)
             assert o_zp == n_zp
 
+        # skipping this for commit tests on smaller model, TODO: add back in for the
```
rahul-tuli, line 165: Not a fan of TODO's in actual code, would be nice if we could remove this and track this internally(maybe via ASANA?)

src/sparseml/transformers/finetune/text_generation.py
rahul-tuli: Is this argument always `False`? Wouldn't it be nice to somehow allow users to override this? or am I missing something here? 
On a side note it would be nice to have a short comment explaining why we need this to be `False` the PR description covers this but I'd like to keep the reason somewhat close to the code? thoughts?

tests/sparseml/transformers/sparsification/test_compress_tensor_utils.py
```diff
@@ -77,7 +77,7 @@ def test_sparse_model_reload(compressed, config, dtype, tmp_path):
     )
 
     inferred_global_sparsity = SparsityConfigMetadata.infer_global_sparsity(model)
-    assert math.isclose(inferred_global_sparsity, 19.6562, rel_tol=1e-3)
+    assert math.isclose(inferred_global_sparsity, 12.234, rel_tol=1e-3)
```
rahul-tuli, line 80: Should we really be checking global sparsity here? or does it make more sense to check tensor sparsity for one of the targeted layers, and assert that the tensor_sparsity `isclose` to the actual sparsity specified in the recipe? To me this seems like a random number that could change if we change the model or make any updates like your diff does?

rahul-tuli (changes requested):
Nice! LGTM pending a response/resolution for comments

Satrat (dismissed):

tests/sparseml/transformers/sparsification/test_compress_tensor_utils.py
mgoin: +1!

bfineran (dismissed):

tests/sparseml/transformers/sparsification/test_compress_tensor_utils.py
```diff
@@ -76,8 +78,15 @@ def test_sparse_model_reload(compressed, config, dtype, tmp_path):
         tmp_path / "oneshot_out", torch_dtype=dtype
     )
 
-    inferred_global_sparsity = SparsityConfigMetadata.infer_global_sparsity(model)
-    assert math.isclose(inferred_global_sparsity, 19.6562, rel_tol=1e-3)
+    def compute_tensor_sparsity(t: torch.Tensor):
+        return t.eq(0).sum().item() / t.numel()
```
rahul-tuli, line 82: We have a function for this, it's called `tensor_sparsity` here https://github.com/neuralmagic/sparseml/blob/c234638a1d7bd276c18bdc46005434fca88f9881/src/sparseml/pytorch/utils/helpers.py#L556

Satrat (approved):

bfineran (approved):

PR #2274, opened 2024-05-09
[Testing] Update/expand finetune tests

dsikka:
# Summary
- Update finetune tests to use the updated testing framework
- Expand to add llama-7b test cases which run on a weekly basis for `test_oneshot_and_finetune` and `test_finetune_wout_recipe_custom_dataset`


tests/sparseml/transformers/finetune/finetune_oneshot_configs/gpu/gpu_config.yaml
```diff
@@ -0,0 +1,6 @@
+cadence: "weekly"
+test_type: "regression"
+model: "zoo:llama2-7b-llama2_pretrain-base"
+dataset: wikitext
+dataset_config_name: "wikitext-2-raw-v1"
```
Satrat, line 5: For the weekly tests I think we should use a model that has already been pretrained on a specific dataset then finetune on the same dataset (this is the typical research) flow. Maybe zoo:llama2-7b-ultrachat200k_llama2_pretrain-base and the ultrachat200k dataset

Satrat (commented):
Test coverage looks good to me, just had a note on the 7b model

tests/sparseml/transformers/finetune/test_oneshot_and_finetune_with_tokenizer.py
```diff
@@ -0,0 +1,72 @@
...
+        if not torch.cuda.is_available():
+            device = "cpu"
+
+        dataset_config_name = "wikitext-2-raw-v1"
+        dataset = load_dataset("wikitext", dataset_config_name, split="train[:50%]")
+        # dataset ="wikitext"
+
+        concatenate_data = True
+        run_stages = True
+        max_steps = 50
```
Satrat, line 54: Sorry one other thing I forgot to say, we should increase this value for the 7b model. This says to only run training for 50 steps and ideally should be a whole epoch (maybe for testing we could get away with half an epoch)

bfineran (dismissed):

tests/sparseml/transformers/finetune/finetune_custom/gpu/gpu_config.yaml
```diff
@@ -1,4 +1,4 @@
 cadence: "weekly"
 test_type: "regression"
-model: "zoo:llama2-7b-llama2_pretrain-base"
-file_extension: json
+model: "zoo:llama2-7b-ultrachat200k_llama2_pretrain-base"
+file_extension: json
```
Satrat, line 4: We should still specify either max_steps or num_train_epochs here, the default is 3 epochs which is way too high( we want 1 epoch)

dsikka: ah sorry I mean to ask you for an appropriate number of steps. Updated to just use num_train_epochs.

Nice catch!

bfineran (approved):

Satrat (approved):

PR #2284, opened 2024-05-14
mask_structure preservation test

rahul-tuli:
This pull request introduces an integration check to ensure the preservation of mask structure across consecutive runs. The process includes:

- **Initial pruning of the model** using a mask structure: `"N:M"`.
- **Further pruning and quantization** of the saved model.
- **Verification** that the original `"N:M"` sparsity structure is maintained, with additional sparsity applied on top.


tests/sparseml/transformers/obcq/test_mask_structure_preservation.py
```diff
@@ -0,0 +1,144 @@
...
+
+    n, m = tuple(map(int, mask.split(":")))
+    # Reshape the tensor into chunks of size m
+    tensor = tensor.view(-1, m)
+
+    # Count the number of zeros in each chunk
+    zero_counts = (tensor == 0).sum(dim=1)
+
+    # Check if the number of zeros in each chunk is n
+    return torch.all(zero_counts == n)
```
bfineran, line 47: Should be greater than or equal to

rahul-tuli: I think this is correct, this function is only used for strict checking, otherwise some other mask structure such as "3:4" will also return True right?

Or are you trying to say that when we follow apply 2:4 sparsity there could be a case where more than 2 weights are zero?

PR #2299, opened 2024-05-23
Fix errors introduced by GPTQ UX

rahul-tuli:
This PR makes main green again by:

1) Install compressed tensors from source wherever transformers deps are needed
2) Apply a commit that was missed (for transformers tests)

### DIFF
--- a/.github/workflows/test-check.yaml
+++ b/.github/workflows/test-check.yaml
@@ -164,6 +164,16 @@ jobs:
         run: pip3 install -U pip && pip3 install setuptools sparsezoo/
       - name: "Clean sparsezoo directory"
         run: rm -r sparsezoo/
+      - uses: actions/checkout@v2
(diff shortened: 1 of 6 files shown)


Satrat:
Could you add one more change to that PR:
```
from src.sparseml.modifiers.quantization.gptq.utils.gptq_wrapper import GPTQWrapper
```
should be
```
from sparseml.modifiers.quantization.gptq.utils.gptq_wrapper import GPTQWrapper
```

otherwise we won't be able to run anything outside of sparseML root directory, this was causing my 7b example runs to fail

Satrat (approved):

Satrat (dismissed):

dsikka (dismissed):

bfineran (approved):

dbogunowicz (approved):

bfineran (approved):

PR #2304, opened 2024-05-24
[GHA] Update End-to-End Nightly Build Process

dsikka:
# Summary 
- Automate the end-to-end nightly build process
- Runs nightly tests, builds wheel, publishes to the public server, builds and publishes nightly docker container
- Removes the old nightly workflows 

# Questions: 
1. Do we want to push to nightly every time we push to main? This was a previous requirement but we can't push multiple wheels to the public server with the same nightly wheel name. What do we want to happen on push to main?
2. Any additional nightly testing?


bfineran (approved):

horheynm (approved):

Satrat (approved):

.github/workflows/test-nightly.yml
```diff
@@ -33,6 +32,5 @@ jobs:
         run: |
           pytest tests/sparseml/transformers/obcq -m integration
```
dhuangnm, line 33: It looks we just run the two test suites from transformers. Can we include more to the nightly run, e.g. the deepsparse target that was run in the Makefile?

.github/workflows/test-wheel-and-publish.yml
```diff
@@ -36,24 +42,37 @@ jobs:
...
             rm -rf src
-            make test
\ No newline at end of file
+            make test
+
+    - name: Make directory for wheel
+      run: |
+          mkdir dist_s3
+          
+    - name: Pull from s3
```
dhuangnm, line 66: We're moving away from using AWS. Andy set up bucket in GCP to save the artifactories. I think we should use that going forward.

.github/workflows/test-wheel-and-publish.yml
```diff
@@ -36,24 +42,37 @@ jobs:
...
+      uses: neuralmagic/nm-actions/actions/s3_pull@main
+      with:
+        filename: ${{ inputs.whl }}
+        dst: dist_s3
+
+    - name: Publish Nightly Wheel
+      if: ${{ inputs.DEV == 'false' && inputs.RELEASE == 'false'}}
```
dhuangnm, line 73: Should we also check the test status here and only push if all tests pass?

.github/workflows/test-nightly.yml
dsikka: Are there deepsparse tests marked to run on a nightly basis?

.github/workflows/test-wheel-and-publish.yml
dsikka: I can update this in a follow-up PR

.github/workflows/test-wheel-and-publish.yml
dsikka: The nightly build wont be published if the tests pass

.github/workflows/test-nightly.yml
dsikka: Ok will put up a follow-up PR

PR #2310, opened 2024-05-29
Update Image training code for downstream datasets (iWildcam, transfer datasets)

ohaijen:
Update trochvision/train.py to accomodate:

- iWildcam. This code would be read through the ImageFolder format, but requires different resizes and crops.
- Transfer datasets (Aircraft, DTD, Flowers) - these would be read through the TorchVision API and cached in /tmp. These don't need any new transforms, just the class loading.


jeanniefinks:

PR #2316, opened 2024-06-04
activation ordering

xkazuka:
Activation Ordering implementation. 

Checked lm_eval value with `actorder=True`

```
"metrics":[{"name":"word_perplexity,none","value":10.17568878732032},

```


robertgshaw2-neuralmagic:
As it relates to the serialization of act_reordering in `compressed-tensors`, @dsikka can you make sure this works well with the code we have in vllm?

src/sparseml/modifiers/utils/layer_compressor.py
```diff
@@ -131,7 +131,7 @@ def revert_layer_wrappers(self):
             module_wrapper.free()
         self.modules = None
 
-    def compress(self):
+    def compress(self, actorder: bool = False):
```
dbogunowicz, line 134: Let's specify what this boolean flag does in the params

src/sparseml/modifiers/quantization/gptq/utils/gptq_wrapper.py
```diff
@@ -81,6 +81,7 @@ def add_batch(self, inp: torch.Tensor, out: torch.Tensor):
 
     def fasterprune(
         self,
+        actorder: bool = False,
```
dbogunowicz, line 84: Let's mention what this boolean flags does.

src/sparseml/modifiers/quantization/gptq/pytorch.py
```diff
@@ -156,7 +156,7 @@ def apply_compression(
                 layer_compressor.pre_compress()
                 _LOGGER.info(f"Calibrating {layer_compressor.name}...")
                 run_calibration_forward(self.model, dataloader, mask_padding=True)
-            layer_compressor.compress()
+            layer_compressor.compress(self.actorder)
```
dbogunowicz, line 159: is this flag also relevant in the context of SparseGPT or only GPTQ?

dbogunowicz (dismissed):
Could we have some (manual) tests in place for this as well? It's a subtle modification of the behavior, it would be great to have tested one way or another.

src/sparseml/modifiers/quantization/gptq/base.py
```diff
@@ -50,6 +50,7 @@ class GPTQModifier(Modifier):
             - LayerCompressor.revert_layer_wrappers()
 
 
+    :param actorder: Whether to use activation reordering or not
```
Satrat, line 53: Could you provide a short description of what activation reordering is? Also a nitpick but since this is a user facing parameter I think calling it activation_reordering would be better

Satrat (changes requested):
Could we add a unit test that tests the perplexity for oneshot done with this flag?

src/sparseml/modifiers/quantization/gptq/pytorch.py
horheynm: Probably just GPTQ - activation reordering is for quantization flow

src/sparseml/modifiers/quantization/gptq/base.py
horheynm: Ok tests yeah i agree - for param name it comes from gptq code 
https://github.com/AutoGPTQ/AutoGPTQ/blob/ea829c7bbe83561c2b1de26795b6592992373ef7/auto_gptq/quantization/gptq.py#L67
Other parts of the code is very similar to this code. I'm more inclined to keep the name as is for now

jeanniefinks:

PR #2323, opened 2024-06-07
Fix for Sparsity Persist

Satrat:
When debugging the Marlin24 kernels I found that the sparsity structure was not being correctly maintained. The vLLM check for sparsity structure was failing. After this GPTQ fix the problem went away, I pulled this code from the nm-AutoGPTQ codebase

### DIFF
--- a/src/sparseml/modifiers/quantization/gptq/utils/gptq_wrapper.py
+++ b/src/sparseml/modifiers/quantization/gptq/utils/gptq_wrapper.py
@@ -103,6 +103,14 @@ class GPTQWrapper(ModuleCompressionWrapper):
             W = W.t()
         W = W.float()
 
+        sparsity = tensor_sparsity(W)
+        preserve_zeros = sparsity >= SPARSITY_THRESHOLD
+        W_nz_mask = (
+            (~torch.isclose(W, torch.zeros(1, device=W.device).float())).float()
+            if preserve_zeros
+            else None
+        )
+
         tick = time.time()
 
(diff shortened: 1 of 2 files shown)


dbogunowicz:
@Satrat nice find! do you think that this may be something that also confused me over the last few days?

dbogunowicz (approved):

rahul-tuli (approved):

src/sparseml/modifiers/quantization/gptq/utils/gptq_wrapper.py
```diff
@@ -216,13 +205,21 @@ def fasterprune(
...
+                    W1[:, i:] -= w1_err
                 Err1[:, i] = err1
 
             W[:, i1:i2] = Q1
             Losses += torch.sum(Losses1, 1) / 2
 
-            W[:, i2:] -= Err1.matmul(Hinv[i1:i2, i2:])
+            w_err = Err1.matmul(Hinv[i1:i2, i2:])
+            if preserve_zeros:
+                W[:, i2:] -= w_err * W_nz_mask[:, i2:]
```
horheynm, line 220: `W_nz_mask[:, i2:]` sorry why it is from i2?
Hinv[i1:i2, i2:] is self.H so it has the full shape
W_nz_mask might be trimmed to W_nz_mask[:, i1:i2], so might not have the full shape?

rahul-tuli (approved):

bfineran (approved):

PR #2326, opened 2024-06-10
Update Test Recipes for Latest Modifier Changes

Satrat:


tests/sparseml/transformers/obcq/recipes/sparse.yaml
```diff
@@ -5,9 +5,8 @@ test_stage:
       block_size: 128
       sequential_update: False
       percdamp: 0.01
-      mask_structure: "0:0"
       targets: [
         "model.layers.0",
         "model.layers.1",
-        "lm_head"
```
dbogunowicz, line 12: why can we remove this target?

dbogunowicz (approved):

tests/sparseml/transformers/obcq/recipes/sparse.yaml
dbogunowicz: oh, because it's linear layer and will be caught by the logic automatically?

dsikka (commented):
Have we confirmed weekly one-shot tests now pass?

bfineran (approved):

PR #2332, opened 2024-06-17
fix wheel build

horheynm:
Import is_release from version.py for setup.py.
Running gdb, _generate_version didnt recognize global variables, so made sure the necessary variables are in scope.


setup.py
```diff
@@ -16,18 +16,25 @@
...
+    version_major,
+    version_minor,
+    version_bug,
+) = get_release_and_version(package_path)
 
 # load and overwrite version and release info from sparseml package
 exec(open(os.path.join("src", "sparseml", "version.py")).read())
 print(f"loaded version {version} from src/sparseml/version.py")
-version_nm_deps = f"{version_major_minor}.0"
+version_nm_deps = f"{version}.0"
```
dhuangnm, line 37: This line should not change since version is the full version string?

horheynm: Is this the different than line 25 on the old code?
`version_major_minor = version`?

dhuangnm: From the old code you can see that it's using major.minor.0, but here in the new code {version} is the full string, so it becomes major.minor.hotfix.0?

src/sparseml/version.py
```diff
@@ -45,7 +49,7 @@ def _generate_version():
     "version_build",
     "version_major_minor",
 ]
-__version__ = _generate_version()
+__version__ = _generate_version(is_release, is_dev, version_base, dev_number)
```
dhuangnm, line 52: Do we need to pass these variables? I think they have been defined in the first part of this file?

horheynm: Yeah we need those parameters, if not they break

bfineran (approved):

PR #2349, opened 2024-07-02
[Fix] Allow to create `SparseAutoModelForCausalLM` with `trust_remote_code=True`

dbogunowicz:
## Feature Description

Now this executes properly:
```python
model = SparseAutoModelForCausalLM.from_pretrained("microsoft/Phi-3-mini-128k-instruct",trust_remote_code=True)
print(model.__class__.__name__)
>> 'Phi3ForCausalLM'
```
The hack was to temporarily rename the class so that the `from_pretrained` method could properly resolve the "remote code load" logic. The model return by the  `from_pretrained` method is unaffected later on, as it will have the class name of the loaded function  (as shown above)

This PR needs to land as well, so that the functionality is fully enabled: https://github.com/neuralmagic/compressed-tensors/pull/104


rahul-tuli (approved):

anmarques (approved):
I tested w/ microsoft/Phi-3-mini-128k-instruct both original and after SparseML quantization and it workd

